PR #2, opened 2020-10-21
add protocol fees fixin

smarx:
## Description

This change adds `FixinProtocolFees`, code that collects protocol fees using per-pool `FeeCollector` contracts. Once per epoch, a function on each `FeeCollector` is called to transfer the held ether (first converted to WETH) to the staking contract. Delaying this transfer and associated accounting yields significant gas savings.

## Testing instructions

`yarn build && yarn test` in `contracts/zero-ex`.

The basic happy path with WETH is tested, but all other tests are missing. Future commits will add more robust testing, including negative testing for failure modes.

## Types of changes

## Checklist:


contracts/zero-ex/contracts/src/fixins/FixinProtocolFees.sol
```diff
@@ -0,0 +1,116 @@
...
+import "@0x/contracts-erc20/contracts/src/v06/IEtherTokenV06.sol";
+import "../external/FeeCollector.sol";
+import "../features/libs/LibTokenSpender.sol";
+
+/// @dev Helpers for collecting protocol fees.
+abstract contract FixinProtocolFees {
+    bytes32 immutable feeCollectorCodeHash;
+
+    constructor() internal {
+        feeCollectorCodeHash = keccak256(type(FeeCollector).creationCode);
```
hysz, line 31: Is the `immutable` cheaper than a `constant` in this case bc it'll fill in the final `bytes32` value instead of the `keccak256(...)?

contracts/zero-ex/contracts/src/fixins/FixinProtocolFees.sol
```diff
@@ -0,0 +1,116 @@
...
+    )
+        internal
+    {
+        require(
+            msg.value == amount || msg.value == 0,
+            "FixinProtocolFees/WRONG_ETHER_AMOUNT");
+
+        FeeCollector feeCollector = _getFeeCollector(poolId);
+
+        if (msg.value == 0) {
```
hysz, line 52: I suppose we're making the assumption that if the taker is selling ETH then their protocol fee must also be paid in ETH. I think this is probably fine, wdyt @dorothy-zbornak? 

... I feel like any workaround could get pretty hairy bc we'd have to track which ETH is for the fill vs the protocol fee. 

If we are going to make this assumption we should document it somewhere in the code, perhaps even in the function description for `_collectProtocolFee`.

contracts/zero-ex/contracts/src/fixins/FixinProtocolFees.sol
```diff
@@ -0,0 +1,116 @@
...
+    /// @param weth The WETH token contract.
+    function _collectProtocolFee(
+        bytes32 poolId,
+        uint256 amount,
+        IERC20TokenV06 weth
+    )
+        internal
+    {
+        require(
+            msg.value == amount || msg.value == 0,
```
hysz, line 47: I think we'll have to do `>= amount` in case the taker is also selling ETH or paying the fee multiple times.

contracts/zero-ex/contracts/src/fixins/FixinProtocolFees.sol
```diff
@@ -0,0 +1,116 @@
...
+        FeeCollector feeCollector = _getFeeCollector(poolId);
+
+        uint256 codeSize;
+        assembly {
+            codeSize := extcodesize(feeCollector)
+        }
+
+        if (codeSize == 0) {
+            // Create and initialize the contract if necessary.
+            new FeeCollector{salt: poolId}();
```
hysz, line 80: When we include `salt` does it implicitly use `create2`?

hysz (commented):
Solidity is looking very clean! A few small comments/questions.

contracts/zero-ex/contracts/src/fixins/FixinProtocolFees.sol
```diff
@@ -0,0 +1,116 @@
...
+    function _collectProtocolFee(
+        bytes32 poolId,
+        uint256 amount,
+        IERC20TokenV06 weth
+    )
+        internal
+    {
+        require(
+            msg.value == amount || msg.value == 0,
+            "FixinProtocolFees/WRONG_ETHER_AMOUNT");
```
dorothy-zbornak, line 48: Hmm, I wonder if we'll ever allow filling a WETH order with ETH, causing this to break. Guess we can cross that bridge when we get there.

contracts/zero-ex/contracts/src/fixins/FixinProtocolFees.sol
smarx: Yeah, my understanding is that a constant with the value `keccak256(...)` will be reevaluated at each access. (I wish Solidity just disallowed it.) But the `immutable` can be computed just once.

contracts/zero-ex/contracts/src/fixins/FixinProtocolFees.sol
dorothy-zbornak, line 80: Syntax god.

dorothy-zbornak (approved):
wtg! I'll fix the changelog thing

contracts/zero-ex/contracts/src/fixins/FixinProtocolFees.sol
dorothy-zbornak: I think it'll be fine for what we have roadmapped. Might need to tweak in the future.

smarx: Ah, I was borrowing the logic from https://github.com/0xProject/0x-monorepo/blob/development/contracts/staking/contracts/src/fees/MixinExchangeFees.sol#L155-L174. But I think I've overdone it... if ETH can be traded directly, my check that `msg.value == amount` won't work.

contracts/zero-ex/contracts/src/fixins/FixinProtocolFees.sol
smarx: Ah, yes, great minds think alike. I just mentioned this in my response to the previous comment. :-) Yeah, this needs to be rethought.

contracts/zero-ex/contracts/src/fixins/FixinProtocolFees.sol
smarx: Yes, exactly.

contracts/zero-ex/contracts/src/fixins/FixinProtocolFees.sol
dorothy-zbornak: I *think* there will be no case in our current roadmap where `msg.value != fee`, but I can change it if it does come up.

hysz (approved):
Nice work! Let's merge (\~‾▿‾)~

PR #4, opened 2020-10-22
chore: [asset swapper] sampler Solidity 0.6 + Bridge addresses in AS

dekz:
## Description

Upgrades the sampler to solidity 0.6.
Refactors into try/catch where possible

[Sell sims](https://metabase.spaceship.0x.org/dashboard/139?run_id=choke-obtainable-scissors&adjusted_amount_bought_win_tolerance__bps_=0.5)
* 0x
* Kyber
* Balancer
* Mooniswap
* DODO
* Uniswap_V2
* SushiSwap
* Swerve
* Curve
* LiquidityProvider
* MultiHop

Missing:
* MStable
* Eth2Dai
* Shell

Included in price comparisons, just not good pricing at the moment
```js
{
name: "Shell",
price: "0.980418",
gas: "300000"
},
{
name: "Eth2Dai",
price: "0.968364",
gas: "400000"
},
```

```
{
name: "mStable",
price: "0.9994",
gas: "700000"
},
```

## Checklist:


packages/asset-swapper/contracts/src/LiquidityProviderSampler.sol
```diff
@@ -136,14 +130,15 @@ contract LiquidityProviderSampler is
         if (registryAddress == address(0)) {
             return address(0);
         }
+
         bytes memory callData = abi.encodeWithSelector(
-            ILiquidityProviderRegistry(0).getLiquidityProviderForMarket.selector,
+        ILiquidityProviderRegistry(0).getLiquidityProviderForMarket.selector,
```
dorothy-zbornak, line 135:
```suggestion
        ILiquidityProviderRegistry.getLiquidityProviderForMarket.selector,
```

packages/asset-swapper/contracts/src/LiquidityProviderSampler.sol
```diff
@@ -136,14 +130,15 @@ contract LiquidityProviderSampler is
...
             takerToken,
             makerToken
         );
         (bool didSucceed, bytes memory returnData) = registryAddress.staticcall(callData);
```
dorothy-zbornak, line 139: Why no `try/catch` here?

dorothy-zbornak (approved):
So much cleaner. Thank you!

mzhu25 (approved):
thanks!

packages/asset-swapper/contracts/src/LiquidityProviderSampler.sol
dekz: There's explicit tests for when the registry doesn't exist, where we want to fail silently. I believe we hit the case where the abi decode failed and reverted.

romainbutteaud:
@dekz I think you forgot SnowSwap GAS_SCHEDULE?

PR #5, opened 2020-10-22
chore: [asset-swapper] Move Bridge Addresses and Gas schedule

dekz:
## Description

Moves the BridgeAddresses into asset-swapper to reduce the number of changes required when adding a new bridge.

Moves the Gas schedule into asset-swapper but allows it to be overridden in consuming app.
Exports these outside of asset-swapper so the consumer can merge

## Checklist:


kimpers (approved):
LGTM 👍

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -283,3 +284,139 @@ export const ONE_SECOND_MS = 1000;
...
+    // tslint:disable-next-line: custom-no-magic-numbers
+    runLimit: 2 ** 15,
+    excludedSources: [],
+    excludedFeeSources: [],
+    includedSources: [],
+    bridgeSlippage: 0.005,
+    maxFallbackSlippage: 0.05,
+    numSamples: 13,
+    sampleDistributionBase: 1.05,
+    feeSchedule: DEFAULT_GAS_SCHEDULE,
```
dorothy-zbornak, line 417: We could technically also embed `FEE_SCHEDULE` because we do have a `PROTOCOL_FEE_MULTIPLIER` in constants. Though maybe it just makes more sense to move that constant out into a SwapQuoter opt.

dorothy-zbornak (approved):
:clap: :clap: 
Should prob add a CHANGELOG entry to `contract-addresses` and `migrations` tho.

PR #8, opened 2020-10-23
Update yarn.lock to match recent dependency update

feuGeneA:
Update yarn.lock to match the recent update of dependency
@0x/quote-server in package.json.


PirosB3 (changes requested):
why are you scaling back quote-server dependencies?

PirosB3:
https://github.com/0xProject/protocol/pull/1

PR #9, opened 2020-10-23
Change contracts/integrations to use LOCAL asset-swapper

feuGeneA:
Change contracts/integrations/package.json to refer to the LOCAL asset-swapper, not a gitpkg one, since asset-swapper does live in this same repo now.


dorothy-zbornak (approved):
ahhh totally forgot about this guy, nice catch!

PR #10, opened 2020-10-23
Have CI ensure there aren't any yarn.lock changes

feuGeneA:
## Description

Add a check to the `build` stage of the CI configuration that will fail if there are any local changes to `yarn.lock` after running `yarn install`.

A "passing" scenario looks like:

image

A "failing" scenario looks like:

image


dorothy-zbornak:
How would this ever fail if CI always installs with `yarn --frozen-lockfile`?

PR #11, opened 2020-10-24
Allow injection of an RfqtFirmQuoteValidator and an RfqtQuoteObserver

feuGeneA:
For use by upcoming work on price-aware RFQ-T requests.


packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -174,8 +183,8 @@ export class QuoteRequestor {
             'firm',
         );
 
-        const result: RFQTFirmQuote[] = [];
-        firmQuoteResponses.forEach(firmQuoteResponse => {
+        const validatedResponses: Array<{ response: RFQTFirmQuote; makerUri: string }> = [];
+        firmQuoteResponses.forEach(async firmQuoteResponse => {
```
PirosB3, line 187: why `async` here?

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -694,20 +694,24 @@ export class MarketOperationUtils {
                         rfqt,
                     );
                     if (firmQuotes.length > 0) {
+                        // Compute the RFQ order fillable amounts. This is done by performing a "soft" order
+                        // validation and by checking order balances that are monitored by our worker.
+                        // If a firm quote validator does not exist, then we assume that all orders are valid.
+                        const firmQuoteSignedOrders = firmQuotes.map(quote => quote.signedOrder);
+                        const rfqOrderFillableAmounts =
+                            rfqt.firmQuoteValidator === undefined
+                                ? firmQuoteSignedOrders.map(signedOrder => signedOrder.takerAssetAmount)
```
alexkroeger, line 703: Just to clarify:

This will be undefined until we build this out, so here we're just assuming it's 100% fillable.

alexkroeger (approved):

packages/asset-swapper/src/utils/market_operation_utils/index.ts
PirosB3: That is correct

PR #16, opened 2020-10-27
Feature/liquidity provider sandbox

mzhu25:
## Description

- Updates LiquidityProviderFeature to use an off-chain registry instead of an on-chain one
- Removes MultiBridge stuff from AS
- Updates the `ILiquidityProvider` interface
- Updates the `ProtocolFeeUnfunded` event in FQT

## Checklist:


contracts/zero-ex/contracts/src/vendor/ILiquidityProvider.sol
```diff
@@ -0,0 +1,90 @@
...
+    /// @param auxiliaryData Arbitrary auxiliary data supplied to the contract.
+    /// @return boughtAmount The amount of `makerToken` bought.
+    function sellEthForToken(
+        address makerToken,
+        address recipient,
+        uint256 minBuyAmount,
+        bytes calldata auxiliaryData
+    )
+        external
+        returns (uint256 boughtAmount);
```
dorothy-zbornak, line 56: Isn't it cheaper to just make this `payable` than having to do a separate transfer call?

dorothy-zbornak:
think a rebase will fix the `test-publish` :crossed_fingers:

contracts/zero-ex/contracts/src/features/LiquidityProviderFeature.sol
```diff
@@ -50,16 +47,14 @@ contract LiquidityProviderFeature is
...
 
-    /// @dev Store the WETH address in an immutable.
-    /// @param weth_ The weth token.
-    constructor(IEtherTokenV06 weth_)
+    constructor(address zeroEx)
         public
         FixinCommon()
     {
-        weth = weth_;
+        sandbox = new LiquidityProviderSandbox(zeroEx);
```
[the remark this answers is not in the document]
mzhu25: yeah I was thinking in the next iteration of this feature, `sandbox` would be passed in as a constructor arg. no need to pass it in for this first version though, might as well deploy it in the same txn

contracts/zero-ex/contracts/src/vendor/ILiquidityProvider.sol
mzhu25: it's actually a bit more expensive because we have to make 2 payable calls (EP => sandbox => liquidity provider).
47194 gas vs 53008 according to ganache

dorothy-zbornak: Ah, forgot about the sandbox. Maybe it should be payable anyway if we ever interact with it from a transformer?

mzhu25: yeah I suppose so

dorothy-zbornak (approved):
:clap: :clap: :clap:

dorothy-zbornak:
Has this been simbotted btw?

PR #27, opened 2020-10-31
EP Native Orders

dorothy-zbornak:
## Description

Adds orders (limit and RFQ) to the EP via the `NativeOrdersFeature`.
 
Some notes:
- Refer to the updated docs (`orders.rst` file) for a high-level overview of EP orders.
- Limit orders and RFQ orders are two different types, with (mostly) different code paths and specialized interaction functions.
- Instead of `cancelOrdersUpTo()` we now have `cancelPairLimitOrdersUpTo()` and `cancelPairRfqOrdersUpTo()` which cancels based on order pair and salt.
- I refactored `FixinProtocolFees` a bit to be nicer to use in this feature. Need to rebase to include its new tests.
- We no longer support paying protocol fees in WETH.
- We only work with `uint128` amounts in orders now.
- We set the (unused) high bit in the `orderHashToTakerTokenFilledAmount` storage slot to indicate the hash was cancelled via `cancelLimitOrder()` or `cancelRfqOrder()`.
- A bunch of tooling was invented for orders and signing and temporarily exists in the `@0x/contracts-zero-ex` package. One day we will move them out to a `@0x/protocol-utils` package for production usage.
  - I deliberately avoided using `@0x/order-utils` where I could so we can more easily make a clean break when that day comes.
- A bunch of new revert reasons are also in the ^ package temporarily.
- There is a `transferProtocolFeesForPools()` function that this new feature exposes that will move accumulated fees out of the fee collector and into staking contract. This is permissionless.

## Checklist:


contracts/zero-ex/contracts/src/features/libs/LibLimitOrder.sol
```diff
@@ -0,0 +1,207 @@
...
+    //       "address taker,",
+    //       "address sender,",
+    //       "address feeRecipient,",
+    //       "uint256 pool,",
+    //       "uint64 expiry,",
+    //       "uint256 salt"
+    //     ")"
+    // ))
+    uint256 private constant _LIMIT_ORDER_TYPEHASH =
+        0x19457468657265756d205369676e6564204d6573736167653a0a333200000000;
```
dorothy-zbornak, line 89: these are totally made up rn

contracts/zero-ex/contracts/src/features/libs/LibLimitOrder.sol
```diff
@@ -0,0 +1,208 @@
...
+            // order.takerTokenFeeAmount;
+            mstore(add(mem, 0xA0), mload(add(order, 0x80)))
+            // order.maker;
+            mstore(add(mem, 0xC0), mload(add(order, 0xA0)))
+            // order.taker;
+            mstore(add(mem, 0xE0), mload(add(order, 0xC0)))
+            // order.sender;
+            mstore(add(mem, 0x100), mload(add(order, 0xE0)))
+            // order.feeRecipient;
+            mstore(add(mem, 0x120), mload(add(order, 0x100)))
```
smarx, line 152: I think we may have to `and` addresses to mask off the upper bytes. Solidity can leave junk in those bits, or it might even be possible for someone to maliciously get them there on purpose via a `LimitOrder` struct passed via call data. I'm honestly not sure when we're guaranteed those bits are zeroed and when we're not.

smarx: Oh, and same thing below in `getRfqOrderStructHash()`.

dorothy-zbornak: Hmm, I wonder if we would ever actually take advantage of the dirty bits to embed extra data that could be part of the hash. I guess there's also an argument for having dirty bits that don't get hashed for injecting metadata post-signing.

contracts/zero-ex/contracts/src/features/LimitOrdersFeature.sol
```diff
@@ -0,0 +1,875 @@
...
+            );
+            LibTokenSpender.spendERC20Tokens(
+                params.order.takerToken,
+                params.taker,
+                params.order.feeRecipient,
+                feeAmount
+            );
+        }
+
+        emit LimitOrderFilled(
```
mzhu25, line 710: no `feeAmountPaid`?

contracts/zero-ex/contracts/src/features/LimitOrdersFeature.sol
```diff
@@ -0,0 +1,875 @@
...
+    /// @return makerTokenFilledAmount How much maker token was filled.
+    function _fillRfqOrderPrivate(
+        LibLimitOrder.RfqOrder memory order,
+        LibSignature.Signature memory signature,
+        uint128 takerTokenFillAmount,
+        address taker
+    )
+        private
+        returns (
+            uint256 ethProtocolFeePaid,
```
mzhu25, line 739: why do we bubble up the `ethProtocolFeePaid` but not the `wethProtocolFeePaid` for these functions?

contracts/zero-ex/contracts/src/features/LimitOrdersFeature.sol
```diff
@@ -0,0 +1,875 @@
...
+            taker,
+            address(order.makerToken),
+            address(order.takerToken),
+            takerTokenFilledAmount,
+            makerTokenFilledAmount,
+            order.pool
+        );
+    }
+
+    /// @dev Settle a the trade between an order's maker and taker.
```
mzhu25, line 805:
```suggestion
    /// @dev Settle the trade between an order's maker and taker.
```

contracts/zero-ex/contracts/src/features/LimitOrdersFeature.sol
```diff
@@ -0,0 +1,875 @@
...
+        makerTokenFilledAmount = uint128(
+            uint256(takerTokenFilledAmount)
+            * uint256(settleInfo.makerAmount)
+            / uint256(settleInfo.takerAmount)
+        );
+
+        if (takerTokenFilledAmount == 0 || makerTokenFilledAmount == 0) {
+            // Nothing to do.
+            return (takerTokenFilledAmount, makerTokenFilledAmount);
+        }
```
mzhu25, line 830: this can be somewhat misleading, no? We could have `takerTokenFilledAmount > 0` but `makerTokenFilledAmount = 0` due to rounding. I think we should `return(0, 0);` here

contracts/zero-ex/contracts/src/features/LimitOrdersFeature.sol
```diff
@@ -0,0 +1,875 @@
...
+
+        // Update filled state for the order.
+        LibLimitOrdersStorage
+            .getStorage()
+            .orderHashToTakerTokenFilledAmount[settleInfo.orderHash] =
+            // OK to overwrite the whole word because we shouldn't get to this
+            // function if the order is cancelled.
+                settleInfo.takerTokenFilledAmount.safeAdd(takerTokenFilledAmount);
```
mzhu25, line 838: I don't think we need safe math here: 
```
takerTokenFilledAmount = LibSafeMathV06.min128(
            settleInfo.takerTokenFillAmount,
            settleInfo.takerAmount.safeSub128(settleInfo.takerTokenFilledAmount)
        );
```
Implies that: `takerTokenFilledAmount + settleInfo.takerTokenFilledAmount <= settleInfo.takerAmount`, which is a `uint128`

contracts/zero-ex/src/eip712_utils.ts
```diff
@@ -0,0 +1,70 @@
...
+];
+
+const EXCHANGE_PROXY_EIP712_DOMAIN_DEFAULT = {
+    chainId: 1,
+    verifyingContract: NULL_ADDRESS,
+    name: 'ZeroEx',
+    version: '1.0.0',
+};
+
+const EXCHANGE_PROXY_DOMAIN_TYPEHASH = hexUtils.hash(
```
mzhu25, line 24: neato

contracts/zero-ex/src/migration.ts
```diff
@@ -90,58 +111,113 @@ export interface FullFeatures extends BootstrapFeatures {
...
 }
 
+/**
+ * Configuration for deploying full features..
+ */
+export interface FullFeaturesDeployConfig {
+    zeroExAddress: string;
+    wethAddress: string;
+    stakingAddress: string;
+    protocolFeeMultiplier: number | BigNumber;
```
mzhu25, line 135: tbh not a huge fan of all the `number | BigNumber` types we've got floating around (especially in asset-swapper)
```suggestion
    protocolFeeMultiplier: number;
```

mzhu25: this is a `uint32` anyway

contracts/zero-ex/src/signature_utils.ts
```diff
@@ -90,24 +108,33 @@ export function ecSignHashWithKey(hash: string, key: string): ECSignature {
...
+        return {
+            r: hexUtils.slice(rpcSig, 0, 32),
+            s: hexUtils.slice(rpcSig, 32, 64),
+            v,
+        };
     }
-    const typeId = parseInt(signatureBytes.slice(-2), 16) as SignatureType;
-    if (!Object.values(SignatureType).includes(typeId)) {
-        throw new Error(`Invalid signatureBytes type ID detected: ${typeId}`);
+    // Format should be V,R,S
```
mzhu25, line 130: I think you're missing a `v = parseInt(rpcSig.slice(2, 4), 16);` here

contracts/zero-ex/test/features/limit_orders_feature_test.ts
```diff
@@ -0,0 +1,1184 @@
...
+            expect(status).to.eq(OrderStatus.Filled); // Still reports filled.
+        });
+
+        it('can cancel a partially filled order', async () => {
+            const order = getTestLimitOrder();
+            await fillLimitOrderAsync(order, order.takerAmount.minus(1));
+            const receipt = await zeroEx.cancelLimitOrder(order).awaitTransactionSuccessAsync({ from: maker });
+            verifyEventsFromLogs(receipt.logs, [{ orderHash: order.getHash() }], IZeroExEvents.OrderCancelled);
+            const { status } = await zeroEx.getLimitOrderInfo(order).callAsync();
+            expect(status).to.eq(OrderStatus.Cancelled); // Still reports filled.
```
mzhu25, line 373:
```suggestion
            expect(status).to.eq(OrderStatus.Cancelled);
```

contracts/zero-ex/contracts/src/features/LimitOrdersFeature.sol
dorothy-zbornak: Yeah, why not.

contracts/zero-ex/contracts/src/features/LimitOrdersFeature.sol
dorothy-zbornak: Just because the weth one isn't useful for our accounting right now.

dorothy-zbornak: Maybe if we decide to emit the protocol fee paid as part of the event? We should discuss this in the v4 meeting.

contracts/zero-ex/src/signature_utils.ts
dorothy-zbornak: yikes

contracts/zero-ex/contracts/src/features/LimitOrdersFeature.sol
```diff
@@ -0,0 +1,875 @@
...
+        uint256 takerTokenFilledAmount;
+        (ethProtocolFeePaid, takerTokenFilledAmount, makerTokenFilledAmount) =
+            _fillRfqOrderPrivate(
+                order,
+                signature,
+                takerTokenFillAmount,
+                msg.sender
+            );
+        // Must have filled exactly the amount requested.
+        if (takerTokenFilledAmount != takerTokenFillAmount) {
```
hysz, line 262: Thinking about rounding errors, it could be tricky to get exactly `takerTokenFilledAmount`. Perhaps we could be a bit more lenient and do `if (takerTokenFilledAmount < takerTokenFillAmount)`? (And above)

contracts/zero-ex/contracts/src/features/LimitOrdersFeature.sol
```diff
@@ -0,0 +1,875 @@
...
+
+
+    /// @dev Cancel a single limit order. The caller must be the maker.
+    ///      Silently succeeds if the order has already been cancelled.
+    /// @param order The limit order.
+    function cancelLimitOrder(LibLimitOrder.LimitOrder memory order)
+        public
+        override
+    {
+        bytes32 orderHash = getLimitOrderHash(order);
```
hysz, line 347: I was just thinking... What if we indexed the `filled` mapping by the maker 🤔. Then we could just call `cancelLimitOrder` with the order hash and probably save a ton of gas.. So long as it would not make fills much more expensive -- I guess one additional `keccak` to lookup the cancelled mapping.

contracts/zero-ex/contracts/src/features/LimitOrdersFeature.sol
```diff
@@ -0,0 +1,875 @@
...
+            LibLimitOrdersRichErrors.OrderNotFillableBySenderError(
+                orderInfo.orderHash,
+                msg.sender,
+                params.order.sender
+            ).rrevert();
+        }
+
+        // Signature must be valid for the order.
+        {
+            address signer = LibSignature.getSignerOfHash(
```
hysz, line 664: Does `LibSignature.getSignerOfHash` revert if the `signer` is `address(0)`?

hysz (commented):
Really nice work! Very clean and thorough implementation. I left some initial comments and will do another deeper pass.

contracts/zero-ex/contracts/src/features/ILimitOrdersFeature.sol
```diff
@@ -0,0 +1,278 @@
...
+        bytes32 pool
+    );
+
+    /// @dev Emitted whenever a limt or RFQ order is cancelled.
+    /// @param orderHash The canonical hash of the order.
+    event OrderCancelled(
+        bytes32 orderHash
+    );
+
+    /// @dev Emitted whenever limit or RFQ orders are cancelled by pair by a maker.
```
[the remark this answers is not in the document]
dorothy-zbornak: Wait, why don't they need this?

contracts/zero-ex/contracts/src/features/LimitOrdersFeature.sol
```diff
@@ -0,0 +1,875 @@
...
+        // Maker of the order.
+        address maker;
+        // Taker of the order.
+        address taker;
+        // Maker token.
+        IERC20TokenV06 makerToken;
+        // Taker token.
+        IERC20TokenV06 takerToken;
+        // Maker token amount.
+        uint128 makerAmount;
```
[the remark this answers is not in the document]
dorothy-zbornak: Looking at the math, I don't think there's enough going on to where we're leaking much gas but we can golf this down if we aren't happy with the final numbers.

contracts/zero-ex/contracts/src/features/LimitOrdersFeature.sol
dorothy-zbornak: ye

contracts/zero-ex/contracts/src/storage/LibNativeOrdersStorage.sol
```diff
@@ -0,0 +1,54 @@
...
+
+    /// @dev Storage bucket for this feature.
+    struct Storage {
+        // How much taker token has been filled in order.
+        // The lower `uint128` is the taker token fill amount.
+        // The high bit will be `1` if the order was directly cancelled.
+        mapping(bytes32 => uint256) orderHashToTakerTokenFilledAmount;
+        // The minimum valid order salt for a given maker and order pair (maker, taker)
+        // for limit orders.
+        mapping(address => mapping(address => mapping(address => uint256)))
```
mzhu25, line 36: I think we can save ~60 gas per order if we flatten this mapping to be `keccak256(abi.encodePacked(maker, makerToken, takerToken)) => uint256`
probably not worth but just throwing it out there

contracts/zero-ex/contracts/src/storage/LibNativeOrdersStorage.sol
```diff
@@ -0,0 +1,54 @@
...
+            limitOrdersMakerToMakerTokenToTakerTokenToMinValidOrderSalt;
+        // The minimum valid order salt for a given maker and order pair (maker, taker)
+        // for RFQ orders.
+        mapping(address => mapping(address => mapping(address => uint256)))
```
mzhu25, line 40: same here

mzhu25 (approved):
looking good, just a gas golf suggestion for your consideration

contracts/zero-ex/contracts/src/features/NativeOrdersFeature.sol
```diff
@@ -0,0 +1,984 @@
...
+            makerTokenFilledAmount
+        );
+    }
+
+    /// @dev Refund any leftover protocol fees in `msg.value` to `msg.sender`.
+    /// @param ethProtocolFeePaid How much ETH was paid in protocol fees.
+    function _refundExcessProtocolFeeToSender(uint256 ethProtocolFeePaid)
+        private
+    {
+        if (msg.value > ethProtocolFeePaid && msg.sender != address(this)) {
```
hysz, line 971: > For small values it may be cheaper for the taker if we do not refund the protocol fee. I'd be down to add a lower bound that refundAmount must be exceed in order for us to refund.

I wasn't sure if we discussed this offline or if it just got lost in the mix. I do have a cheeky update though. Danning and I were looking into the data and it appears that there are several cases where folks are just 1 gwei off from the correct protocol fee. Which could be caused by a rounding error on their end, not sure. But that could be a min dust amount that we just don't refund. So like `if (ethProtocolFeePaid - msg.value <= 1 gwei) {do not refund}`. Wdyt?

hysz (commented):
Looks solid! Left a couple comments, mostly follow-ups from the convos that happened earlier today. Should be good to merge after 💯

contracts/zero-ex/contracts/src/features/NativeOrdersFeature.sol
```diff
@@ -0,0 +1,984 @@
...
+                takerToken: IERC20TokenV06(params.order.takerToken),
+                makerAmount: params.order.makerAmount,
+                takerAmount: params.order.takerAmount,
+                takerTokenFillAmount: params.takerTokenFillAmount,
+                takerTokenFilledAmount: orderInfo.takerTokenFilledAmount
+            })
+        );
+
+        // Pay the fee recipient.
+        results.takerTokenFeeFilledAmount;
```
smarx, line 807: What is this line doing? (If it's needed, a comment explaining it would be nice.)

contracts/zero-ex/contracts/src/features/NativeOrdersFeature.sol
```diff
@@ -0,0 +1,984 @@
...
+        );
+    }
+
+    /// @dev Refund any leftover protocol fees in `msg.value` to `msg.sender`.
+    /// @param ethProtocolFeePaid How much ETH was paid in protocol fees.
+    function _refundExcessProtocolFeeToSender(uint256 ethProtocolFeePaid)
+        private
+    {
+        if (msg.value > ethProtocolFeePaid && msg.sender != address(this)) {
+            uint256 refundAmount = msg.value.safeSub(ethProtocolFeePaid);
```
smarx, line 972: This safe math isn't necessary, given the previous check that `msg.value > ethProtocolFeePaid`. I'd only remove it, though, if the gas saving was significant, which I expect it's not. So just pointing it out. :-)

contracts/zero-ex/contracts/src/storage/LibNativeOrdersStorage.sol
smarx: If you're checking on this, I'd try `abi.encode()` in addition to `abi.encodePacked()`. I don't have a good guess as to which is chaper.

smarx (approved):
This code looks really good.

Caveat on my approve: I did _not_ spend much time looking at tests. (I'm happy to if you want.)

contracts/zero-ex/contracts/src/features/NativeOrdersFeature.sol
dorothy-zbornak: How often? Is this an arbitrary number we're going to be chasing forever? Is it more robust to say `1 * tx.gasprice`? I feel like if there's a rounding error, that's an error we should try to fix in the client before hacking a fix in the protocol. :shrug:

hysz: *** Note to reader - this has been backlogged.

contracts/zero-ex/contracts/src/features/INativeOrdersFeature.sol
```diff
@@ -0,0 +1,326 @@
...
+    /// @param maker The maker of the order.
+    /// @param taker The taker of the order.
+    /// @param feeRecipient Fee recipient of the order.
+    /// @param takerTokenFilledAmount How much taker token was filled.
+    /// @param makerTokenFilledAmount How much maker token was filled.
+    /// @param protocolFeePaid How much protocol fee was paid.
+    /// @param pool The fee pool associated with this order.
+    event LimitOrderFilled(
+        bytes32 orderHash,
+        address maker,
```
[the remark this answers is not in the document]
hysz: @dorothy-zbornak this is still in progress. I've created a separate task in our backlog to index if Maker team requests.

hysz (approved):
Amazing work! Let's M E R G E (\~‾▿‾)~

PR #30, opened 2020-11-03
adds amendments to the rollout feature flag. Rollout indicative quote…

PirosB3:
Adds some new logic to the `isPriceAwareRFQEnabled` feature flag.

By default, indicative quotes should go through the new price-aware route (regardless of the value in `isPriceAwareRFQEnabled`). Firm quotes will go through the new price-aware route only when enabled by `isPriceAwareRFQEnabled` flag.

### DIFF
--- a/packages/asset-swapper/src/swap_quoter.ts
+++ b/packages/asset-swapper/src/swap_quoter.ts
@@ -41,6 +41,7 @@ import { ProtocolFeeUtils } from './utils/protocol_fee_utils';
 import { QuoteRequestor } from './utils/quote_requestor';
 import { sortingUtils } from './utils/sorting_utils';
 import { SwapQuoteCalculator } from './utils/swap_quote_calculator';
+import { getPriceAwareRFQRolloutFlags } from './utils/utils';
 import { ERC20BridgeSamplerContract } from './wrappers';
 
 export class SwapQuoter {
@@ -700,7 +701,7 @@ export class SwapQuoter {
 
         if (
             opts.rfqt && // This is an RFQT-enabled API request
-            !opts.rfqt.isPriceAwareRFQEnabled && // If Price-aware RFQ is enabled, firm quotes are requested later on in the process.
+            !getPriceAwareRFQRolloutFlags(opts.rfqt.isPriceAwareRFQEnabled).isFirmPriceAwareEnabled && // If Price-aware RFQ is enabled, firm quotes are requested later on in the process.
             opts.rfqt.intentOnFilling && // The requestor is asking for a firm quote
             opts.rfqt.apiKey &&
             this._isApiKeyWhitelisted(opts.rfqt.apiKey) && // A valid API key was provided
--- a/packages/asset-swapper/src/swap_quoter.ts
+++ b/packages/asset-swapper/src/swap_quoter.ts
@@ -701,7 +701,7 @@ export class SwapQuoter {
 
         if (
             opts.rfqt && // This is an RFQT-enabled API request
-            !getPriceAwareRFQRolloutFlags(opts.rfqt.isPriceAwareRFQEnabled).isFirmPriceAwareEnabled && // If Price-aware RFQ is enabled, firm quotes are requested later on in the process.
+            !getPriceAwareRFQRolloutFlags(opts.rfqt.priceAwareRFQFlag).isFirmPriceAwareEnabled && // If Price-aware RFQ is enabled, firm quotes are requested later on in the process.
             opts.rfqt.intentOnFilling && // The requestor is asking for a firm quote
             opts.rfqt.apiKey &&
             this._isApiKeyWhitelisted(opts.rfqt.apiKey) && // A valid API key was provided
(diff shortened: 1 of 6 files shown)


BMillman19 (approved):

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -726,8 +728,8 @@ export class MarketOperationUtils {
                         optimizerOpts,
                     );
                 }
-            } else {
-                // A firm quote is being requested. Ensure that `intentOnFilling` is enabled.
+            } else if (!rfqt.isIndicative && isFirmPriceAwareEnabled) {
```
alexkroeger, line 731: should there be an 'else if...' if priceAwareEnabled is false for both firm and indicative?

PirosB3: good question! if firm and indicative quotes are both disabled, none of that if/else block is triggered. In this scenario, the `optimizedOrders` variable is never replaced and proceeds with what was previously set before. This is expected.

Does that make sense? or am I missing smth?

alexkroeger: We have routes for

(rfqt.isIndicative && isIndicativePriceAwareEnabled)
and
(!rfqt.isIndicative && isFirmPriceAwareEnabled)

but was wondering if we needed routes for:

(rfqt.isIndicative && !isIndicativePriceAwareEnabled)
and
(!rfqt.isIndicative && !isFirmPriceAwareEnabled)

Or are we only using RFQT if it's price aware?

PirosB3: oh, that's a really good question. So, this pathway that starts [here](https://github.com/0xProject/protocol/pull/30/files#diff-863cb8194bd0f4e5a026d8ba07c4a5256dacd6e342a427d6953fa759b28f8239R683) is only accessed when price-aware RFQ is enabled. When Price-aware RFQ is disabled, we still do RFQ but at a different section of the code: [here](https://github.com/0xProject/protocol/pull/30/files#diff-24bb4e154d234a6991fb599588e16823703a26243956e07cacd9ea7722de7ba4L703) for firm quotes and [here](https://github.com/0xProject/protocol/pull/30/files#diff-863cb8194bd0f4e5a026d8ba07c4a5256dacd6e342a427d6953fa759b28f8239R218) for indicative quotes

PirosB3: oh, that's a really good question. So, this pathway that starts [here](https://github.com/0xProject/protocol/pull/30/files#diff-863cb8194bd0f4e5a026d8ba07c4a5256dacd6e342a427d6953fa759b28f8239R683) is only accessed when price-aware RFQ is enabled. When Price-aware RFQ is disabled, we still do RFQ but at a different section of the code: [here](https://github.com/0xProject/protocol/pull/30/files#diff-24bb4e154d234a6991fb599588e16823703a26243956e07cacd9ea7722de7ba4L703) for firm quotes and [here](https://github.com/0xProject/protocol/pull/30/files#diff-863cb8194bd0f4e5a026d8ba07c4a5256dacd6e342a427d6953fa759b28f8239R218) for indicative quotes

packages/asset-swapper/src/utils/utils.ts
```diff
@@ -4,35 +4,25 @@ import { BigNumber, NULL_BYTES } from '@0x/utils';
...
-    } else {
-        return {
-            isIndicativePriceAwareEnabled: true,
-            isFirmPriceAwareEnabled: false,
-        };
-    }
+export function getPriceAwareRFQRolloutFlags(priceAwareRFQFlags?: PriceAwareRFQFlags): PriceAwareRFQFlags {
+    return priceAwareRFQFlags !== undefined
+        ? priceAwareRFQFlags
+        : {
```
BMillman19, line 22: Only comment is this default is in two places in the code, here and above in `constants.ts`

BMillman19 (approved):

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -700,7 +701,7 @@ export class SwapQuoter {
 
         if (
             opts.rfqt && // This is an RFQT-enabled API request
-            !opts.rfqt.isPriceAwareRFQEnabled && // If Price-aware RFQ is enabled, firm quotes are requested later on in the process.
+            !getPriceAwareRFQRolloutFlags(opts.rfqt.priceAwareRFQFlag).isFirmPriceAwareEnabled && // If Price-aware RFQ is enabled, firm quotes are requested later on in the process.
```
alexkroeger, line 704: Should this include a similar check for the indicative flag?

e.g !getPriceAwareRFQRolloutFlags(opts.rfqt.priceAwareRFQFlag).isIndicativePriceAwareEnabled &&

PirosB3: no, because the firm and indicative quotes are 2 independent flags.

alexkroeger: indicative quotes are pulled here in the case of a false `isIndicativePriceAwareEnabled` flag: https://github.com/0xProject/protocol/blob/bf3035ad258f9a02581f7b5fca8ee8c6af047f37/packages/asset-swapper/src/utils/market_operation_utils/index.ts#L369

LGTM

alexkroeger (approved):

PR #31, opened 2020-11-04
Convert MetaTransactionsFeature to use the new LibSignature

smarx:
## Description

Convert `MetaTransactionsFeature` to use the new `LibSignature`

## Testing instructions

`yarn build && yarn test` from `contracts/zero-ex`

## Types of changes

* non-breaking change to use `LibSignature` instead of the signature validation feature

## Checklist:

-   [X] Prefix PR title with `[WIP]` if necessary.
-   [X] Add tests to cover changes as needed.
-   [X] Update documentation as needed.

### DIFF
--- a/contracts/zero-ex/CHANGELOG.json
+++ b/contracts/zero-ex/CHANGELOG.json
@@ -21,6 +21,10 @@
             {
                 "note": "Add a gas limit to first `LibTokenSpender` and `UniswapFeature` transfer",
                 "pr": 38
+            },
+            {
+                "note": "Convert metatransactions to use `LibSignature`",
+                "pr": 31
             }
         ]
     },
--- a/contracts/zero-ex/contracts/src/features/IMetaTransactionsFeature.sol
+++ b/contracts/zero-ex/contracts/src/features/IMetaTransactionsFeature.sol
@@ -20,10 +20,27 @@ pragma solidity ^0.6.5;
 pragma experimental ABIEncoderV2;
 
 import "@0x/contracts-erc20/contracts/src/v06/IERC20TokenV06.sol";
-
+import "./libs/LibSignature.sol";
 
 /// @dev Meta-transactions feature.
 interface IMetaTransactionsFeature {
+    /// @dev Describes the state of a meta transaction.
+    struct ExecuteState {
+        // Sender of the meta-transaction.
+        address sender;
+        // Hash of the meta-transaction data.
+        bytes32 hash;
+        // The meta-transaction data.
+        MetaTransactionData mtx;
+        // The meta-transaction signature (by `mtx.signer`).
+        LibSignature.Signature signature;
+        // The selector of the function being called.
+        bytes4 selector;
+        // The ETH balance of this contract before performing the call.
+        uint256 selfBalance;
+        // The block number at which the meta-transaction was executed.
+        uint256 executedBlockNumber;
+    }
 
     /// @dev Describes an exchange proxy meta transaction.
     struct MetaTransactionData {
@@ -88,15 +105,10 @@ interface IMetaTransactionsFeature {
 
     /// @dev Execute a meta-transaction via `sender`. Privileged variant.
     ///      Only callable from within.
-    /// @param sender Who is executing the meta-transaction..
-    /// @param mtx The meta-transaction.
-    /// @param signature The signature by `mtx.signer`.
+    /// @param state The `ExecuteState` for this metatransaction, with `sender`,
+    ///              `hash`, `mtx`, and `signature` fields filled.
     /// @return returnResult The ABI-encoded result of the underlying call.
-    function _executeMetaTransaction(
-        address sender,
-        MetaTransactionData calldata mtx,
-        bytes calldata signature
-    )
+    function _executeMetaTransaction(ExecuteState memory state)
         external
         payable
         returns (bytes memory returnResult);
--- a/contracts/zero-ex/contracts/src/features/IMetaTransactionsFeature.sol
+++ b/contracts/zero-ex/contracts/src/features/IMetaTransactionsFeature.sol
@@ -24,24 +24,6 @@ import "./libs/LibSignature.sol";
 
 /// @dev Meta-transactions feature.
 interface IMetaTransactionsFeature {
-    /// @dev Describes the state of a meta transaction.
-    struct ExecuteState {
-        // Sender of the meta-transaction.
-        address sender;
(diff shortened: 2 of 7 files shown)


contracts/zero-ex/contracts/src/features/MetaTransactionsFeature.sol
```diff
@@ -347,17 +348,34 @@ contract MetaTransactionsFeature is
...
+            s: state.signature.readBytes32(33)
+        });
+
+        if (LibSignature.getSignerOfHash(state.hash, sig) != state.mtx.signer) {
             LibMetaTransactionsRichErrors
                 .MetaTransactionInvalidSignatureError(
                     state.hash,
                     state.signature,
-                    err
+                    LibSignatureRichErrors.SignatureValidationError(
```
mzhu25, line 373: I think we can drop the nested error here and instead just use `LibSignatureRichErrors.SignatureValidationError` (@dorothy-zbornak opinion?). If we do want to keep the nested error however, we'd probably want the invalid length error above to also be nested.

smarx: My rationale here was to keep the existing tests working. So the errors are (hopefully) the same as what we used to return.

I'd definitely be happy to drop the nested error; I'm just pointing out that this would theoretically be a breaking change.

smarx: Oh, but I think it's true that the invalid length error _should_ be nested to preserve the earlier behavior.

So I failed at keeping the errors the same. Let's make a call on nesting or not, and I'll make one change or the other.

dorothy-zbornak: I'm down for removing the nested error. Probably not terribly useful in production.

contracts/zero-ex/contracts/src/features/MetaTransactionsFeature.sol
```diff
@@ -347,17 +348,34 @@ contract MetaTransactionsFeature is
...
                 ).rrevert();
         }
-        // Must be signed by signer.
-        try
-            ISignatureValidatorFeature(address(this))
-                .validateHashSignature(state.hash, state.mtx.signer, state.signature)
-        {}
-        catch (bytes memory err) {
+
+        if (state.signature.length != 66) {
```
dorothy-zbornak, line 352: I do think `state.signature` should be the deconsructed signature already and we unpack it in `executeMetaTransaction()` and `_executeMetaTransaction()`. Maybe `_executeMetaTransactionPrivate()` should just take an `ExecuteState` object as the sole input as well.

contracts/zero-ex/CHANGELOG.json
```diff
@@ -9,6 +9,10 @@
             {
                 "note": "Use new `checkAllowance` flag in LiquidityProviderFeature, TransformERC20Feature, and MetaTransactionsFeature",
                 "pr": 39
+            },
+            {
+                "note": "Convert metatransactions to use `LibSignature`",
+                "pr": 31
             }
```
dorothy-zbornak, line 16: This needs to go in a new version block because it got published (has a timestamp)

mzhu25 (approved):
lgtm

contracts/zero-ex/contracts/src/features/IMetaTransactionsFeature.sol
```diff
@@ -88,15 +105,10 @@ interface IMetaTransactionsFeature {
...
-    /// @param signature The signature by `mtx.signer`.
+    /// @param state The `ExecuteState` for this metatransaction, with `sender`,
+    ///              `hash`, `mtx`, and `signature` fields filled.
     /// @return returnResult The ABI-encoded result of the underlying call.
-    function _executeMetaTransaction(
-        address sender,
-        MetaTransactionData calldata mtx,
-        bytes calldata signature
-    )
+    function _executeMetaTransaction(ExecuteState memory state)
```
dorothy-zbornak, line 111: This function is intended to be called from outside this feature so it needs to take params like
```
address sender,
MetaTransactionData calldata mtx,
LibSignature.Signature calldata sigature
```

contracts/zero-ex/contracts/src/features/IMetaTransactionsFeature.sol
```diff
@@ -20,10 +20,27 @@ pragma solidity ^0.6.5;
 pragma experimental ABIEncoderV2;
 
 import "@0x/contracts-erc20/contracts/src/v06/IERC20TokenV06.sol";
-
+import "./libs/LibSignature.sol";
 
 /// @dev Meta-transactions feature.
 interface IMetaTransactionsFeature {
+    /// @dev Describes the state of a meta transaction.
+    struct ExecuteState {
```
dorothy-zbornak, line 28: Then this can move to inside the implementation contract.

contracts/zero-ex/contracts/src/features/MetaTransactionsFeature.sol
```diff
@@ -348,18 +318,17 @@ contract MetaTransactionsFeature is
...
+
+        if (LibSignature.getSignerOfHash(state.hash, state.signature) !=
+                state.mtx.signer) {
+            LibSignatureRichErrors.SignatureValidationError(
+                LibSignatureRichErrors.SignatureValidationErrorCodes.WRONG_SIGNER,
+                state.hash,
+                state.mtx.signer,
+                // TODO: Remove this field from SignatureValidationError
+                //       when rich reverts are part of the protocol repo.
+                ""
```
dorothy-zbornak, line 330: :+1:

dorothy-zbornak (commented):
Let me check with taker, but maybe we can just update the original execute functions to take a `LibSignature.Signature` type.

dorothy-zbornak:
Yeah they say it's prob fine to just change the signatures to all accept a `LibSignature.Signature` type since the mtx service is off now.

contracts/zero-ex/contracts/src/features/IMetaTransactionsFeature.sol
smarx: I don't understand why.

dorothy-zbornak: `ExecuteState` has a bunch of fields that outsiders shouldn't have to concern themselves with.

dorothy-zbornak (approved):
:+1:

PR #32, opened 2020-11-10
Improved ComparisonPrice function

alexkroeger:
## Description

The comparison price is meant to provide RFQT market makers with real-time info about competitor pricing.

This PR makes several improvements to the comparison price calculation:
- refactors the calculation to be an imported function instead of inline code
- uses fill information instead of `optimizedOrders` that have slipped values
- accounts for backup orders by comparing the fills against the amount requested to be filled

To do:
- [x] Test out the function (see https://0xproject.quip.com/UbaGAT5gxrjs/Comparison-Price-Testing)
- [x] Add unit tests

## Checklist:


alexkroeger:
@PirosB3 this PR returns an extensible ComparisonPrice object (see https://github.com/0xProject/protocol/blob/c3d54aeea2bcc7a796eda180103c76abc4cb997b/packages/asset-swapper/src/utils/market_operation_utils/types.ts#L393) like we discussed.

However, the response schema for comparison price is already set as a single value for RFQT requests/quote reports. Right now, my solution is to leave that field alone (keeping it populated with a whole order price) and leave the option option to create another field later on (e.g. detailedComparisonPrice) that is a key-value object if we want to add additional detail here.

What do you think?

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
```diff
@@ -0,0 +1,75 @@
...
+): ComparisonPrice {
+    // the adjusted rate is defined as maker/taker
+    // input is the taker token for sells, input is the maker token for buys
+    const makerTakerOptimalRate = optimizerResult.adjustedRate;
+    const takerMakerOptimalRate = optimizerResult.adjustedRate.pow(-1);
+
+    // fees for a native order
+    const fees = optimizerResult.exchangeProxyOverhead(SOURCE_FLAGS.Native);
+    // Calc native order fee penalty in output unit (maker units for sells, taker unit for buys)
+    const feePenalty = marketSideLiquidity.ethToOutputRate.isZero()
```
alexkroeger, line 36: add `!`

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    } else if (marketSideLiquidity.side === MarketOperation.Buy) {
+        orderMakerAmount = amount;
+
+        orderTakerAmount = takerMakerOptimalRate.times(orderMakerAmount).minus(feePenalty);
+    } else {
+        throw new Error(`Unexpected marketOperation ${marketSideLiquidity.side}`);
+    }
+
+    if (orderMakerAmount.gt(0)) {
```
alexkroeger, line 58: Check `orderTakerAmount` instead

PirosB3:
> @PirosB3 this PR returns an extensible ComparisonPrice object (see
> 
> https://github.com/0xProject/protocol/blob/c3d54aeea2bcc7a796eda180103c76abc4cb997b/packages/asset-swapper/src/utils/market_operation_utils/types.ts#L393
> 
> ) like we discussed.
> However, the response schema for comparison price is already set as a single value for RFQT requests/quote reports. Right now, my solution is to leave that field alone (keeping it populated with a whole order price) and leave the option option to create another field later on (e.g. detailedComparisonPrice) that is a key-value object if we want to add additional detail here.
> 
> What do you think?

That's a great idea! let's do that :)

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -375,7 +375,7 @@ export const ONE_HOUR_IN_SECONDS = 60 * 60;
 export const ONE_SECOND_MS = 1000;
 export const NULL_BYTES = '0x';
 export const NULL_ADDRESS = '0x0000000000000000000000000000000000000000';
-export const COMPARISON_PRICE_DECIMALS = 5;
+export const COMPARISON_PRICE_DECIMALS = 10;
```
PirosB3, line 378: Just curious, why did you increase the decimal places? was there a precision issue?

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -578,10 +580,13 @@ export class MarketOperationUtils {
             }
         }
         const collapsedPath = optimalPath.collapse(orderOpts);
+
         return {
             optimizedOrders: collapsedPath.orders,
             liquidityDelivered: collapsedPath.collapsedFills as CollapsedFill[],
             sourceFlags: collapsedPath.sourceFlags,
+            adjustedRate: collapsedPath.adjustedRate(),
+            exchangeProxyOverhead: optimizerOpts.exchangeProxyOverhead,
```
PirosB3, line 589: Why do you want to return `exchangeProxyOverhead`? can't you simply fetch it from `_opts`?

PirosB3: (L599)

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -620,32 +625,17 @@ export class MarketOperationUtils {
         }
 
         // If RFQ liquidity is enabled, make a request to check RFQ liquidity
-        let comparisonPrice: BigNumber | undefined;
+        let comparisonPrice: ComparisonPrice | undefined;
```
PirosB3, line 628: Looking at this in more detail, do we really need `comparisonPrice` at this point? or can we just keep `wholeOrderPrice`?

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -578,10 +580,13 @@ export class MarketOperationUtils {
             }
         }
         const collapsedPath = optimalPath.collapse(orderOpts);
+
         return {
             optimizedOrders: collapsedPath.orders,
             liquidityDelivered: collapsedPath.collapsedFills as CollapsedFill[],
             sourceFlags: collapsedPath.sourceFlags,
+            adjustedRate: collapsedPath.adjustedRate(),
```
PirosB3, line 588: Shouldn't this be `optimalPathRate`?

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -657,7 +647,7 @@ export class MarketOperationUtils {
                     nativeOrders[0].takerAssetData,
                     side,
                     amount,
-                    comparisonPrice,
```
PirosB3, line 660: If we are going to change this environment variable, should we also change the function parameter name `comparisonPrice` for `getRfqtIndicativeQuotesAsync`?

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
```diff
@@ -0,0 +1,73 @@
...
+    // the adjusted rate is defined as maker/taker
+    // input is the taker token for sells, input is the maker token for buys
+    const makerTakerOptimalRate = optimizerResult.adjustedRate;
+    const takerMakerOptimalRate = optimizerResult.adjustedRate.pow(-1);
+
+    // fees for a native order
+    const fees = optimizerResult.exchangeProxyOverhead(SOURCE_FLAGS.Native);
+
+    // Calc native order fee penalty in output unit (maker units for sells, taker unit for buys)
+    const feePenalty = !marketSideLiquidity.ethToOutputRate.isZero()
```
PirosB3, line 33: sorry I keep on forgetting, when could this be 0?

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
```diff
@@ -0,0 +1,73 @@
...
+ * returns the price of the taker asset in terms of the maker asset
+ * So the RFQT MM should aim for a higher price
+ * @param optimizerResult the output of the optimizer, contains the best orders
+ * @param amount the amount specified by the client
+ * @param marketSideLiquidity the results from querying liquidity sources
+ * @return ComparisonPrice object with the prices for RFQ MMs to beat
+ */
+export function getComparisonPrices(
+    optimizerResult: OptimizerResult,
+    amount: BigNumber,
```
PirosB3, line 21: Does `amount` already remove the penalty from all the other sources that make the optimized on-chain quote?

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
```diff
@@ -0,0 +1,73 @@
...
+
+        orderTakerAmount = takerMakerOptimalRate.times(orderMakerAmount).minus(feePenalty);
+    } else {
+        throw new Error(`Unexpected marketOperation ${marketSideLiquidity.side}`);
+    }
+
+    if (orderTakerAmount.gt(0)) {
+        const optimalMakerUnitAmount = Web3Wrapper.toUnitAmount(
+            // round up maker amount -- err to giving more competitive price
+            orderMakerAmount.integerValue(BigNumber.ROUND_UP),
```
PirosB3, line 59: why are you converting these to ints? wouldn't you be cutting off decimal amount which is essential for pairs like `BTC/USDC` or `YFI/USDC`?

PirosB3 (changes requested):
Initial review! amazing stuff :)

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
PirosB3: Discussed

packages/asset-swapper/src/utils/market_operation_utils/index.ts
PirosB3: discussed

packages/asset-swapper/src/utils/market_operation_utils/index.ts
PirosB3: optimalPathRate

packages/asset-swapper/src/utils/market_operation_utils/index.ts
PirosB3:
```
                wholeOrderPrice = getComparisonPrices(optimizerResult, amount, marketSideLiquidity).
```

packages/asset-swapper/src/utils/market_operation_utils/index.ts
PirosB3: discussed, no change

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
PirosB3: when we don't have `ETH/tokenA`

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
```diff
@@ -0,0 +1,73 @@
...
+        orderMakerAmount = makerTakerOptimalRate.times(orderTakerAmount).plus(feePenalty);
+    } else if (marketSideLiquidity.side === MarketOperation.Buy) {
+        orderMakerAmount = amount;
+
+        orderTakerAmount = takerMakerOptimalRate.times(orderMakerAmount).minus(feePenalty);
+    } else {
+        throw new Error(`Unexpected marketOperation ${marketSideLiquidity.side}`);
+    }
+
+    if (orderTakerAmount.gt(0)) {
```
PirosB3, line 56: orderTakerAmount.gt(0) && orderMakerAmount.gt(0)

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
alexkroeger: Need ints for the web3wrapper function--they are base units

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
alexkroeger: keeping this for now

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
alexkroeger: done

packages/asset-swapper/src/utils/market_operation_utils/index.ts
alexkroeger: done

packages/asset-swapper/src/utils/market_operation_utils/index.ts
alexkroeger: done

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
```diff
@@ -4,56 +4,42 @@ import * as _ from 'lodash';
...
-    const makerTakerOptimalRate = optimizerResult.adjustedRate;
-    const takerMakerOptimalRate = optimizerResult.adjustedRate.pow(-1);
-
-    // fees for a native order
-    const fees = optimizerResult.exchangeProxyOverhead(SOURCE_FLAGS.Native);
-
     // Calc native order fee penalty in output unit (maker units for sells, taker unit for buys)
     const feePenalty = !marketSideLiquidity.ethToOutputRate.isZero()
-        ? marketSideLiquidity.ethToOutputRate.times(fees)
+        ? marketSideLiquidity.ethToOutputRate.times((feeSchedule[ERC20BridgeSource.Native] as FeeEstimate)(undefined))
```
PirosB3, line 28: I'm not sure I get what you are doing here, why is there `undefined` here?

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
```diff
@@ -4,56 +4,42 @@ import * as _ from 'lodash';
...
-        orderTakerAmount = takerMakerOptimalRate.times(orderMakerAmount).minus(feePenalty);
-    } else {
-        throw new Error(`Unexpected marketOperation ${marketSideLiquidity.side}`);
-    }
+    // the adjusted rate is defined as maker/taker
+    // input is the taker token for sells, input is the maker token for buys
+    const orderMakerAmount =
+        marketSideLiquidity.side === MarketOperation.Sell ? adjustedRate.times(amount).plus(feePenalty) : amount;
+    const orderTakerAmount =
+        marketSideLiquidity.side === MarketOperation.Sell ? amount : amount.dividedBy(adjustedRate).minus(feePenalty);
```
PirosB3, line 39: OK so you are not doing the `^ -1` any more? or is there some other change?

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -484,7 +479,7 @@ export class MarketOperationUtils {
                             allowFallback: _opts.allowFallback,
                         },
                     );
-                    return optimizerResult;
+                    return optimizedOrders;
```
PirosB3: Great! thanks

PirosB3 (approved):
This is great! thanks for doing such a great job. I have not had time to run this code, have you tried running it? do the price hints look reasonable?

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
dorothy-zbornak: We should prob comment this with a `// HACK:`.

dorothy-zbornak: Might as well put it into a variable as well since it appears twice.

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
alexkroeger: I simplified this block of code to just use this ternary logic. I condensed 7-8 lines or so into these two. Functionally they should be the same.

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
```diff
@@ -0,0 +1,73 @@
...
+export function getComparisonPrices(
+    optimizerResult: OptimizerResult,
+    amount: BigNumber,
+    marketSideLiquidity: MarketSideLiquidity,
+): ComparisonPrice {
+    // the adjusted rate is defined as maker/taker
+    // input is the taker token for sells, input is the maker token for buys
+    const makerTakerOptimalRate = optimizerResult.adjustedRate;
+    const takerMakerOptimalRate = optimizerResult.adjustedRate.pow(-1);
+
```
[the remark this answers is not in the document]
alexkroeger: yea I agree 👍

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
alexkroeger: Talked to @merklejerk about this. It takes in a `Fill` object, but we don't have one here of course. This should return the cost of filling a single native order, and from testing it looks like that's the case.

I'm going to test one more time after making final edits.

I'll put a comment explaining the intended use.

alexkroeger: It works bc the native order schedule just returns a constant value: https://github.com/0xProject/protocol/blob/4f82543bdf361dc82d5fa8eaf728550033003907/packages/asset-swapper/src/utils/market_operation_utils/constants.ts#L441

If the gas schedule for native orders had a dependency on the `fillData` tho, I think it wouldn't work.

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
```diff
@@ -0,0 +1,79 @@
...
+    let feeInEth: BigNumber | number;
+
+    // HACK: get the fee penalty of a single 0x native order
+    // The FeeSchedule function takes in a `FillData` object and returns a fee estimate in ETH
+    // We don't have fill data here, we just want the cost of a single native order, so we pass in undefined
+    // This works because the feeSchedule returns a constant for Native orders, this will need
+    // to be tweaked if the feeSchedule for native orders uses the fillData passed in
+    // 2 potential issues: there is no native fee schedule or the fee schedule depends on fill data
+    if (feeSchedule[ERC20BridgeSource.Native] === undefined) {
+        logUtils.warn('ComparisonPrice function did not find native order fee schedule');
```
PirosB3, line 36: When would this happen exactly?

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
```diff
@@ -0,0 +1,79 @@
...
+
+        return { wholeOrder };
+    } else {
```
PirosB3, line 39: can you just drop the `else` here? it looks like it is not needed?

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
```diff
@@ -0,0 +1,79 @@
...
+        try {
+            feeInEth = new BigNumber((feeSchedule[ERC20BridgeSource.Native] as FeeEstimate)(undefined));
```
PirosB3, line 41: When would this throw an error exactly? is there any way that we can avoid the try/catch?

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
alexkroeger: this would happen if a given fee schedule does not include native orders. Since asset-swapper you to specify your own fee schedule, it is possible to not have a function defined for native orders.

PirosB3: @alexkroeger is it worth having a `getFeeInETH(feeSchedule): BigNumber | undefined` function to factor out this logic here?

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
alexkroeger: It would throw if the feeSchedule you specify for native orders depends on the content of the orders themselves. See protocol/packages/asset-swapper/src/utils/market_operation_utils/constants.ts. The default native order fee is just a constant, but it's possible to use the fill data in a function.

packages/asset-swapper/src/utils/market_operation_utils/comparison_price.ts
alexkroeger: This issue was actually causing the error in the test prior to fixing it.

PR #40, opened 2020-11-18
protocol fee / staking integration tests

smarx:
## Description

New integration test that ensures basic functionality in collecting protocol fees actually works with the `Staking` contract.

## Testing instructions

`yarn test` from `contracts/integration`

## Types of changes

* added tests

## Checklist:


hysz (approved):
LGTM, nice work! (\~‾▿‾)~

dorothy-zbornak (approved):
Looks great. What do you hate so much about this?

smarx:
> Looks great. What do you hate so much about this?

@dorothy-zbornak what I hate is stuff like https://github.com/0xProject/protocol/pull/40/files#diff-0a0c41b85e30731306401f208914d19228b5f6195d04d95380dc3af360f0d076. I'm essentially just using a test contract from another package, but I have to do it by creating a new contract and inheriting cross-package. It seems like jumping through hoops for no gain because I'm still introducing cross-package dependencies on non-published test contracts.

dorothy-zbornak:
> > Looks great. What do you hate so much about this?
> 
> @dorothy-zbornak what I hate is stuff like https://github.com/0xProject/protocol/pull/40/files#diff-0a0c41b85e30731306401f208914d19228b5f6195d04d95380dc3af360f0d076. I'm essentially just using a test contract from another package, but I have to do it by creating a new contract and inheriting cross-package. It seems like jumping through hoops for no gain because I'm still introducing cross-package dependencies on non-published test contracts.

I think the gain is that we didn't pollute the packages people actually use. 🤷‍♀️ Maybe the long term answer is to move all the contract tests into a separate package.

PR #44, opened 2020-11-23
limit order metatransactions

smarx:
## Description

This adds support for metatransactions that invoke `fillLimitOrder` and `fillRfqOrder`.

## Testing instructions

`yarn test`

## Types of changes

* New feature

## Checklist:


contracts/zero-ex/contracts/src/features/MetaTransactionsFeature.sol
```diff
@@ -453,6 +457,81 @@ contract MetaTransactionsFeature is
...
+            from := add(data, 36) // skip length and selector
+            to := add(args, 32)   // after length
+        }
+        LibBytesV06.memCopy(to, from, data.length - 4);
+
+        (order, signature, takerTokenFillAmount) = abi.decode(args, (LibNativeOrder.LimitOrder, LibSignature.Signature, uint128));
+
+        return _callSelf(
+            state.hash,
+            abi.encodeWithSelector(
```
hysz, line 486: Actually... I wonder if if we could just modify in-place the `mtx.state.callData`. This is untested, but something like:
```
assembly {
    // length on input calldata
    let mtxCallDataLen := mload(data)

    // we'll extend the input calldata with two extra fields (0x40)
    mstore(data, add(mtxCallDataLen, 0x40)) 

    // where in memory to append these fields (right after state.mtx.callData)
    let mtxCallDataEnd := add(data, add(0x20, mul(mtxCallDataLen, 0x20))) 

    // add fields to end of state.mtx.callData
    mstore(mtxCallDataEnd, taker)
    mstore(add(mtxCallDataEnd, 0x20), sender)
}
```

hysz (commented):
Looking good! A few questions about optimizations. As well, let me confirm that we only need MTX's on the `fill` and not `fillOrKill`.

contracts/zero-ex/contracts/src/features/MetaTransactionsFeature.sol
hysz: ^^ we'd also have to replace the selector w/ `INativeOrdersFeature._fillLimitOrder.selector`

contracts/zero-ex/contracts/src/features/MetaTransactionsFeature.sol
```diff
@@ -453,6 +457,81 @@ contract MetaTransactionsFeature is
...
+    {
+        LibNativeOrder.LimitOrder memory order;
+        LibSignature.Signature memory signature;
+        uint128 takerTokenFillAmount;
+
+        bytes memory data = state.mtx.callData;
+        bytes memory args = new bytes(data.length - 4);
+        uint256 from;
+        uint256 to;
+        assembly {
```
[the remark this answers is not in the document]
smarx: We can do that, but I'm not convinced it's going to be (much) better. We're already doing a low-level memory copy.

contracts/zero-ex/contracts/src/features/MetaTransactionsFeature.sol
smarx: IIUC, this is a bit dangerous. We'd be writing to memory that may not be safe to touch. How important is gas efficiency here?

smarx: I guess we do control the struct, so we can make sure we have room to overwrite things. But I think we still need to decide how important gas efficiency is before we do that sort of thing.

hysz: Yeah... I was thinking we could probably show it's safe so long as we don't add code after executing the MTX. But, looking more thoroughly, I see we could also be running in a `batchExecute` in which case it's probably pretty likely we're overwriting someone else's memory.

contracts/zero-ex/contracts/src/features/MetaTransactionsFeature.sol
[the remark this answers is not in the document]
hysz: Yeah I was thinking this could save us the `abi.decode`, which looking in Remix is ~4000 gas . Alternatively we could probably also just use `memCpy` to extract the `order`/`signature`.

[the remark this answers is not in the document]
hysz: Actually given the time constraints I'm down to punt on optimizations for now. We can do more golfing in 4.1.

hysz (approved):
Nice work! (\~‾▿‾)~

dorothy-zbornak (approved):
Looking good. I'm not worried about optimizing this right now. We can wait to see if tokenlon has issues with it first.

PR #46, opened 2020-11-24
TransformERC20: Fix selling entire ETH balance

dorothy-zbornak:
## Description

In the `TransformERC20Feature`, when `inputTokenAmount == uint256(-1)`, we interpret this as trying to sell the taker's *entire* available balance. But we did not check if the input token was ETH. This causes reverts because `0xeeeee...` is not a token. Now we just set `inputTokenAmount = msg.value` when ETH is being sold and `inputTokenAmount == uint256(-1)`.

## Checklist:


smarx (approved):
Looks great!

hysz (approved):
LGTM

PR #49, opened 2020-11-25
feat: add takerAssetToEthRate and makerAssetToEthRate  to quote response

kimpers:
## Description

Adds `takerAssetToEthRate` and `makerAssetToEthRate` to the swap quote response object, to be able to determine the ETH value of these assets for any quoted pair.

**NOTE:** Rates can sometimes be missing, in those cases a rate of `0` will be returned.

## Checklist:

### DIFF
--- a/packages/asset-swapper/CHANGELOG.json
+++ b/packages/asset-swapper/CHANGELOG.json
@@ -1,5 +1,14 @@
 [
     {
+        "version": "5.4.0",
+        "changes": [
+            {
+                "note": "Add `takerAssetToEthRate` and `makerAssetToEthRate` to swap quote response",
+                "pr": 49
+            }
+        ]
+    },
+    {
         "timestamp": 1607036724,
         "version": "5.3.1",
         "changes": [
--- a/packages/asset-swapper/src/types.ts
+++ b/packages/asset-swapper/src/types.ts
@@ -198,6 +198,8 @@ export interface SwapQuoteBase {
     isTwoHop: boolean;
     makerTokenDecimals: number;
     takerTokenDecimals: number;
+    ethToTakerAssetRate: BigNumber;
+    ethToMakerAssetRate: BigNumber;
 }
 
 /**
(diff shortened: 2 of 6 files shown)


dekz (approved):
quick changelog then LGTM

packages/asset-swapper/src/types.ts
```diff
@@ -190,6 +190,8 @@ export interface SwapQuoteBase {
     isTwoHop: boolean;
     makerTokenDecimals: number;
     takerTokenDecimals: number;
+    ethToInputRate: BigNumber;
```
xianny, line 193: can we rename them as `takerAssetDataToEthRate` and `makerAssetDataToEthRate` (or eth to x) ? `Input` and `output` are confusing IMO

xianny:
https://github.com/0xProject/0x-api/pull/433/files#r534620017

I found this weirdness about the ethToX rates that I described in the link above.

If I'm right, we should either rename the field or invert the value. I suggest renaming the field as it's less likely to create silent bugs 🙏

kimpers:
> https://github.com/0xProject/0x-api/pull/433/files#r534620017
> 
> I found this weirdness about the ethToX rates that I described in the link above.
> 
> If I'm right, we should either rename the field or invert the value. I suggest renaming the field as it's less likely to create silent bugs 🙏

Great find, thanks! 🙏  I didn't notice that 😅  The naming is indeed wrong, the rate that is given is how many of X token will get you 1 ETH.  I will update the naming to be `takerAssetToEthRate` and `makerAssetToEthRate`

packages/asset-swapper/CHANGELOG.json
```diff
@@ -1,4 +1,14 @@
 [
+    {
+        "version": "5.4.0",
+        "changes": [
+            {
+                "note": "Add `takerAssetToEthRate` and `makerAssetToEthRate` to swap quote response",
+                "pr": 49
+            }
+        ],
+        "timestamp": 1607347477
```
xianny, line 10: Did you add this timestamp manually or...?

Asking because the publish workflow works by looking for changelog entries WITHOUT timestamps, publishing them, and then adding a timestamp after publishing.

kimpers: I most certainly did... Removing it 🤦

PR #52, opened 2020-11-26
[WIP] permissionless transformers with safety check and create2

smarx:
**STATUS**: This really needs some tests for `isDelegateCallSafe()`, so it's still a WIP, but I wanted to share early so we could course-correct if somebody doesn't like the general direction.

## Description

This change makes it possible for anyone to deploy a transformer. This is safe due to the new `isDelegateCallSafe()` function, which checks that a contract doesn't use any forbidden instructions like SELFDESTRUCT or DELEGATECALL. We also check for SLOAD and SSTORE. This isn't strictly necessary, but it helps prevent mistakes.

## Testing instructions

`yarn test`, but the actual functionality of `isDelegateCallSafe()` is largely untested so far.

## Types of changes

* New feature
* Breaking change

## Checklist:

### DIFF
--- a/contracts/zero-ex/contracts/src/external/TransformerDeployer.sol
+++ b/contracts/zero-ex/contracts/src/external/TransformerDeployer.sol
@@ -19,8 +19,6 @@
 pragma solidity ^0.6.5;
 pragma experimental ABIEncoderV2;
 
-import "@0x/contracts-utils/contracts/src/v06/AuthorizableV06.sol";
-
 
 /// @dev Deployer contract for ERC20 transformers.
 ///      Only authorities may call `deploy()` and `kill()`.
@@ -58,7 +56,7 @@ contract TransformerDeployer {
     /// @dev Checks whether a given address is safe to be called via
     ///      delegatecall. A contract is considered unsafe if it includes any
     ///      of the following opcodes: CALLCODE, DELEGATECALL, SELFDESTRUCT,
-    ///      SLOAD, and STORE. This code is adapted from
+    ///      CREATE, CREATE2, SLOAD, and STORE. This code is adapted from
     ///      https://github.com/dharma-eng/dharma-smart-wallet/blob/master/contracts/helpers/IndestructibleRegistry.sol
     /// @param target The address to check.
     /// @return True if the contract is considered safe for delegatecall.
@@ -101,6 +99,8 @@ contract TransformerDeployer {
                     op == 242 || // callcode
                     op == 244 || // delegatecall
                     op == 255 || // selfdestruct
+                    op == 240 || // create
+                    op == 245 || // create2
                     op == 84  || // sload
                     op == 85     // sstore
                 ) {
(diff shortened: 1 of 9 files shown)


contracts/zero-ex/contracts/src/external/TransformerDeployer.sol
```diff
@@ -22,64 +22,96 @@ pragma experimental ABIEncoderV2;
...
+                    continue;
+                }
+
+                // If opcode is impermissible, contract is unsafe.
+                if (
+                    op == 242 || // callcode
+                    op == 244 || // delegatecall
+                    op == 255 || // selfdestruct
+                    op == 84  || // sload
+                    op == 85     // sstore
```
hysz, line 105: This list looks good. In addition, we probably don't want the Flash Wallet deploying contracts so let's also blacklist `create` and `create2`.

contracts/zero-ex/contracts/src/external/TransformerDeployer.sol
```diff
@@ -22,64 +22,96 @@ pragma experimental ABIEncoderV2;
 import "@0x/contracts-utils/contracts/src/v06/AuthorizableV06.sol";
```
hysz, line 22: Let's put this in a new file, something like `PermissionlessTransformerDeployer`. We will likely continue using the existing `TransformerDeployer` for some time.

hysz (commented):
This is awesome!

contracts/zero-ex/contracts/src/external/TransformerDeployer.sol
smarx: Oh, great catch. Anything deployed via `create2` would be implicitly trusted, so it's really important to block that. (`create` is probably less critical but a good idea.)

contracts/zero-ex/contracts/src/external/TransformerDeployer.sol
smarx: There are lots of test changes that need to go along with switching the deployer because we move from nonces to init code hashes and salts. I think those are hard to keep side-by-side, so I'll try to figure out a way to preserve those. Maybe I'll just leave this branch with all the test changes, and we can just never merge it. But at least we'll have those changes handy when it's time for someone to recreate them.

Then I can just make a new PR that just adds `PermissionlessTransformerDeployer` and unit tests just its functionality.

What do you think?

smarx: It's not just tests, actually. `TransformERC20Feature` has to change, etc. So there's no really good side-by-side version of this, but we can still park just the `PermissionlessTransformerDeployer`. We just need to be aware that when it's time to use it, there will be a cascade of other code changes required.

hysz: Oh yeah, I would be in favor of splitting into two PR's. We can merge and audit the Solidity next week, then keep an open WIP PR w/ the integrations and testing. We can build off the testing PR when we're ready to make the switch.

smarx:
This pull request should never be merged. I'm leaving it here so when it's time to switch to the `PermissionlessTransformDeployer` there will be a head start on making the other necessary changes.

A noticeable omission in this PR is that I didn't manage to get integration tests working. It was tricky for a variety of reasons, so I punted. :-)

PR #54, opened 2020-11-27
Splits `PairOrdersCancelled` event into Limit/RFQ

hysz:
## Description

Right now we have a single `PairOrdersCancelled` event, which is emitted when cancelling a pair for Limit or Market orders. In this PR we split this event into `PairCancelledLimitOrders` and `PairCancelledRfqOrders` so that a listener will know which order type the market was cancelled under, without having to dig into the calldata.

## Checklist:


smarx (approved):

dorothy-zbornak (approved):
Good catch

PR #59, opened 2020-11-30
EP: FeeCollectorController

dorothy-zbornak:
## Description

Two issues:
- `NativeOrdersFeature` was not deployable because we compile the deployment bytecode the `FeeCollector` contract in `FixinProtocolFees` (2KB over).
- `FixinProtocolFees` computed the init code hash of the `FeeCollector` contract in its constructor. This means if we redeploy `NatieOrdersFeature` with a different compiler (or probably even if we don't), we will lose access to the existing `FeeController` instances.

So now we have a separate, non-upgradeable `FeeCollectorController` contract that is in charge of:
- Storing the deployment code and hash for `FeeController`s.
- Deploying `FeeController`s and wrapping their ETH (all in one step).

TODO:
- [x] Fix the broken migrations.
- [x] Fix the broken tests + probably add new ones

## Checklist:


mzhu25 (approved):
some 🇨🇦 , but lgtmow

PR #67, opened 2020-12-02
Independent compilation for contracts-zero-ex package

dorothy-zbornak:
## Description

Use independent compilation for `@0x/contracts-zero-ex` so we can verify contracts on etherscan again (and they'll look nicer). See https://github.com/0xProject/tools/pull/12 for context.

Also update org deps and unpin `@0x/contracts-gen` (not sure why it was in the first place).

## Checklist:


contracts/zero-ex/compiler.json
```diff
@@ -4,6 +4,7 @@
     "useDockerisedSolc": false,
     "isOfflineMode": false,
     "shouldSaveStandardInput": true,
+    "shouldCompileIndependently": true,
```
dorothy-zbornak, line 7: Here it is.

mzhu25 (approved):

PR #69, opened 2020-12-02
fix: handle potentially undefined path

xianny:
## Checklist:


packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -26,6 +26,7 @@ import {
     createSignedOrdersWithFillableAmounts,
     getNativeOrderTokens,
 } from './orders';
+import { Path } from './path';
```
kimpers, line 29: This seems unused or am I blind? 😂

kimpers (approved):
Thanks for fixing this! 🙏

PR #74, opened 2020-12-07
ExchangeProxy selector collision test

mzhu25:
## Description
Note that this test can raise false positives, because it tests the selectors of _all_ external/public functions of _all_ non-abstract contracts in the `@0x/contracts-zero-ex` package. 

## Checklist:


dorothy-zbornak (approved):
ye is cool

PR #80, opened 2020-12-10
fix: [asset-swapper] prevent error when multihop data is not present

dekz:
## Description

Filter out any two hops where the data just doesn't exist, i.e a route could not be found. Prevents errors downstream where it is expected to exist (quote report)

## Checklist:


mzhu25 (approved):
uhh better late than never eheh 😛

PR #88, opened 2020-12-16
fix: [asset-swapper] Rework Bancor to only use paths and sample best path

dekz:
## Description

[Sims](https://metabase.spaceship.0x.org/dashboard/139?run_id=bancor-1&adjusted_amount_bought_win_tolerance__bps_=0.5)

Only use the bancor-sdk to get the paths if they exist. It's a complete hack to reach in (all other functions are async) so made it fail gracefully. 

The on-chain option does not produce a good default path, so checking the best out of the candidates makes Bancor more competitive. 

Bancor uses ETH instead of WETH, so it must be withdrawn and deposited (when ETH is bought via Bancor).

Enables Bancor always for sells.

TODO
- [ ] Choose either `bridgeTransferFrom` with fallback hack, or implement new ILiquidityProvider `sellTokenForToken`
- [x] Deploy Bancor Bridge

## Checklist:


packages/asset-swapper/contracts/src/BancorSampler.sol
```diff
@@ -0,0 +1,131 @@
...
+                    .rateByPath
+                        {gas: BANCOR_CALL_GAS}
+                        (paths[i], takerTokenAmounts[takerTokenAmounts.length-1])
+                returns (uint256 amount)
+            {
+                if (amount > maxBoughtAmount) {
+                    maxBoughtAmount = amount;
+                    path = paths[i];
+                }
+            } catch (bytes memory) {
```
dorothy-zbornak, line 76:
```suggestion
            } catch {
```

packages/asset-swapper/contracts/src/BancorSampler.sol
```diff
@@ -0,0 +1,131 @@
...
+    function sampleBuysFromBancor(
+        address[][] memory paths,
+        address takerToken,
+        address makerToken,
+        uint256[] memory makerTokenAmounts
+    )
+        public
+        view
+        returns (address bancorNetwork, address[] memory path, uint256[] memory takerTokenAmounts)
+    {
```
dorothy-zbornak, line 120: What's the plan for this?

packages/asset-swapper/src/utils/market_operation_utils/bancor_service.ts
```diff
@@ -1,65 +1,34 @@
...
-            };
-        });
-    }
-
-    public getBancorNetworkAddress(): string {
-        const blockchain = this.sdk._core.blockchains[BlockchainType.Ethereum] as Ethereum;
-        return blockchain.bancorNetwork._address;
+    public getPaths(_fromToken: string, _toToken: string): string[][] {
+        // HACK: We reach into the blockchain object and pull in it's cache of tokens
+        // and we use it's internal non-async getPathsFunc
```
dorothy-zbornak, line 24: Might be worth pinning the `@bancor/sdk` revision if we're doing this.

packages/asset-swapper/src/utils/market_operation_utils/bancor_service.ts
```diff
@@ -1,65 +1,34 @@
...
-    return {
-        blockchainType,
-        blockchainId: address,
-    };
-}
+const findToken = (tokenAddress: string, graph: object): string =>
+    // If we're looking for WETH it is stored by Bancor as the 0xeee address
+    tokenAddress.toLowerCase() === TOKENS.WETH.toLowerCase()
+        ? '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE'
+        : Object.keys(graph).filter(k => k.toLowerCase() === tokenAddress.toLowerCase())[0];
```
dorothy-zbornak, line 12: Nit: regular function

packages/asset-swapper/src/utils/market_operation_utils/bancor_service.ts
```diff
@@ -1,65 +1,34 @@
...
-
-    public getBancorNetworkAddress(): string {
-        const blockchain = this.sdk._core.blockchains[BlockchainType.Ethereum] as Ethereum;
-        return blockchain.bancorNetwork._address;
+    public getPaths(_fromToken: string, _toToken: string): string[][] {
+        // HACK: We reach into the blockchain object and pull in it's cache of tokens
+        // and we use it's internal non-async getPathsFunc
+        try {
+            const blockchain = this.sdk._core.blockchains[BlockchainType.Ethereum] as Ethereum;
+            const fromToken = findToken(_fromToken, blockchain.graph);
```
dorothy-zbornak, line 27: Does this graph just keep itself updated continuously or does it need to be triggered?

packages/asset-swapper/src/utils/market_operation_utils/bancor_service.ts
dorothy-zbornak: Should we also revert if the token is not found?

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -84,15 +84,13 @@ export class SamplerOperations {
         public readonly provider?: SupportedProvider,
         public readonly balancerPoolsCache: BalancerPoolsCache = new BalancerPoolsCache(),
         public readonly creamPoolsCache: CreamPoolsCache = new CreamPoolsCache(),
-        protected readonly getBancorServiceFn?: () => BancorService, // for dependency injection in tests
         protected readonly tokenAdjacencyGraph: TokenAdjacencyGraph = { default: [] },
         public readonly liquidityProviderRegistry: LiquidityProviderRegistry = LIQUIDITY_PROVIDER_REGISTRY,
-    ) {}
+    ) {
+        void this.getBancorServiceAsync();
```
dorothy-zbornak, line 90: This syntax is new to me. Does it not work without `void`?
This is to prime the pump, right? Can we drop a comment mentioning that?

dorothy-zbornak (approved):
ty for figuring this out!

packages/asset-swapper/src/utils/market_operation_utils/bancor_service.ts
dekz: there's a background `refresh` running in the sdk

packages/asset-swapper/src/utils/market_operation_utils/bancor_service.ts
dekz: nah I want it to fail silently, like if it encounters a token that doesn't exist in Bancor. It'll just fail the lookup inside Bancor anyway

packages/asset-swapper/src/utils/market_operation_utils/bancor_service.ts
dekz: Good call

packages/asset-swapper/contracts/src/BancorSampler.sol
dekz: Figure it out one day. Will probably have to do what we do with Curve here.

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
dekz: It's a hack to avoid lint errors as we don't want to await it and we don't care about what it returns. 

I've reworked to 

```
        this.initBancorServiceAsync().catch(/* do nothing */);
```

PR #101, opened 2020-12-28
feat: opt-in positive slippage fee for integrators

romainbutteaud:
## Description

- 0xAPI change: https://github.com/0xProject/0x-api/pull/483

```
SELL UNI->ETH 228720.51 -> 2465.48 ($686161.52) after 36.3s @ dev
	✔ PASS 228720.51 -> 2465.48 ($4340140.44) ($71.95)
	Uniswap_V2: 88.42999999999999%, SushiSwap: 11.57%
	gas: 356964
SELL UNI->ETH 228720.51 -> 2465.48 ($686161.52) after 36.3s @ prod
	✔ PASS 228720.51 -> 2465.57 ($4340302.59) ($68.28)
	Uniswap_V2: 88.42999999999999%, SushiSwap: 11.57%
	gas: 338731
SELL WBTC->ETH 60.58 -> 1712.01 ($3033616.49) after 54.3s @ prod
	✔ PASS 60.58 -> 1712.63 ($3014616.28) ($274.58)
	Uniswap_V2: 11.57%, Kyber: 17.22%, Balancer: 5.646%, Mooniswap: 3.311%, SushiSwap: 62.25000000000001%
	gas: 1362231
SELL WBTC->ETH 60.58 -> 1712.01 ($3033616.49) after 54.3s @ dev
	✔ PASS 60.58 -> 1712.01 ($3013525.78) ($278.24)
	Uniswap_V2: 11.57%, Kyber: 17.22%, Balancer: 5.646%, Mooniswap: 3.311%, SushiSwap: 62.25000000000001%
	gas: 1380418
// buys less then best case
SELL ETH->DAI 288.17 -> 504691.88 ($507287.28) after 74.8s @ dev
	✔ PASS 288.17 -> 504529.00 ($504346.23) ($182.77)
	MultiHop: 100%
	gas: 906761
SELL ETH->DAI 288.17 -> 504691.88 ($507287.28) after 74.8s @ prod
	✔ PASS 288.17 -> 504529.00 ($504348.84) ($180.16)
	MultiHop: 100%
	gas: 893830
```

## Checklist:


contracts/zero-ex/contracts/src/transformers/PositiveSlippageFeeTransformer.sol
```diff
@@ -0,0 +1,81 @@
...
+    using LibSafeMathV06 for uint256;
+    using LibERC20Transformer for IERC20TokenV06;
+
+    /// @dev Information for a single fee.
+    struct TokenFee {
+        // The token to transfer to `recipient`.
+        IERC20TokenV06 token;
+        // Amount of each `token` to transfer to `recipient`.
+        // If `amount == uint256(-1)`, the entire balance of `token` will be
+        // transferred.
```
dekz, line 45:
```suggestion

```

contracts/zero-ex/contracts/src/transformers/PositiveSlippageFeeTransformer.sol
```diff
@@ -0,0 +1,81 @@
...
+        uint256 bestCaseAmount;
+        // Positive slippage threashold amount before sending the fees to `recipient`
+        // This is to avoid paying for extra gas to send positive slippage dust
+        uint256 positiveSlippageFeeThresholdAmount;
+        // Recipient of `token`.
+        address payable recipient;
+    }
+
+    uint256 private constant MAX_UINT256 = uint256(-1);
```
dekz, line 54:
```suggestion

```

contracts/zero-ex/contracts/src/transformers/PositiveSlippageFeeTransformer.sol
```diff
@@ -0,0 +1,81 @@
...
+    {
+        TokenFee[] memory fees = abi.decode(context.data, (TokenFee[]));
+
+        // Transfer tokens to recipients.
+        for (uint256 i = 0; i < fees.length; ++i) {
+            uint256 bestCaseAmount = fees[i].bestCaseAmount;
+            uint256 positiveSlippageFeeThresholdAmount = fees[i].positiveSlippageFeeThresholdAmount;
+            uint256 transformerAmount = LibERC20Transformer.getTokenBalanceOf(fees[i].token, address(this));
+            if (transformerAmount > bestCaseAmount) {
+                uint256 positiveSlippageAmount = transformerAmount - bestCaseAmount;
```
dekz, line 72: Rethinking this, can this be expressed by us increasing `bestCaseAmount` on behalf of the user. That is, say the `bestCaseAmount` was 100 DAI. Could we set this to 102 DAI to cover the cost of transferring any excess? Or really, in the consumer `bestCaseAmount + 20,000 * gasPrice * ethToMakerRate`

romainbutteaud: @dekz good point, more straightforward for the integrator as well. I made the changes, let me know

packages/order-utils/src/index.ts
```diff
@@ -77,6 +77,8 @@ export {
     AffiliateFeeTransformerData,
     encodeAffiliateFeeTransformerData,
     decodeAffiliateFeeTransformerData,
+    encodePositiveSlippageFeeTransformerData,
```
dorothy-zbornak, line 80: TODO (for the v4 version): move these to `protocol-utils`

packages/order-utils/src/transformer_utils.ts
```diff
@@ -152,7 +152,7 @@ export function decodePayTakerTransformerData(encoded: string): PayTakerTransfor
 }
 
 /**
- * ABI encoder for `PayTakerTransformer.TransformData`
+ * ABI encoder for `affiliateFeetransformer.TransformData`
```
dorothy-zbornak, line 155: 😇

packages/protocol-utils/src/transformer_utils.ts
```diff
@@ -148,7 +148,7 @@ export function decodePayTakerTransformerData(encoded: string): PayTakerTransfor
 }
 
 /**
- * ABI encoder for `PayTakerTransformer.TransformData`
```
dorothy-zbornak, line 151: Maybe better to just leave this package alone and worry about it in the v4 branch

contracts/zero-ex/contracts/src/transformers/PositiveSlippageFeeTransformer.sol
```diff
@@ -0,0 +1,71 @@
...
+    function transform(TransformContext calldata context)
+        external
+        override
+    returns (bytes4 success)
+    {
+        TokenFee[] memory fees = abi.decode(context.data, (TokenFee[]));
+
+        // Transfer tokens to recipients.
+        for (uint256 i = 0; i < fees.length; ++i) {
+            uint256 bestCaseAmount = fees[i].bestCaseAmount;
```
dorothy-zbornak, line 61: How exactly do multiple positive slippage fees work?

romainbutteaud: good question, I kept the same structure as for the `AffiliateFeeTransformer`, it can take an array of fees, but its not currently used as such. Is there a specific reason why would affiliateFee would take an array? If specific to affiliateFee we can remove it for PositiveSlippageFee. 

https://github.com/0xProject/protocol/blob/72a74e7c66e27da02dd9f4ce604ad057c740c304/contracts/zero-ex/contracts/src/transformers/AffiliateFeeTransformer.sol#L65

dorothy-zbornak: IIRC we specifically got asked for multiple affiliate fees by an integrator. But we don't expose it so no one is using it and no one complained so it's all pointless.

romainbutteaud: okay then I'm not gonna use it. Would removing it from `AffiliateFeeTransformer` reduce our gas footprint? Adding a fee already cost a lot in gas, wondering if we could optimize a bit.

romainbutteaud: okay I've removed the `fees` array

romainbutteaud:
rebased

packages/protocol-utils/src/transformer_utils.ts
romainbutteaud: okay I've removed it

romainbutteaud:
I was able to get the positive slippage extracted when there's any with simbot. 
Running now a simbot with more pairs to confirm no overall regressions.

romainbutteaud:
@dekz @merklejerk can you give it an other review? thanks!

romainbutteaud:
image

romainbutteaud:
simbot: https://metabase.spaceship.0x.org/dashboard/139?run_id=positive-slippage-01-20-001&adjusted_amount_bought_win_tolerance__bps_=0.5

packages/asset-swapper/src/quote_consumers/exchange_proxy_swap_quote_consumer.ts
```diff
@@ -225,25 +232,44 @@ export class ExchangeProxySwapQuoteConsumer implements SwapQuoteConsumerBase {
             });
         }
 
-        // This transformer pays affiliate fees.
-        if (buyTokenFeeAmount.isGreaterThan(0) && feeRecipient !== NULL_ADDRESS) {
+        if (feeType === AffiliateFeeType.PositiveSlippageFee && feeRecipient !== NULL_ADDRESS) {
+            // bestCaseAmount is increased to cover gas cost of sending positive slipapge fee to fee recipient
+            const bestCaseAmount = quote.bestCaseQuoteInfo.makerAssetAmount.plus(
+                POSITIVE_SLIPPAGE_FEE_TRANSFORMER_GAS.multipliedBy(quote.gasPrice).multipliedBy(
+                    quote.makerAssetToEthRate,
```
dorothy-zbornak, line 239: Should this be `ethToMakerAssetRAte`?

dorothy-zbornak, line 239: Should this be `ethToMakerAssetRate`?

packages/asset-swapper/src/types.ts
```diff
@@ -131,7 +131,13 @@ export interface ForwarderExtensionContractOpts {
     feeRecipient: string;
 }
 
-export interface AffiliateFee {
+export enum AffiliateFeeType {
+    PercentageFee,
```
dorothy-zbornak, line 135: Wondering if it's cleaner/clearer to also have an explicit `None` type.

romainbutteaud: right, I think it does and would make "do we have an affiliateFee" checks easier

packages/order-utils/src/transformer_utils.ts
```diff
@@ -195,6 +195,50 @@ export function decodeAffiliateFeeTransformerData(encoded: string): AffiliateFee
...
+            ],
+        },
+    ],
+});
+
+/**
+ * `PositiveSlippageFeeTransformer.TransformData`
+ */
+export interface PositiveSlippageFeeTransformerData {
+    fee: {
```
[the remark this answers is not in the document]
romainbutteaud: fixed it

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -532,8 +532,8 @@ export class MarketOperationUtils {
         };
 
         // NOTE: For sell quotes input is the taker asset and for buy quotes input is the maker asset
-        const takerAssetToEthRate = side === MarketOperation.Sell ? ethToInputRate : ethToOutputRate;
-        const makerAssetToEthRate = side === MarketOperation.Sell ? ethToOutputRate : ethToInputRate;
```
romainbutteaud, line 536: @kimpers Kim, I think you named these variables, does it make sense to rename them? I think it should be `ethToXAssetRate`, correct?

dorothy-zbornak (approved):
yay!! great job on this

packages/asset-swapper/src/utils/market_operation_utils/index.ts
kimpers: There were a lot of discussions back and forth about whether `ethToXAssetRate` or `XAssetToEthRate` is the correct naming 😅  I'm open to changing it but maybe we should just have a vote to get a consensus on what makes the most sense? 🤔 

Another thing to consider is that we now have `sellTokenToEthRate` and `buyTokenToEthRate` in the quote response, so if we change this we may or may not want to change that as well (could have implications for integrators if anyone is relying on that value).

dorothy-zbornak: I don't see how the original naming was correct. E.g., in a sell:
- `taker == input`
- `maker == output`
- so `ethToInputRate == ETH -> taker`, not `taker -> ETH`
??

kimpers: > I don't see how the original naming was correct. E.g., in a sell:
> 
> * `taker == input`
> * `maker == output`
> * so `ethToInputRate == ETH -> taker`, not `taker -> ETH`
>   ??

I dug into the old relevant PRs (for ref 0xProject/protocol#49 and 0xProject/0x-api/pull/433).

The problem here is (and correct me if I'm wrong) that `ethToInputRate` and `ethToOutputRate` are named in a funky way ([See discussions](https://github.com/0xProject/0x-api/pull/433#discussion_r534613501), and it should have been cleaned up but wasn't.

Let's use an example sell: 1000 USDC for ETH: https://api.0x.org/swap/v1/quote?sellAmount=1000000000&sellToken=USDC&buyToken=ETH.

In this case `sellTokenToEthRate   == takerAssetToEthRate == ethToInputRate`, the value is `1350.879851 USDC`. So `ethToInputRate` is actually the rate of for going takerToken to ETH. Or am I missing something? 🤔

dorothy-zbornak: I think this is a case where two wrongs make a right. Perhaps the computation in API undoes this weirdness (so that would break), but `ethToInputRate` and `ethToOutputRate` appear to be getting populated correctly earlier on from what I can see.

```ts
            // Get ETH -> makerToken token price.
            this._sampler.getMedianSellRate(feeSourceFilters.sources, makerToken, this._wethAddress, ONE_ETHER),
            // Get ETH -> taker token price.
            this._sampler.getMedianSellRate(feeSourceFilters.sources, takerToken, this._wethAddress, ONE_ETHER),
...
    public getMedianSellRate(
        sources: ERC20BridgeSource[],
        makerToken: string,
        takerToken: string,
        takerFillAmount: BigNumber,
    ): BatchedOperation<BigNumber> {

```

kimpers: We had a long discussion about this after Thursday's daily and came to the conclusion that both namings are in fact correct they just have a different meaning of the word `to`.

In `ethToInputRate` the `to` implies a rate to multiply with, in `takerAssetToEthRate` the `to` implies takerAsset over eth. The conclusion is that we should clean this up and rename the variables in a less ambiguous way 👍

dorothy-zbornak: imo `to` is always multiply

romainbutteaud: Thanks Kim, @dorothy-zbornak we want to rename these on the asset-swapper side to `takerAssetPriceForOneEth` to remove all ambiguity

xianny (changes requested):
I think we wanted to change these too, if I remember our discussion correctly: https://github.com/0xProject/protocol/pull/101/commits/592e5035483e463162c77e7b0407c744422127ba#diff-2d9ac686bf841b06dd5b6bd4d1b37a7238116fbdf6940a0a12097bdf8deea301L375-R376

packages/asset-swapper/src/types.ts
```diff
@@ -205,8 +205,8 @@ export interface SwapQuoteBase {
     isTwoHop: boolean;
     makerTokenDecimals: number;
     takerTokenDecimals: number;
-    ethToTakerAssetRate: BigNumber;
-    ethToMakerAssetRate: BigNumber;
+    takerAssetPriceForOneEth: BigNumber;
+    makerAssetPriceForOneEth: BigNumber;
```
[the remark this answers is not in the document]
dorothy-zbornak: I think these are actually the start of the problem. Can we fix it on API side?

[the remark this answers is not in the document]
romainbutteaud: I think @kimpers wanted to change these too right? I changed them in API too, see https://github.com/0xProject/0x-api/pull/453/files#diff-00fddc728f792d80847a0c68dac37537408b6fbf239c80b80df46a19b3a739fcR274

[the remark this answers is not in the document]
kimpers: Yeah, changing these is perfect 👍  The API PR keeps the quote response fields as `sellTokenToEthRate` and `buyTokenToEthRate`

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -539,8 +539,8 @@ export class MarketOperationUtils {
         };
 
         // NOTE: For sell quotes input is the taker asset and for buy quotes input is the maker asset
-        const ethToTakerAssetRate = side === MarketOperation.Sell ? ethToInputRate : ethToOutputRate;
-        const ethToMakerAssetRate = side === MarketOperation.Sell ? ethToOutputRate : ethToInputRate;
+        const takerAssetPriceForOneEth = side === MarketOperation.Sell ? ethToInputRate : ethToOutputRate;
```
kimpers, line 542: Did we want to update the naming of `ethToInputRate` and `ethToOutputRate` as well? I'm a bit fuzzy on what we decided upon 🤔

romainbutteaud: @kimpers 😂 yeah, I mean I like that new naming, it does make things clearer to me. So ok for `inputPriceForOneEth` and `outputPriceForOneEth`?

romainbutteaud: @dorothy-zbornak @xianny @dekz

packages/asset-swapper/src/types.ts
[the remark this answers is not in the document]
romainbutteaud: correct!

packages/asset-swapper/src/utils/market_operation_utils/index.ts
dorothy-zbornak: Honestly I am even more confused by these new names. "Price" is weird to me here because it is usually expressed in a different, base token. Why not just `inputTokensPerEth`, etc.

PR #108, opened 2021-01-06
fix: Upgrade quote server and Asset Swapper types (and specify protocol ve…

PirosB3:
- Upgrades Asset Swapper to the latest `@0x/quote-server`
- Updates all types to specify protocol version (in case format changes)
- Passes the `protocolVersion` as an explicit RFQ request parameter


alexkroeger (approved):
LGTM. I imagine most MMs rn will just ignore the new `protocolVersion` request param rn.

Before merging I would try an RFQ request with the `protocolVersion` param with the largest MMs.

PR #115, opened 2021-01-16
Add test for weird order issue

Recmo:
## Checklist:


package.json
```diff
@@ -82,5 +82,6 @@
     },
     "resolutions": {
         "merkle-patricia-tree": "^2.3.2"
-    }
+    },
+    "version": "0.0.0"
```
Recmo, line 86: I have no idea where these changes and the `v0.0.0` commit come from. Please ignore. :D

PR #140, opened 2021-02-10
MultiplexFeature and BatchFillNativeOrdersFeature

mzhu25:
## Checklist:


contracts/zero-ex/contracts/src/features/NativeOrdersFeature.sol
```diff
@@ -346,7 +345,6 @@ contract NativeOrdersFeature is
                 taker: taker,
                 sender: sender
             }));
-        _refundExcessProtocolFeeToSender(results.ethProtocolFeePaid);
```
mzhu25, line 349: This is a no-op if `msg.sender == address(this)`, and this function is `onlySelf`

contracts/zero-ex/contracts/src/features/TransformERC20Feature.sol
```diff
@@ -313,7 +313,7 @@ contract TransformERC20Feature is
             to.transfer(msg.value);
         }
         // Transfer input tokens.
-        if (!LibERC20Transformer.isTokenETH(inputToken)) {
+        if (!LibERC20Transformer.isTokenETH(inputToken) && amount != 0) {
```
mzhu25, line 316: This allows for a "transfer then execute" pattern for `_transformERC20`

contracts/zero-ex/test/features/multiplex_test.ts
```diff
@@ -0,0 +1,771 @@
...
+                                      /––––PLP–––> USDC
+                                     /                \
+                                    /                 PLP
+                                   /––Uni (via USDC)–––\
+                                  /                    V
+                ETH ––wrap––> WETH ––––Uni/Sushi–––> USDT ––Sushi––> LON
+                                \                                      ^
+                                 ––––––––––––––– Uni –––––––––––––––––/
+            */
+            // Taker has to have approved the EP for the intermediate tokens :/
```
mzhu25, line 668: We could avoid this if we allow the EP itself to hold the intermediate tokens, but this may cause some wonky behavior... alternatively, we can write a Transformer that calls the Multiplex feature, at the cost of some efficiency

contracts/zero-ex/contracts/src/features/NativeOrdersFeature.sol
mzhu25: this was the only functional change to this feature, so no need to redeploy

contracts/zero-ex/contracts/src/features/TransformERC20Feature.sol
mzhu25: not urgent to redeploy this feature either; I doubt 0x API will be making use of this any time soon

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
```diff
@@ -0,0 +1,799 @@
...
+                args.outputToken = IERC20TokenV06(fillData.tokens[i + 1]);
+                args.minOutputTokenAmount = 0;
+                args.taker = payable(_computeHopRecipient(fillData.calls, i));
+                if (nextTarget != address(0)) {
+                    // If `nextTarget` was set in the previous hop, then the input
+                    // token was already sent to the FlashWallet. Setting
+                    // `inputTokenAmount` to 0 indicates that no tokens need to
+                    // be pulled into the FlashWallet before executing the
+                    // transformations.
+                    args.inputTokenAmount = 0;
```
dorothy-zbornak, line 474: tricksy hobbitses

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
```diff
@@ -0,0 +1,799 @@
...
+                    ethValue,
+                    totalEth,
+                    remainingEth
+                );
+                try ITransformERC20Feature(address(this))._transformERC20
+                    {value: ethValue}
+                    (args)
+                    returns (uint256 outputTokenAmount_)
+                {
+                    remainingEth -= ethValue;
```
dorothy-zbornak, line 333: Would we tell FQT to transfer unspent protocol fees back to us or the taker?

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
```diff
@@ -0,0 +1,799 @@
...
+                    wrappedCall.data,
+                    (ITransformERC20Feature.Transformation[], uint256)
+                );
+                // Do not spend more than the remaining ETH.
+                ethValue = LibSafeMathV06.min256(ethValue, remainingEth);
+                // Call `_transformERC20`.
+                outputTokenAmount = ITransformERC20Feature(address(this))
+                    ._transformERC20{value: ethValue}(args);
+                // Decrement the remaining ETH.
+                remainingEth -= ethValue;
```
dorothy-zbornak, line 515: Where do unspent protocol fees ETH go?

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
```diff
@@ -0,0 +1,799 @@
...
+                        outputTokenAmount,
+                        ethValue
+                    );
+                } else {
+                    batchFillData.sellAmount = outputTokenAmount;
+                }
+                // Decrement the remaining ETH.
+                remainingEth -= ethValue;
+                uint256 leftoverEth;
+                (outputTokenAmount, leftoverEth) = _batchFill(batchFillData, ethValue);
```
dorothy-zbornak, line 541: Surprised this even works.

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
```diff
@@ -0,0 +1,799 @@
...
+                    (address[], bool)
+                );
+                recipient = _computeUniswapPairAddress(tokens[0], tokens[1], isSushi);
+            } else if (nextCall.selector == this._sellToLiquidityProvider.selector) {
+                (recipient, ,) = abi.decode(
+                    nextCall.data,
+                    (address, bytes, uint256)
+                );
+            } else if (nextCall.selector == IEtherTokenV06.withdraw.selector) {
+                recipient = address(this);
```
dorothy-zbornak, line 707: Why? If we just end up transferring to the taker anyway?

dorothy-zbornak (commented):
Still thinking about this.

contracts/zero-ex/contracts/src/features/interfaces/IMultiplexFeature.sol
```diff
@@ -0,0 +1,113 @@
...
+        // of `sellToken`, where `1e18+ == 100%` and `0.5e18 == 50%`, etc.
+        uint256 sellAmount;
+        // ABI-encoded parameters needed to perform the call.
+        bytes data;
+    }
+
+    // Parameters for `multiHopFill`.
+    struct MultiHopFillData {
+        // The sell path.
+        address[] tokens;
```
hysz, line 56: Nit - could we elaborate a bit on this struct, ex:`[input_token, hop_token_1, ..., hop_token_N, output_token]`

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
```diff
@@ -0,0 +1,799 @@
...
+        IERC20TokenV06 outputToken = IERC20TokenV06(fillData.tokens[fillData.tokens.length - 1]);
+        uint256 senderBalanceBefore = outputToken.getTokenBalanceOf(msg.sender);
+        uint256 ethBalanceBefore = address(this).balance.safeSub(msg.value);
+
+        _multiHopFill(fillData, msg.value);
+
+        // RFC: I think these balance checks are arguably unnecessary,
+        // see reasoning above.
+
+        // The `outputTokenAmount` returned by `_multiHopFill` may not
```
hysz, line 186: Same as above re: return value.

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
```diff
@@ -0,0 +1,799 @@
...
+        public
+        returns (uint256 outputTokenAmount, uint256 remainingEth)
+    {
+        // Track the remaining ETH allocated to this call.
+        remainingEth = totalEth;
+        // Track the amount of input token sold.
+        uint256 soldAmount;
+        for (uint256 i = 0; i != fillData.calls.length; i++) {
+            // Check if we've hit our target.
+            if (soldAmount >= fillData.sellAmount) { break; }
```
hysz, line 224: Is it possible we oversell? If the user signs off on selling 1 WETH then we should sell at most 1 ETH. Rounding should always favor the sender.

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
```diff
@@ -0,0 +1,799 @@
...
+            WrappedBatchCall memory wrappedCall = fillData.calls[i];
+            // Compute the fill amount.
+            uint256 inputTokenAmount = _normalizeAmount(
+                wrappedCall.sellAmount,
+                fillData.sellAmount,
+                fillData.sellAmount.safeSub(soldAmount)
+            );
+            if (wrappedCall.selector == INativeOrdersFeature._fillRfqOrder.selector) {
```
hysz, line 232: Nit - this is a really long if statement. It could help readability to split logic from each clause into their own function. I imagine this would not affect efficiency since it's just jump.

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
```diff
@@ -0,0 +1,799 @@
...
+                // Decode the RFQ order and signature.
+                (
+                    LibNativeOrder.RfqOrder memory order,
+                    LibSignature.Signature memory signature
+                ) = abi.decode(
+                    wrappedCall.data,
+                    (LibNativeOrder.RfqOrder, LibSignature.Signature)
+                );
+                if (order.expiry <= uint64(block.timestamp)) {
```
hysz, line 241: +1

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
```diff
@@ -0,0 +1,799 @@
...
+                wrappedCall.sellAmount,
+                fillData.sellAmount,
+                fillData.sellAmount.safeSub(soldAmount)
+            );
+            if (wrappedCall.selector == INativeOrdersFeature._fillRfqOrder.selector) {
+                // Decode the RFQ order and signature.
+                (
+                    LibNativeOrder.RfqOrder memory order,
+                    LibSignature.Signature memory signature
+                ) = abi.decode(
```
hysz, line 237: I imagine this will get pretty expensive. For each RFQ order we decode / encode / decode. I think it would save a lot of gas if we just had the single decode in NativeOrdersFeature. 

Manual memory management does come with risks and added complexity. Thinking a bit on this. One way to sidestep could be encoding `wrappedCall.data` as the input to the NativeOrdersFeature: `(Order, Signature, Amount, Sender)`. Then we simply fill in the Amount/Sender here execute the call.

hysz (commented):
Partial review. Still reading the multiple feature.

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
```diff
@@ -0,0 +1,799 @@
...
+                outputTokenAmount = outputTokenAmount.safeAdd(outputTokenAmount_);
+                // Add back any ETH that wasn't used by the nested multi-hop fill.
+                remainingEth += leftoverEth;
+            } else {
+                revert("MultiplexFeature::_batchFill/UNRECOGNIZED_SELECTOR");
+            }
+        }
+    }
+
+    function _multiHopFill(MultiHopFillData memory fillData, uint256 totalEth)
```
hysz, line 370: Function comment pls 🙏

contracts/zero-ex/contracts/src/features/BatchFillNativeOrdersFeature.sol
```diff
@@ -0,0 +1,477 @@
...
+    /// @dev Fills the given limit orders by selling up to the given
+    ///      amount of taker token.
+    /// @param orders Array of limit orders.
+    /// @param signatures Array of signatures corresponding to each order.
+    /// @param takerTokenSellAmount The total amount of taker token to sell.
+    /// @param revertIfIncomplete If true, reverts if this function fails to
+    ///        fill the full sell amount.
+    /// @return takerTokenSoldAmount The total amount of taker token sold.
+    /// @return makerTokenBoughtAmount The total amount of maker token bought.
+    function marketSellLimitOrders(
```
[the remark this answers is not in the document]
hysz: I confirmed with Maker / Taker teams that there is no demand for these functions.

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
```diff
@@ -0,0 +1,799 @@
...
+                            msg.sender
+                        )
+                    returns (uint128 takerTokenFilledAmount, uint128 makerTokenFilledAmount)
+                {
+                    // Increment the sold and bought amounts.
+                    soldAmount = soldAmount.safeAdd(takerTokenFilledAmount);
+                    outputTokenAmount = outputTokenAmount.safeAdd(makerTokenFilledAmount);
+                } catch {}
+            } else if (wrappedCall.selector == this._sellToUniswap.selector) {
+                (address[] memory tokens, bool isSushi) = abi.decode(
```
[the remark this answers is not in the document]
mzhu25: yep

contracts/zero-ex/contracts/src/features/native_orders/NativeOrdersFeature.sol
```diff
@@ -0,0 +1,94 @@
+// SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
mzhu25: ye I'm down

contracts/zero-ex/contracts/src/features/BatchFillNativeOrdersFeature.sol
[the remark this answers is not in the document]
mzhu25: deleted

contracts/zero-ex/contracts/src/features/BatchFillNativeOrdersFeature.sol
```diff
@@ -0,0 +1,477 @@
...
+        external
+        payable
+        override
+        returns (
+            uint128 takerTokenSoldAmount,
+            uint128 makerTokenBoughtAmount
+        )
+    {
+        uint256 protocolFee = uint256(INativeOrdersFeature(address(this)).getProtocolFeeMultiplier())
+            .safeMul(tx.gasprice);
```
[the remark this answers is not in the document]
mzhu25: after deleting the market functions, there is no longer code duplication so leaving as is

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
mzhu25: back to the taker

contracts/zero-ex/contracts/src/features/BatchFillNativeOrdersFeature.sol
```diff
@@ -0,0 +1,477 @@
...
+    /// @param takerTokenFillAmounts Array of desired amounts to fill each order.
+    /// @param revertIfIncomplete If true, reverts if this function fails to
+    ///        fill the full fill amount for any individual order.
+    /// @return takerTokenFilledAmounts Array of amounts filled, in taker token.
+    /// @return makerTokenFilledAmounts Array of amounts filled, in maker token.
+    function batchFillLimitOrders(
+        LibNativeOrder.LimitOrder[] calldata orders,
+        LibSignature.Signature[] calldata signatures,
+        uint128[] calldata takerTokenFillAmounts,
+        bool revertIfIncomplete
```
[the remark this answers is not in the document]
mzhu25: `revertIfIncomplete` is more plainly understandable, imo 🤷

contracts/zero-ex/contracts/src/features/interfaces/INativeOrdersEvents.sol
```diff
@@ -0,0 +1,116 @@
...
+pragma solidity ^0.6.5;
+pragma experimental ABIEncoderV2;
+
+import "@0x/contracts-erc20/contracts/src/v06/IERC20TokenV06.sol";
+import "../libs/LibSignature.sol";
+import "../libs/LibNativeOrder.sol";
+
+
+/// @dev Events emitted by NativeOrdersFeature.
+interface INativeOrdersEvents {
```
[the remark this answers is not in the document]
mzhu25: nope, just a refactor

contracts/zero-ex/contracts/src/features/native_orders/NativeOrdersFeature.sol
```diff
@@ -0,0 +1,94 @@
...
+/*
```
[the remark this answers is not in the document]
mzhu25: just a refactor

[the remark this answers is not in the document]
mzhu25: I originally refactored it because I was going to add the batch functions to the NativeOrdersFeature itself, rather than as a separate feature

contracts/zero-ex/contracts/src/features/BatchFillNativeOrdersFeature.sol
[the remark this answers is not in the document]
hysz: That's cool. For consistency, wdyt of renaming `FillOrKillFailedError` to `BatchFillIncompleteError`?

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
mzhu25: back to the taker –– the actual taker, which may differ from `args.taker` here, so we'd set the FQT `refundReceiver` to the actual taker's address rather than `address(1)`

contracts/zero-ex/contracts/src/features/BatchFillNativeOrdersFeature.sol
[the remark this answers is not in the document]
hysz: Tying the param to the corresponding error I think would be helpful for clients.

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
```diff
@@ -0,0 +1,799 @@
...
+        ILiquidityProvider provider,
+        address recipient,
+        bytes memory auxiliaryData
+    )
+        public
+        returns (uint256 outputTokenAmount)
+    {
+        uint256 balanceBefore = IERC20TokenV06(outputToken).getTokenBalanceOf(recipient);
+        if (IERC20TokenV06(inputToken).isTokenETH()) {
+            sandbox.executeSellEthForToken(
```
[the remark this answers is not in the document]
mzhu25: I think it's kind of unnecessary; API doesn't apply a fallback for PLP so we'd be pretty brittle to reverts here anyway.

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
```diff
@@ -0,0 +1,799 @@
...
+            // specify a fraction of `totalBalance`.
+            return LibSafeMathV06.min256(
+                totalBalance
+                    * LibSafeMathV06.min256(rawAmount & LOWER_255_BITS, 1e18)
+                    / 1e18,
+                remainingBalance
+            );
+        }
+        return LibSafeMathV06.min256(rawAmount, remainingBalance);
+    }
```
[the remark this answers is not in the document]
mzhu25: the logic is slightly different than the function in FQT, so I'm going to leave as is for now... I think in the future we will want FQT to use this logic though

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
mzhu25: this is the case where the _next_ call is unwrapping WETH, so we want the current call to transfer the WETH to the EP.

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
hysz: Looks like the gas savings are already substantial so probably not worth further optimizing at this time. But, something to consider going forward.

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
```diff
@@ -0,0 +1,799 @@
...
+
+        // RFC: I think these balance checks are arguably unnecessary,
+        // and that we can instead trust the `outputTokenAmount` and
+        // `remainingEth` values returned by `_batchFill`. AFAICT the
+        // `remainingEth` value should always be accurate, and the
+        // `outputTokenAmount` is only inaccurate if the sender has
+        // an allowance set on a malicious token contract or liquidity
+        // provider.
+
+        // The `outputTokenAmount` returned by `_batchFill` may not
```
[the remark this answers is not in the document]
mzhu25: The `outputTokenAmount` returned by `_batchFill` is used if its nested in a `_multiHopFill` though

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
mzhu25: I don't think it's possible to oversell (barring the taker having set an allowance on a malicious token contract or something), since the `_normalizeAmount` function clips the input token amount to the remaining sell amount. But we can revert if `soldAmount > fillData.sellAmount` I guess?

contracts/zero-ex/contracts/src/features/BatchFillNativeOrdersFeature.sol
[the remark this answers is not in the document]
mzhu25: yeah I like `BatchFillIncompleteError`

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
```diff
@@ -0,0 +1,799 @@
...
+                    nextCall.data,
+                    (address, bytes, uint256)
+                );
+            } else if (nextCall.selector == IEtherTokenV06.withdraw.selector) {
+                recipient = address(this);
+            } else if (nextCall.selector == ITransformERC20Feature._transformERC20.selector) {
+                recipient = address(
+                    ITransformERC20Feature(address(this)).getTransformWallet()
+                );
+            }
```
hysz, line 712: Can anything go wrong if `nextCall.selector` is not one of the listed selectors? Should we revert?

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
```diff
@@ -0,0 +1,799 @@
...
+                uint256 ethValue = LibSafeMathV06.min256(outputTokenAmount, remainingEth);
+                // Wrap ETH.
+                weth.deposit{value: ethValue}();
+                nextTarget = _computeHopRecipient(fillData.calls, i);
+                weth.transfer(nextTarget, ethValue);
+                remainingEth -= ethValue;
+            } else if (wrappedCall.selector == IEtherTokenV06.withdraw.selector) {
+                // Unwrap WETH and send to `msg.sender`.
+                weth.withdraw(outputTokenAmount);
+                msg.sender.transfer(outputTokenAmount);
```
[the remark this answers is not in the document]
mzhu25: hypothetically, yes, but it makes ETH accounting even more of a nightmare 😩

contracts/zero-ex/contracts/src/features/interfaces/IMultiplexFeature.sol
```diff
@@ -0,0 +1,113 @@
...
+
+    // Represents a call nested within a `batchFill`.
+    struct WrappedBatchCall {
+        // The selector of the function to call.
+        bytes4 selector;
+        // Amount of `inputToken` to sell.
+        // Setting the high-bit indicates that `sellAmount & LOW_BITS`
+        // should be treated as a `1e18` fraction of the current balance
+        // of `sellToken`, where `1e18+ == 100%` and `0.5e18 == 50%`, etc.
+        uint256 sellAmount;
```
hysz, line 48: I don't feel too strongly, but given that this is client-facing I might lean towards splitting into two variables.

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
```diff
@@ -0,0 +1,799 @@
...
+    function _computeHopRecipient(
+        WrappedMultiHopCall[] memory calls,
+        uint256 i
+    )
+        private
+        view
+        returns (address recipient)
+    {
+        recipient = msg.sender;
+        if (i != calls.length - 1) {
```
hysz, line 693: Since this function is independent from the iterator, `i`, I would probably strictly enforce `i < calls.length - 1`

hysz, line 693: Given that the iterator, `i`, is computed elsewhere I might strictly enforce that `i < calls.length`.

hysz: That said, new versions of Solidity may revert on invalid indices for `calls`. Not totally sure.

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
mzhu25: hm I guess we wouldn't be able to do that if the taker address isn't passed to the API, but...
let's just not fill limit orders from within a multihop fill 🤝

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
mzhu25: it would revert in `_multiHopFill`

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
hysz, line 541: Above there was a comment around how this value is not always accurate. Could that pose an issue here?

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
```diff
@@ -0,0 +1,799 @@
...
+        bool isSushi,
+        address pairAddress,
+        address recipient
+    )
+        public
+        returns (uint256 outputTokenAmount)
+    {
+        require(tokens.length > 1, "MultiplexFeature::_sellToUniswap/InvalidTokensLength");
+
+        if (pairAddress == address(0)) {
```
hysz, line 583: Are there cases when `pairAddress != address(0)`?

mzhu25: yes, in `_multiHopFill` the pair address for the next hop may be pre-computed and stored in the `nextTarget` variable. This is for a scenario like a PLP => Uniswap multihop fill, where we want the PLP to transfer the bought tokens directly into the Uniswap pair contract

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
mzhu25: afaict `outputTokenAmount` is only ever inaccurate if the taker has set an allowance on a malicious token contract, and that contract is called somewhere in the `_batchFill`. This malicious contract could pull tokens from the taker such that the outputTokenAmount returned by e.g. `_sellToUniswap` is inaccurate. But this is the same risk profile as the UniswapFeature, so I think it's reasonable to rely on the returned value here.

contracts/zero-ex/contracts/src/features/BatchFillNativeOrdersFeature.sol
```diff
@@ -0,0 +1,477 @@
...
+                        takerTokenFillAmounts[i],
+                        msg.sender,
+                        msg.sender
+                    )
+                returns (uint128 takerTokenFilledAmount, uint128 makerTokenFilledAmount)
+            {
+                // Update amounts filled.
+                (takerTokenFilledAmounts[i], makerTokenFilledAmounts[i]) =
+                    (takerTokenFilledAmount, makerTokenFilledAmount);
+                ethProtocolFeePaid = ethProtocolFeePaid.safeAdd(protocolFee);
```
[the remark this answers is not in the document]
mzhu25: API won't be using this function, so not worth IMO 😬

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
mzhu25: will write up my understanding of the security model in the audit doc

contracts/zero-ex/contracts/src/features/interfaces/IMultiplexFeature.sol
mzhu25: leaving as is for consistency with FQT

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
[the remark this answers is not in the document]
dorothy-zbornak: probably missing something but is something like this here enough?
```solidity
if (nextTarget != msg.sender) {
  remainingEth += outputTokenAmount;
}
```

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
dorothy-zbornak: oh gotcha

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
```diff
@@ -0,0 +1,799 @@
...
+                (outputTokenAmount, leftoverEth) = _batchFill(batchFillData, ethValue);
+                // Add back any ETH that was unused by the `_batchFill`.
+                remainingEth += leftoverEth;
+                nextTarget = address(0);
+            } else if (wrappedCall.selector == IEtherTokenV06.deposit.selector) {
+                uint256 ethValue = LibSafeMathV06.min256(outputTokenAmount, remainingEth);
+                // Wrap ETH.
+                weth.deposit{value: ethValue}();
+                nextTarget = _computeHopRecipient(fillData.calls, i);
+                weth.transfer(nextTarget, ethValue);
```
hysz, line 519: Can you think of any risks that may arise if `nextTarget` re-enters the EP?

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
hysz: Ah, I see. Should we perform any validation ~ is it an issue if the tokens are transferred somewhere other than a Uniswap pool?

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
hysz: Oh, as well is this the standard now?
```
(bool success,) = msg
                .sender
                .call{value: ethValue}("");
```

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
```diff
@@ -0,0 +1,799 @@
...
+        );
+
+        uint256 ethBalanceAfter = address(this).balance;
+        require(
+            ethBalanceAfter >= ethBalanceBefore,
+            "MultiplexFeature::multiHopFill/OVERSPENT_ETH"
+        );
+        // Refund ETH
+        if (ethBalanceAfter > ethBalanceBefore) {
+            payable(msg.sender).transfer(ethBalanceAfter - ethBalanceBefore);
```
hysz, line 185: Same question ~ any difference now between `transfer` and `call{value: x}("");`?

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
hysz: Yeah, actually it could be good to return `sellAmount` and then assert `sellAmount <= fillData.sellAmount` in the entry functions, where we validate `boughtAmount >= minBoughtAmount`.

hysz: Or, possibly even using a balance check. In that case, so long as the user is not transferring out intermediate tokens, audting the security would be very simple.

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
mzhu25: `nextTarget` can only be set to one of the following:
- address(0)
- msg.sender
- a Uniswap/Sushiswap pair contract
- a LiquidityProvider contract
- the FlashWallet
- the EP
The only case in which `nextTarget` is called directly is if it's a Uniswap or Sushiswap pair contract, which is computed using their respective factories and init code hashes. If it's a user-supplied LiquidityProvider contract, it's called using the sandbox as a proxy. It's only set to the EP if the next hop is a WETH unwrap.

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
mzhu25: yeah the call syntax is good practice 👍

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
mzhu25: that should never happen, since `nextTarget` would have been set to the Uniswap pair address as computed using `create2` in the previous hop. At worst the pair contract doesn't exist, but it can't go anywhere unexpected

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
mzhu25: idk, I'm still pretty convinced that even the output token balance check is unnecessary. If we look at all of the functions this feature can call, we should be able to trust the output token amount returned by all of them:
- `_sellToUniswap` we compute the output token amount. This is what we do in the UniswapFeature as well, where we do not perform any balance checks. ✅ 
- `_sellToLiquidityProvider` we use balance checks to compute the output token amount. ✅ 
- `_fillRfqOrder` has the same guarantees as filling an RFQ order through the NativeOrdersFeature ✅ 
- `_transformERC20` performs balance checks for the output token amount ✅ 
- WETH withdraw/deposit is obvious ✅ 
- By induction, we should be able to trust the output token amount of a nested `_multiHopFill` as well ✅

mzhu25: Of course, if the user has set an allowance on some malicious token contract that gets called in this feature, all bets are off. But this is also the case for all our other features. So I don't think we'd be introducing any new security assumptions by not performing balance checks on the input token (or even the output token)

dorothy-zbornak: @mzhu25 I agree with you but I think the optics and audits will go better with the check in place since very few people would be able to prove it in this manner. It also protects us if we violate these assumptions by updating a feature or w/e.

mzhu25: I'm fine with leaving the output token balance check in, but I think checking balances for the input token or intermediate tokens would be overkill

contracts/zero-ex/contracts/src/features/interfaces/IMultiplexFeature.sol
```diff
@@ -75,6 +75,22 @@ interface IMultiplexFeature {
         address recipient
     );
 
+    event MultiplexBatchFill(
+        address indexed taker,
```
dorothy-zbornak, line 79: Probably no need for indexing anything.

contracts/zero-ex/contracts/src/features/MultiplexFeature.sol
dorothy-zbornak: Could just have the input token check as part of a test to catch regressions.

hysz (approved):
Nice work! Let's get those gas savings 🥇

dorothy-zbornak (approved):
Looks good. Regen artifacts/wrappers before merging?

PR #146, opened 2021-02-17
fix: Protocol fee in fee schedule

dekz:
## Description

Fixes the gas schedule override into fee schedule setting all Native to include a protocol fee.

## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/market_operation_utils/constants.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/constants.ts
@@ -1,7 +1,6 @@
 import { FillQuoteTransformerOrderType } from '@0x/protocol-utils';
 import { BigNumber } from '@0x/utils';
 
-import { SignedNativeOrder } from '../../types';
 import { TokenAdjacencyGraphBuilder } from '../token_adjacency_graph_builder';
 
 import { SourceFilters } from './source_filters';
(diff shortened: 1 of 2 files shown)


packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -501,9 +503,11 @@ export const BALANCER_MAX_POOLS_FETCHED = 3;
...
 // tslint:disable:custom-no-magic-numbers
 export const DEFAULT_GAS_SCHEDULE: Required<FeeSchedule> = {
-    [ERC20BridgeSource.Native]: _fillData => {
-        // const nativeFillData = (_fillData as NativeRfqOrderFillData|NativeLimitOrderFillData)
-        return 100e3;
+    [ERC20BridgeSource.Native]: fillData => {
+        const nativeFillData = fillData as SignedNativeOrder;
+        return nativeFillData && nativeFillData.type === FillQuoteTransformerOrderType.Limit
+            ? PROTOCOL_FEE_MULTIPLIER.plus(100e3).toNumber()
+            : 100e3;
```
alexkroeger, line 510: should be 80 to reflect lower cost of RFQ?

alexkroeger: synced over slack, should be accurate for RFQ, might leave a note that the limit order cost might need revisiting

alexkroeger (approved):

PR #155, opened 2021-02-23
FQT deploy + no gas limit ETH refund

dorothy-zbornak:
## Description

- Update *future* FQT address.
- Remove the gas limit on ETH refund in FQT so we can potentially account for it in simbot when limit orders are involved. Still won't work though because I don't think we actually set `refundReceiver` in API right now, but one day :roll_eyes:.

### TODO
- [x] deploy FQT

## Checklist:

### DIFF
--- a/packages/contract-addresses/CHANGELOG.json
+++ b/packages/contract-addresses/CHANGELOG.json
@@ -4,6 +4,15 @@
         "changes": [
             {
                 "note": "Deploy new FQT",
+                "pr": 155
+            }
+        ]
+    },
+    {
+        "version": "5.10.0",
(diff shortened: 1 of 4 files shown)


packages/contract-addresses/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "5.10.0",
```
mzhu25, line 3:
```suggestion
        "version": "5.11.0",
```

mzhu25 (approved):

dekz (approved):

PR #158, opened 2021-02-25
improve logging for alt rfq request

alexkroeger:
## Description

More helpful logging for alt RFQ requests (approximating logs for the standard RFQ implementation).

## Checklist:


packages/asset-swapper/src/utils/alt_mm_implementation_utils.ts
```diff
@@ -211,13 +215,43 @@ export async function returnQuoteFromAltMMAsync<ResponseT>(
...
-        headers: { Authorization: `Bearer ${apiKey}` },
-        timeout: maxResponseTimeMs,
-    });
+    // let response: AxiosResponse<any>;
+    const response = await axiosInstance
+        .post(`${url}/quotes`, data, {
+            headers: { Authorization: `Bearer ${apiKey}` },
+            timeout: maxResponseTimeMs,
+        })
+        .then(resp => {
```
BMillman19, line 224: Not sure if we need this `then`, can't we just call `.catch` right after `.post` ?

alexkroeger: nope 👍

BMillman19 (approved):

packages/asset-swapper/src/utils/alt_mm_implementation_utils.ts
```diff
@@ -211,13 +215,40 @@ export async function returnQuoteFromAltMMAsync<ResponseT>(
         }
     }
 
-    const response = await axiosInstance.post(`${url}/quotes`, data, {
-        headers: { Authorization: `Bearer ${apiKey}` },
-        timeout: maxResponseTimeMs,
-    });
+    // let response: AxiosResponse<any>;
```
phil-ociraptor, line 218: nit: looks like commented code

packages/asset-swapper/src/utils/alt_mm_implementation_utils.ts
```diff
@@ -12,8 +12,11 @@ import {
     AltQuoteRequestData,
     AltQuoteSide,
     AltRfqtMakerAssetOfferings,
+    LogFunction,
 } from '../types';
 
+const SUCCESS_CODE = 201;
```
phil-ociraptor, line 18: Would be nice if our `StatusCodes` package was more robust...

phil-ociraptor (approved):
Looks good! Just a couple minor nit comments

packages/asset-swapper/src/utils/alt_mm_implementation_utils.ts
```diff
@@ -211,13 +215,40 @@ export async function returnQuoteFromAltMMAsync<ResponseT>(
...
+        };
+    }
+    // successful handling but no quote is indicated by status = 'rejected'
     if (response.data.status === 'rejected') {
-        throw new Error('alt MM rejected quote');
+        return {
+            data: (emptyResponse as unknown) as ResponseT,
+            // hack: set the http status to 204 no content so we can more
+            // easily track when no quote is returned
+            status: 204,
```
phil-ociraptor, line 250: A small thing - I do think the hack will be easy to trip on in the future. If possible, could we add a log statement aboe this, similar to what we have for the if case above?

packages/asset-swapper/src/utils/alt_mm_implementation_utils.ts
PirosB3: Great idea, isn't this code in our `StatusCodes` package?

packages/asset-swapper/src/utils/alt_mm_implementation_utils.ts
```diff
@@ -211,13 +215,40 @@ export async function returnQuoteFromAltMMAsync<ResponseT>(
...
-        headers: { Authorization: `Bearer ${apiKey}` },
-        timeout: maxResponseTimeMs,
-    });
+    // let response: AxiosResponse<any>;
+    const response = await axiosInstance
+        .post(`${url}/quotes`, data, {
+            headers: { Authorization: `Bearer ${apiKey}` },
+            timeout: maxResponseTimeMs,
+        })
+        .catch(err => {
```
PirosB3, line 224: why not use `try/catch`? if you `await` a response, a failed promise will bubble up as a thrown error

packages/asset-swapper/src/utils/alt_mm_implementation_utils.ts
```diff
@@ -211,13 +215,40 @@ export async function returnQuoteFromAltMMAsync<ResponseT>(
...
 
+    // tslint:disable-next-line:custom-no-magic-numbers
+    if (response.status !== SUCCESS_CODE) {
+        const rejectedRequestInfo = {
+            status: response.status,
+            message: response.data,
+        };
+        warningLogger(rejectedRequestInfo, `Alt RFQ MM did not return a status of ${SUCCESS_CODE}`);
+        return {
+            data: (emptyResponse as unknown) as ResponseT,
```
PirosB3, line 240: Why did you define `emptyResponse` here? can't you just do `data: {}`? both here and on L247?

PirosB3 (changes requested):

packages/asset-swapper/src/utils/alt_mm_implementation_utils.ts
alexkroeger: 👍

packages/asset-swapper/src/utils/alt_mm_implementation_utils.ts
alexkroeger: I don't think it is

packages/asset-swapper/src/utils/alt_mm_implementation_utils.ts
alexkroeger: I believe this is to appease typescript

packages/asset-swapper/src/utils/alt_mm_implementation_utils.ts
alexkroeger: is there any benefit to doing so?

packages/asset-swapper/src/utils/alt_mm_implementation_utils.ts
alexkroeger: no type assertion on object literals

packages/asset-swapper/src/utils/alt_mm_implementation_utils.ts
alexkroeger: It's not

packages/asset-swapper/src/utils/alt_mm_implementation_utils.ts
alexkroeger: good idea! added a warning log

PirosB3 (approved):

PR #162, opened 2021-03-02
FQT: Pack Protocol/source name into source ID

dorothy-zbornak:
## Description

I'm going to target this at the `bsc` branch for now because we need to give notice to data and 0xtracker if we put it on prod.

We now pack the protocol ID and the source name into a single bytes32 like:
```solidity
// right-aligned protocol ID in upper 16 bytes
bytes32(uint128(PROTOCOL_ID) << 128)
  // left-aligned ASCII name in lower 16 bytes
  | bytes32(uint256(uint128(bytes16(ASCII_SOURCE_NAME))))
```

The `BridgeFill` event now uses a `bytes32` instead of a `uint256` for the `source` param, so it's a different signature/topic0. I figure it's safer this way?

Benefits:
- The protocol ID is used to identify the compatible protocol for a source (e.g., Swerve uses Curve protocol). This means we don't have to keep updating and redeploying the FQT every time want to roll up a new fork of an existing protocol.
- The ASCII source name is read by the [subgraph](https://thegraph.com/explorer/subgraph/0xproject/exchange-v4). Storing a readable string means the subgraph does not have to be constantly updated and resynced when a new source is integrated. Probably helps other data pipelines as well.

## Checklist:


contracts/zero-ex/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "0.20.0-bsc",
```
dorothy-zbornak, line 3: So this is not gonna work with our current publish job. Immediate TODO is to update the job/scripts to allow for publishing named forks. The alternative is to gitpkg everything (yuck).

hysz (approved):
LGTM!

contracts/zero-ex/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "0.21.0-bsc.0",
```
dorothy-zbornak, line 3: To publish the `bsc` branch:
- Go to GH actions
- Click `publish` workflow
- Enter `branch: bsc`, `prerelease: bsc`
- hit run and wait :crossed_fingers: 
- published packages will have versions `x.x.x-bsc.x`

dorothy-zbornak: Note that we jump ahead of non-prerelease changelog entry even if that one's not published since these changes won't get included in the main branch publish.

dorothy-zbornak: We technically don't need to add the `-bsc.0` suffx here in the changelog. Even this was just `x.x.x`, the publish script will still publish as `x.x.x-bsc.x`. But I recommend adding it here so when we eventually merge or rebase this branch with development we can tell the two changelog histories apart and don't accidentally merge two unrelated blocks.

PR #165, opened 2021-03-03
Make the proposal/quorum thresholds updatable

mzhu25:
## Description
Changes `proposalThreshold` and `quorumThreshold` from immutables to state variables than can be updated via the new `updateThresholds` function. The `updateThresholds` function can only be called by the Treasury itself, so an update would need to be encoded in a proposal. 

## Checklist:

### DIFF
--- a/contracts/treasury/CHANGELOG.json
+++ b/contracts/treasury/CHANGELOG.json
@@ -1,5 +1,14 @@
 [
     {
+        "version": "1.1.0",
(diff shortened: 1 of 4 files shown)


dorothy-zbornak (approved):
:heavy_check_mark: :heavy_check_mark:

hysz (approved):
LGTM!

PR #169, opened 2021-03-11
add tx origin blacklist to RFQ options

alexkroeger:
## Description

Add the ability to blacklist specified `txOrigin`s from using RFQT.

## Checklist:

### DIFF
--- a/packages/asset-swapper/src/constants.ts
+++ b/packages/asset-swapper/src/constants.ts
@@ -52,6 +52,7 @@ const DEFAULT_SWAP_QUOTER_OPTS: SwapQuoterOpts = {
     rfqt: {
         takerApiKeyWhitelist: [],
         makerAssetOfferings: {},
+        txOriginBlacklist: [],
     },
     tokenAdjacencyGraph: DEFAULT_TOKEN_ADJACENCY_GRAPH,
 };
--- a/packages/asset-swapper/src/constants.ts
+++ b/packages/asset-swapper/src/constants.ts
@@ -52,7 +52,7 @@ const DEFAULT_SWAP_QUOTER_OPTS: SwapQuoterOpts = {
     rfqt: {
         takerApiKeyWhitelist: [],
         makerAssetOfferings: {},
-        txOriginBlacklist: [],
+        txOriginBlacklist: new Set(),
     },
     tokenAdjacencyGraph: DEFAULT_TOKEN_ADJACENCY_GRAPH,
 };
--- a/packages/asset-swapper/src/swap_quoter.ts
+++ b/packages/asset-swapper/src/swap_quoter.ts
@@ -410,6 +410,14 @@ export class SwapQuoter {
         return whitelistedApiKeys.includes(apiKey);
     }
 
+    private _isTxOriginBlacklisted(txOrigin: string | undefined): boolean {
+        if (!txOrigin) {
+            return false;
+        }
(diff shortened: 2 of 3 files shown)


packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -410,6 +410,14 @@ export class SwapQuoter {
         return whitelistedApiKeys.includes(apiKey);
     }
 
+    private _isTxOriginBlacklisted(txOrigin: string | undefined): boolean {
+        if (!txOrigin) {
+            return false;
+        }
+        const blacklistedApiKeys = this._rfqtOptions ? this._rfqtOptions.txOriginBlacklist : [];
```
alexkroeger, line 417: should this be a set?

PirosB3: Yes, or a `Set` or a HashSet (or object, in JS)

packages/asset-swapper/src/constants.ts
```diff
@@ -52,6 +52,7 @@ const DEFAULT_SWAP_QUOTER_OPTS: SwapQuoterOpts = {
     rfqt: {
         takerApiKeyWhitelist: [],
         makerAssetOfferings: {},
+        txOriginBlacklist: [],
```
PirosB3, line 55: Could you make this a set?

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -410,6 +410,14 @@ export class SwapQuoter {
         return whitelistedApiKeys.includes(apiKey);
     }
 
+    private _isTxOriginBlacklisted(txOrigin: string | undefined): boolean {
+        if (!txOrigin) {
```
PirosB3, line 414: Are you sure that txOrigin can be undefined?

PirosB3 (approved):
Missing tests, otherwise it's awesome!

packages/asset-swapper/src/swap_quoter.ts
alexkroeger: it's explicitly checked for later in the code block

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -410,6 +410,14 @@ export class SwapQuoter {
...
+            return false;
+        }
+        const blacklistedTxOrigins = this._rfqtOptions ? this._rfqtOptions.txOriginBlacklist : new Set();
```
PirosB3, line 417: Without having to create a `new Set` every time this fn is called, can we change something like
```
if (this._rfqtOptions && this._rfqtOptions.txOriginBlacklist && this._rfqtOptions.txOriginBlacklist.has(txOrigin.toLowerCase())) {
     return true;
}
return false;
```

this is optional, if it makes sense to you

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -410,6 +410,14 @@ export class SwapQuoter {
...
+        return blacklistedTxOrigins.has(txOrigin.toLowerCase());
```
PirosB3, line 418: good catch, I had forgotten about case sensitivity

PirosB3 (approved):

packages/asset-swapper/src/swap_quoter.ts
alexkroeger: 👍

PR #184, opened 2021-03-24
Revive quote report

PirosB3:
Fixes Quote Report so that is *only* reports the following:
- All RFQ sources “considered”
- The final sources “delivered” (the order salad delivered in the quote)

This should result in < 1% off the existing logging while providing the same value as before

### DIFF
--- a/packages/asset-swapper/src/utils/quote_report_generator.ts
+++ b/packages/asset-swapper/src/utils/quote_report_generator.ts
@@ -1,4 +1,4 @@
-import { FillQuoteTransformerOrderType, Signature } from '@0x/protocol-utils';
+import { FillQuoteTransformerOrderType, RfqOrderFields, Signature } from '@0x/protocol-utils';
 import { BigNumber } from '@0x/utils';
 import _ = require('lodash');
 
@@ -44,6 +44,7 @@ export interface NativeRfqOrderQuoteReportEntry extends QuoteReportEntryBase {
     fillData: NativeFillData;
     fillableTakerAmount: BigNumber;
     isRfqt: true;
+    nativeOrder: RfqOrderFields;
     makerUri: string;
     comparisonPrice?: number;
 }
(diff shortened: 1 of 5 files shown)


packages/asset-swapper/src/utils/quote_report_generator.ts
```diff
@@ -65,25 +66,17 @@ export interface QuoteReport {
  */
 export function generateQuoteReport(
     marketOperation: MarketOperation,
-    dexQuotes: DexSample[],
-    multiHopQuotes: Array<DexSample<MultiHopFillData>>,
+    _dexQuotes: DexSample[],
```
alexkroeger, line 69: can we just get rid of these arguments: `dexQuotes` and `multiHopQuotes`?

packages/asset-swapper/test/quote_report_generator_test.ts
```diff
@@ -58,32 +58,18 @@ describe('generateQuoteReport', async () => {
     it('should generate report properly for sell', () => {
         const marketOperation: MarketOperation = MarketOperation.Sell;
 
-        const kyberSample1: DexSample = {
-            source: ERC20BridgeSource.Kyber,
-            input: new BigNumber(10000),
-            output: new BigNumber(10001),
-            fillData: {},
-        };
         const kyberSample2: DexSample = {
```
phil-ociraptor, line 61: super nit: now that `kyberSample1` is removed, can this just be named `kyberSample`. Same for uni below

phil-ociraptor (approved):
Looks good on my side! I don't entirely understand the removal of the two hop and multihop stuff, but overall, my understanding is this should reduce the amount of data that gets logged

packages/asset-swapper/src/utils/quote_report_generator.ts
```diff
@@ -44,6 +44,7 @@ export interface NativeRfqOrderQuoteReportEntry extends QuoteReportEntryBase {
     fillData: NativeFillData;
     fillableTakerAmount: BigNumber;
     isRfqt: true;
+    nativeOrder: RfqOrderFields;
```
alexkroeger, line 47: no similar adjustment needed for limit orders?

alexkroeger (approved):

PR #187, opened 2021-03-30
[MKR-3] Prepare Asset Swapper for RFQM

phil-ociraptor:
## Description

Contains a small amount of refactoring to make our existing code support both RFQT and RFQM flows. Unit tests have been updated and new coverage is added for RFQM flows.

## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/quote_requestor.ts
+++ b/packages/asset-swapper/src/utils/quote_requestor.ts
@@ -83,7 +83,7 @@ export class QuoteRequestor {
         sellTokenAddress: string, // taker token
         assetFillAmount: BigNumber,
         comparisonPrice?: BigNumber,
-        isLastLook = false,
+        isLastLook: boolean = false,
     ): TakerRequestQueryParams {
         const { buyAmountBaseUnits, sellAmountBaseUnits } =
             marketOperation === MarketOperation.Buy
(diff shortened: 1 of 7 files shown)


alexkroeger (approved):
lgtm, please fix static tests

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -236,42 +238,15 @@ export class QuoteRequestor {
...
             takerToken,
             assetFillAmount,
             marketOperation,
             comparisonPrice,
             _opts,
-            'indicative',
+            this._rfqtAssetOfferings,
         );
-
-        // validate
```
alexkroeger, line 249: good to separate out this logic, consider a `validateQuote` function (non-async)

phil-ociraptor: This is a bit trickier than I originally thought. It's straightforward to create two functions: `validateFirmQuote` and a `validateIndicativeQuote` - but gets a bit gnarly when trying to unify them into a single `validateQuote` fn. 

Thoughts on having two separate functions?

phil-ociraptor: Started down the unified path and got here: https://github.com/0xProject/protocol/commit/0255b4deb2cff4d608bab9df4e85666167da776d

But haven't found something that I'm happy with. Will look into this in a separate PR if I have time

PR #200, opened 2021-04-14
only set Last Look parameter if it's explicitly set

PirosB3:
Disables `isLastLook` parameter when not explicitly set.


phil-ociraptor (approved):
Looks good! Verified that axios will filter out `undefined` values

PR #204, opened 2021-04-19
Feat/new rfqm order

alexkroeger:
## Checklist:


contracts/zero-ex/test/features/native_orders_feature_test.ts
```diff
@@ -1569,4 +1595,310 @@ blockchainTests.resets('NativeOrdersFeature', env => {
...
+                SignatureType.EthSign,
+                taker,
+            );
+
+            await makerToken.mint(maker, order.makerAmount).awaitTransactionSuccessAsync();
+            await takerToken.mint(taker, order.takerAmount).awaitTransactionSuccessAsync();
+
+            await takerToken.mint(maker, new BigNumber(1)).awaitTransactionSuccessAsync();
+
+            logUtils.log(`maker amount`);
```
dorothy-zbornak, line 1812: Do we really need the logs in these tests?

contracts/zero-ex/test/features/native_orders_feature_test.ts
```diff
@@ -1569,4 +1595,310 @@ blockchainTests.resets('NativeOrdersFeature', env => {
             }
         });
     });
+
+    describe.only('fillTakerSignedRfqOrder', () => {
+        it('gas cost if if there is no balance is fully sold, new tokens after trade for both parties', async () => {
```
dorothy-zbornak, line 1600: Can we convert these "gas cost" tests into real unit tests?

packages/protocol-utils/src/orders.ts
```diff
@@ -316,3 +325,83 @@ export class RfqOrder extends OrderBase {
...
+            },
+            domain: createExchangeProxyEIP712Domain(this.chainId, this.verifyingContract) as any,
+            primaryType: TakerSignedRfqOrder.STRUCT_NAME,
+            message: {
+                makerToken: this.makerToken,
+                takerToken: this.takerToken,
+                makerAmount: this.makerAmount.toString(10),
+                takerAmount: this.takerAmount.toString(10),
+                maker: this.maker,
+                taker: this.maker,
```
dorothy-zbornak, line 400:
```suggestion
                taker: this.taker,
```

contracts/zero-ex/test/features/native_orders_feature_test.ts
dorothy-zbornak: I would also prefer to delete any tests that are just gas cost tests since ganache is a total lie anyway.

contracts/zero-ex/contracts/src/features/native_orders/NativeOrdersSettlement.sol
```diff
@@ -514,6 +559,87 @@ abstract contract NativeOrdersSettlement is
...
+                uint8(takerSignedOrderInfo.status)
+            ).rrevert();
+        }
+
+        {
+            LibNativeOrdersStorage.Storage storage stor =
+                LibNativeOrdersStorage.getStorage();
+
+            // Must be fillable by the tx.origin.
+            if (order.txOrigin != tx.origin && !stor.originRegistry[order.txOrigin][tx.origin]) {
```
dorothy-zbornak, line 590: Have we evaluated whether this could pose a risk if the origin registries for RFQ and RFQM are the same? It feels pretty low stakes but just wondering if anyone else had other thoughts.

dorothy-zbornak (changes requested):
should probably check that the feature is deployable after these changes.

contracts/zero-ex/contracts/test/GetTypeHash.sol
```diff
@@ -0,0 +1,66 @@
...
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+pragma solidity ^0.6.5;
+pragma experimental ABIEncoderV2;
+
+
+contract GetTypeHash {
```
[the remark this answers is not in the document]
alexkroeger: meant to delete, just used it to get the typehash

contracts/zero-ex/contracts/src/storage/LibNativeOrdersStorage.sol
```diff
@@ -43,6 +43,8 @@ library LibNativeOrdersStorage {
         // For a given order origin, which tx.origin addresses are allowed to
         // fill the order.
         mapping(address => mapping(address => bool)) originRegistry;
+        // Whether or not the order has been filled
+        mapping(bytes32 => bool) orderHashToFilledBool;
```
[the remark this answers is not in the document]
alexkroeger: we could and just use the low bit, that's probably better 👍

contracts/zero-ex/test/features/native_orders_feature_test.ts
alexkroeger: these tests are not finalized, I just wanted to get a rough estimate of relative costs, if we move forward I'll change these into useful tests

contracts/zero-ex/contracts/src/features/native_orders/NativeOrdersSettlement.sol
alexkroeger: I was thinking it would be fine.

I can't think of a case where you'd specify an address with registered tx origins and not be okay with it for one order type and not the others. In either case you're giving power of tx submission to the entity behind the specified tx origin who may or may not have delegated other senders.

contracts/zero-ex/contracts/src/features/native_orders/NativeOrdersInfo.sol
```diff
@@ -112,6 +123,31 @@ abstract contract NativeOrdersInfo is
...
+
+        // Check for missing taker.
+        if (order.taker == address(0)) {
+            takerSignedOrderInfo.status = LibNativeOrder.OrderStatus.INVALID;
+        }
+        // Check for missing txOrigin.
+        if (order.txOrigin == address(0)) {
+            takerSignedOrderInfo.status = LibNativeOrder.OrderStatus.INVALID;
+        }
+    }
```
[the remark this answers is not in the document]
alexkroeger: while it's nonsense to have an order with 0 address for taker, not sure we need to enforce after thinking about it again. not sure we have to enforce tx origin either

contracts/zero-ex/contracts/src/features/native_orders/NativeOrdersSettlement.sol
```diff
@@ -514,6 +563,89 @@ abstract contract NativeOrdersSettlement is
...
+                orderHash: takerSignedOrderInfo.orderHash,
+                maker: order.maker,
+                taker: order.taker,
+                makerToken: IERC20TokenV06(order.makerToken),
+                takerToken: IERC20TokenV06(order.takerToken),
+                makerAmount: order.makerAmount,
+                takerAmount: order.takerAmount,
+                txOrigin: order.txOrigin,
+                txOriginNonce: order.txOriginNonce
+            })
```
mzhu25, line 635: instead of this `SettleTakerSignedOrderInfo` struct, let's just pass in the orderHash and the `order`.

contracts/zero-ex/contracts/src/features/libs/LibNativeOrder.sol
```diff
@@ -118,6 +137,23 @@ library LibNativeOrder {
...
+    //     "TakerSignedRfqOrder(",
+    //       "address makerToken,",
+    //       "address takerToken,",
+    //       "uint128 makerAmount,",
+    //       "uint128 takerAmount,",
+    //       "address maker,",
+    //       "address taker,",
+    //       "address txOrigin,",
+    //       "uint64 expiry,",
+    //       "uint256 salt"
```
mzhu25, line 151: should be `uint256 txOriginNonce`

contracts/zero-ex/contracts/src/features/libs/LibNativeOrder.sol
```diff
@@ -69,13 +69,32 @@ library LibNativeOrder {
         uint256 salt;
     }
 
+    /// @dev A Taker-signed RFQ order.
+    struct TakerSignedRfqOrder {
```
mzhu25, line 73: `OtcOrder` perhaps? Just to differentiate

mzhu25 (commented):
I like it!

PR #217, opened 2021-05-01
fix: Temporarily remove PancakeV2/Bakery from VIP

dekz:
## Description

Need to update all references to forked values as `BakerySwap` is now `2` and `PancakeSwapV2` is `1`

Temporarily disabling VIP for these two sources

## Checklist:


contracts/zero-ex/contracts/src/features/PancakeSwapFeature.sol
```diff
@@ -280,7 +280,7 @@ contract PancakeSwapFeature is
 
                 // Call pair.swap()
                 switch mload(0xA20) // fork
-                    case 1 {
+                    case 2 {
```
dekz, line 283: Bakery is now `2`

PR #219, opened 2021-05-03
fix: add separate priceComparisonsReport to fix missing quoteReport data [TKR-91]

kimpers:
## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/market_operation_utils/index.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/index.ts
@@ -16,7 +16,14 @@ import {
     getNativeAdjustedMakerFillAmount,
 } from '../utils';
 
-import { generateQuoteReport, QuoteReport } from './../quote_report_generator';
+import {
+    dexSampleToReportSource,
+    generateQuoteReport,
+    multiHopSampleToReportSource,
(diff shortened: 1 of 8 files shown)


packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -677,7 +684,25 @@ export class MarketOperationUtils {
                 wholeOrderPrice,
             );
         }
-        return { ...optimizerResult, quoteReport };
+
+        let priceComparisonsReport: PriceComparisonsReport | undefined;
+        if (_opts.shouldIncludePriceComparisonsReport) {
```
dekz, line 689: Can this be bumped to a static function like the `_computeQuoteReport`

dekz (approved):

packages/asset-swapper/src/utils/market_operation_utils/index.ts
kimpers: I think a simple Assertible test is good enough, I set one up with some jq magic to count number of items in `priceComparisons` where price is **not** null (`.priceComparisons.jq('[.[] |select(.price != null)] | length')`). We can then just assert that there's at least 1 valid price in the response 👍

PR #223, opened 2021-05-04
Fix: PLP now generates a fallback

dekz:
## Description

Observed collisions with PLP are above an acceptable amount. Thus PLP now creates a fallback in the event of an on-chain failure.

[Simbot](https://metabase.spaceship.0x.org/dashboard/186?run_id=jacob-mainnet-plp-fallback&adjusted_amount_bought_win_tolerance__bps_=1)

## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/market_operation_utils/index.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/index.ts
@@ -554,25 +554,29 @@ export class MarketOperationUtils {
             throw new Error(AggregationError.NoOptimalPath);
         }
 
-        // Generate a fallback path if native orders are in the optimal path.
-        const nativeFills = optimalPath.fills.filter(f => f.source === ERC20BridgeSource.Native);
-        if (opts.allowFallback && nativeFills.length !== 0) {
+        // Generate a fallback path if sources requiring a fallback (fragile) are in the optimal path.
+        // Native is relatively fragile (limit order collision, expiry, or lack of available maker balance)
+        // LiquidityProvider is relatively fragile (collision)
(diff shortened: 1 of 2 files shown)


packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -554,25 +554,29 @@ export class MarketOperationUtils {
...
+        // Native is relatively fragile (limit order collision, expiry, or lack of available maker balance)
+        // LiquidityProvider is relatively fragile (collision)
+        const fragileSources = [ERC20BridgeSource.Native, ERC20BridgeSource.LiquidityProvider];
+        const fragileFills = optimalPath.fills.filter(f => fragileSources.includes(f.source));
+        if (opts.allowFallback && fragileFills.length !== 0) {
             // We create a fallback path that is exclusive of Native liquidity
             // This is the optimal on-chain path for the entire input amount
-            const nonNativeFills = fills.filter(p => p.length > 0 && p[0].source !== ERC20BridgeSource.Native);
-            const nonNativeOptimalPath = await findOptimalPathAsync(side, nonNativeFills, inputAmount, opts.runLimit, {
+            const sturdyFills = fills.filter(p => p.length > 0 && p[0].source !== ERC20BridgeSource.Native);
```
mzhu25, line 565:
```suggestion
            const sturdyFills = fills.filter(p => p.length > 0 && !fragileSources.includes(p[0].source));
```

dekz: fml

PR #226, opened 2021-05-05
feat: async balancer v2 pool cache

dekz:
## Description

Cache pools for 30mins, fetch them async if not fresh.

## Checklist:


kimpers (approved):
Nice! This is a better flow 👍 

Just one small nit (but I think it's an existing problem in prod today), since we just blindly swallow all cache pool update errors we wouldn't really know if something broke amirite?

PR #228, opened 2021-05-05
feat: Balancer V2 load and cache top pools by num swaps on startup [TKR-96]

kimpers:
## Description
This implements a slightly hacky solution to pre-populate the Balancer V2 pool cache on start-up so we avoid having empty a cache the first time we try to sample a token pair. Once Balance releases a new version of SOR that is compatible with both V1 and V2 we can remove the copy-pasta and clean this up 😄 

## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/market_operation_utils/pools_cache/balancer_v2_utils.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/pools_cache/balancer_v2_utils.ts
@@ -1,19 +1,60 @@
-// import { getPoolsWithTokens, parsePoolData } from '@balancer-labs/sor'; // TODO - upgrade to v2
 import { BigNumber } from '@0x/utils';
+// import { parsePoolData } from '@balancer-labs'; // TODO - upgrade to v2
 import { Pool } from '@balancer-labs/sor/dist/types';
-import { request } from 'graphql-request';
+import { gql, request } from 'graphql-request';
 
-import { BALANCER_MAX_POOLS_FETCHED, BALANCER_V2_SUBGRAPH_URL } from '../constants';
+import { DEFAULT_WARNING_LOGGER } from '../../../constants';
+import { LogFunction } from '../../../types';
+import { BALANCER_MAX_POOLS_FETCHED, BALANCER_TOP_POOLS_FETCHED, BALANCER_V2_SUBGRAPH_URL } from '../constants';
 
+import { parsePoolData } from './balancer_sor_v2';
 import { CacheValue, PoolsCache } from './pools_cache';
 
+// tslint:disable-next-line:custom-no-magic-numbers
+const ONE_DAY_MS = 24 * 60 * 60 * 1000;
+
+interface BalancerPoolResponse {
+    id: string;
+    swapFee: string;
+    tokens: Array<{ address: string; decimals: number; balance: string; weight: string; symbol: string }>;
+    tokensList: string[];
+    totalWeight: string;
+    totalShares: string;
+    amp: string | null;
+}
+
 export class BalancerV2PoolsCache extends PoolsCache {
+    private static _parseSubgraphPoolData(pool: any, takerToken: string, makerToken: string): Pool {
+        const tToken = pool.tokens.find((t: any) => t.address === takerToken);
+        const mToken = pool.tokens.find((t: any) => t.address === makerToken);
+        const swap = pool.swaps && pool.swaps[0];
+        const tokenAmountOut = swap ? swap.tokenAmountOut : undefined;
+        const tokenAmountIn = swap ? swap.tokenAmountIn : undefined;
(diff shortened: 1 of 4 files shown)


packages/asset-swapper/src/utils/market_operation_utils/pools_cache/balancer_v2_utils.ts
```diff
@@ -29,8 +70,74 @@ export class BalancerV2PoolsCache extends PoolsCache {
...
+                        this._warningLogger(err, `Failed to load Balancer V2 top pools`);
+                        // soldier on
+                    }
+                }
+            }
+        });
+    }
     protected async _fetchPoolsForPairAsync(takerToken: string, makerToken: string): Promise<Pool[]> {
-        const query = `
+        const query = gql`
```
dorothy-zbornak, line 140: oh, never seen this syntax before, how does this work?

packages/asset-swapper/src/utils/market_operation_utils/pools_cache/balancer_v2_utils.ts
```diff
@@ -1,19 +1,60 @@
...
         private readonly subgraphUrl: string = BALANCER_V2_SUBGRAPH_URL,
         private readonly maxPoolsFetched: number = BALANCER_MAX_POOLS_FETCHED,
+        private readonly _topPoolsFetched: number = BALANCER_TOP_POOLS_FETCHED,
+        private readonly _warningLogger: LogFunction = DEFAULT_WARNING_LOGGER,
         cache: { [key: string]: CacheValue } = {},
     ) {
         super(cache);
+        void this._loadTopPoolsAsync();
+        // Reload the top pools every 12 hours
+        setInterval(async () => void this._loadTopPoolsAsync(), ONE_DAY_MS / 2);
```
dorothy-zbornak, line 57: why?

packages/asset-swapper/src/utils/market_operation_utils/pools_cache/balancer_v2_utils.ts
```diff
@@ -29,8 +70,74 @@ export class BalancerV2PoolsCache extends PoolsCache {
...
+                    }
+                    try {
+                        // The list of pools must be relevant to `from` and `to`  for `parsePoolData`
+                        const [poolData] = parsePoolData({ [pool.id]: pool as any }, from, to);
+                        fromToPools[from][to].push(
+                            BalancerV2PoolsCache._parseSubgraphPoolData(poolData[pool.id], from, to),
+                        );
+                        // Cache this as we progress through
+                        const expiresAt = Date.now() + this._cacheTimeMs;
+                        this._cachePoolsForPair(from, to, fromToPools[from][to], expiresAt);
```
dorothy-zbornak, line 130: `this` is not clobbered by the `forEach()`?

packages/asset-swapper/src/utils/market_operation_utils/pools_cache/balancer_v2_utils.ts
```diff
@@ -29,8 +70,74 @@ export class BalancerV2PoolsCache extends PoolsCache {
...
+        pools.forEach(pool => {
+            const { tokensList } = pool;
+            for (const from of tokensList) {
+                for (const to of tokensList.filter(t => t.toLowerCase() !== from.toLowerCase())) {
+                    if (!fromToPools[from]) {
+                        fromToPools[from] = {};
+                    }
+                    if (!fromToPools[from][to]) {
+                        fromToPools[from][to] = [];
+                    }
```
dorothy-zbornak, line 121: perhaps
```suggestion
                    fromToPools[from] = fromToPools[from] || {};
                    fromToPools[from][to] = fromToPools[from][to] || {};
```

dorothy-zbornak (approved):
:+1: :eye: :lips: :eye: :+1:

packages/asset-swapper/src/utils/market_operation_utils/pools_cache/balancer_v2_utils.ts
dekz: If it wasn't discovered by the last deployment (on init), then the only way a pool can be discovered is through natural search. 

We want a method to discover decently liquid pools that users may not naturally search. Maybe a liquid BAL/USDT pool appears but no one ever searches BAL/USDT, they always search BAL/USDC.  We don't go searching on Multihop intermediary hops automatically.

packages/asset-swapper/src/utils/market_operation_utils/pools_cache/balancer_v2_utils.ts
kimpers: I think it's just a tagged string template so it doesn't do so much. However, now that the editor understands that it's a GQL query you can get prettier formatting and some basic linting :)

PR #229, opened 2021-05-05
AS/Sampler: Tweak compiler settings for smaller sampler size

dorothy-zbornak:
## Description

We probably care about having a smaller bytecode footprint than being efficient in the sampler since gas is free and we have to pass the full contract to the RPC endpoint for every eth_call.

Prev: ~51KB
Now: ~40KB

## Checklist:


mzhu25 (approved):
good call 👍

PR #233, opened 2021-05-10
fix: TwoHopSampler call

dekz:
## Description

`TwoHopSampler` needs to perform a `call` instead of a `staticcall`. The `amounts` need to be the last value returned by one of the sampler functions.

[Simbot](https://metabase.spaceship.0x.org/dashboard/186?run_id=jacob-fix-two-hop-sampler&adjusted_amount_bought_win_tolerance__bps_=0.5)

## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/market_operation_utils/constants.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/constants.ts
@@ -453,7 +453,10 @@ export const NATIVE_FEE_TOKEN_BY_CHAIN_ID = valueByChainId<string>(
     NULL_ADDRESS,
 );
 
-export const NATIVE_FEE_TOKEN_AMOUNT_BY_CHAIN_ID = valueByChainId({}, ONE_ETHER);
+export const NATIVE_FEE_TOKEN_AMOUNT_BY_CHAIN_ID = valueByChainId(
+    { [ChainId.Mainnet]: ONE_ETHER.times(0.1) },
+    ONE_ETHER,
+);
 
 // Order dependent
 const CURVE_TRI_POOL_MAINNET_TOKENS = [MAINNET_TOKENS.DAI, MAINNET_TOKENS.USDC, MAINNET_TOKENS.USDT];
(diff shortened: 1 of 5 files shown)


packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -838,7 +841,10 @@ export class SamplerOperations {
                     };
                 });
             },
-            () => [],
+            () => {
+                logUtils.warn('SamplerContractOperation: Two hop sampler reverted');
```
dorothy-zbornak, line 845: Delete?

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -453,7 +453,10 @@ export const NATIVE_FEE_TOKEN_BY_CHAIN_ID = valueByChainId<string>(
     NULL_ADDRESS,
 );
 
-export const NATIVE_FEE_TOKEN_AMOUNT_BY_CHAIN_ID = valueByChainId({}, ONE_ETHER);
+export const NATIVE_FEE_TOKEN_AMOUNT_BY_CHAIN_ID = valueByChainId(
+    { [ChainId.Mainnet]: ONE_ETHER.times(0.1) },
```
dorothy-zbornak, line 457: ?

dorothy-zbornak (approved):

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
dekz: Oh yeah, I felt like getting querying $4000 of the token value might be a bit high these days. WDYT

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
dekz: I think keep so we can catch it when looking at logs, similar to the base sampler contract operation.

https://github.com/0xProject/protocol/blob/development/packages/asset-swapper/src/utils/market_operation_utils/sampler_contract_operation.ts#L59

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
dorothy-zbornak: Yeah seems fair

PR #240, opened 2021-05-12
Polygon support

kimpers:
## Checklist:


packages/asset-swapper/package.json
```diff
@@ -60,14 +60,15 @@
     "dependencies": {
         "@0x/assert": "^3.0.27",
         "@0x/base-contract": "^6.4.0",
-        "@0x/contract-addresses": "^6.1.0",
-        "@0x/contract-wrappers": "^13.16.1",
+        "@0x/contract-addresses": "0xProject/gitpkg-registry#0x-contract-addresses-v6.1.0-d0006cbc3",
+        "@0x/contract-wrappers": "0xProject/gitpkg-registry#0x-contract-wrappers-v13.16.1-baafa8a7e",
```
kimpers, line 64: **TODO:** Remove once we're ready to publish real versions

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -1,5 +1,7 @@
 import { ChainId, getContractAddressesForChainOrThrow } from '@0x/contract-addresses';
 import { FillQuoteTransformerOrderType } from '@0x/protocol-utils';
+// TODO(kimpers): does this belong here?
+import { valueByChainId } from '@0x/token-metadata';
```
kimpers, line 4: It's a bit weird to import this from here, but we used to it duplicated here and in `token-metadata`. In that case you would need to add any new chain in both places, which tripped me up initially.

I think it could be better to just move it here for now, rather than duplicating but I'm also open to reverting back to the previous duplication. Wdyt?

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -124,8 +113,16 @@ export const SELL_SOURCE_FILTER_BY_CHAIN_ID = valueByChainId<SourceFilters>(
             ERC20BridgeSource.JulSwap,
             ERC20BridgeSource.LiquidityProvider,
         ]),
+        [ChainId.Polygon]: new SourceFilters([
```
dekz, line 116: I think DODO v1 is also deployed on Polygon https://dodoex.github.io/docs/docs/deployedInfoMatic#dodo-v1

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -426,6 +446,14 @@ export const DEFAULT_INTERMEDIATE_TOKENS_BY_CHAIN_ID = valueByChainId<string[]>(
             '0xad6d458402f60fd3bd25163575031acdce07538d', // DAI
             '0x07865c6e87b9f70255377e024ace6630c1eaa37f', // USDC
         ],
+        [ChainId.Polygon]: [
+            getContractAddressesForChainOrThrow(ChainId.Polygon).etherToken,
+            '0x7ceb23fd6bc0add59e62ac25578270cff1b9f619', // WETH
```
dekz, line 451: I'd re-use `POLYGON_TOKENS` since its been defined. Should've done it for BSC.

DAI address here is mainnet DAI?

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
dekz: I'm not entire sure the purpose of `@0x/token-metadata` if we also have addresses declared here as well.

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -694,6 +726,19 @@ export const CURVE_MAINNET_INFOS: { [name: string]: CurveInfo } = {
...
+export const CURVE_POLYGON_INFOS: { [name: string]: CurveInfo } = {
+    ['aave_exchangeunderlying']: createCurveExchangeUnderlyingPool({
+        tokens: [POLYGON_TOKENS.DAI, POLYGON_TOKENS.USDC, POLYGON_TOKENS.USDT],
+        pool: CURVE_POLYGON_POOLS.aave,
+        gasSchedule: 1e3,
+    }),
+    ['aave_exchange']: createCurveExchangePool({
+        tokens: [POLYGON_TOKENS.amDAI, POLYGON_TOKENS.amUSDC, POLYGON_TOKENS.amUSDT],
+        pool: CURVE_POLYGON_POOLS.aave,
+        gasSchedule: 1e3,
```
dekz, line 738:
```suggestion
        gasSchedule: 150e3,
```
I'm guessing

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -694,6 +726,19 @@ export const CURVE_MAINNET_INFOS: { [name: string]: CurveInfo } = {
     }),
 };
 
+export const CURVE_POLYGON_INFOS: { [name: string]: CurveInfo } = {
+    ['aave_exchangeunderlying']: createCurveExchangeUnderlyingPool({
+        tokens: [POLYGON_TOKENS.DAI, POLYGON_TOKENS.USDC, POLYGON_TOKENS.USDT],
+        pool: CURVE_POLYGON_POOLS.aave,
+        gasSchedule: 1e3,
```
dekz, line 733:
```suggestion
        gasSchedule: 300e3,
```
I'm guessing

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -1189,42 +1286,10 @@ export const DEFAULT_GAS_SCHEDULE: Required<FeeSchedule> = {
...
-        const path = (fillData as UniswapV2FillData).tokenAddressPath;
-        if (path.length > 2) {
-            gas += (path.length - 2) * 60e3; // +60k for each hop.
-        }
-        return gas;
-    },
+    [ERC20BridgeSource.UniswapV2]: uniswapV2CloneGasSchedule,
+    [ERC20BridgeSource.SushiSwap]: uniswapV2CloneGasSchedule,
+    [ERC20BridgeSource.CryptoCom]: uniswapV2CloneGasSchedule,
+    [ERC20BridgeSource.Linkswap]: uniswapV2CloneGasSchedule,
```
dekz, line 1292: 😂

dekz (approved):
looking good. Few minor comments but I think we're ready to roll

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
kimpers: I think it was originally created as a way to share the token metadata between 0x API and the data service.

As a follow-up to this, it might make sense to move **all** to token related data into that package and stop having addresses sprinkled everywhere 😂

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
kimpers: 👍  whoops, I must've copied from the left column instead of the right one on https://mapper.matic.today 😂

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
kimpers: Integrated and verified working 👍

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
kimpers: I'm reverting this change in order to remove the unwieldy circular dependency that has come up here

PR #244, opened 2021-05-13
OTC orders feature

mzhu25:
## Description

Basically refactors https://github.com/0xProject/protocol/pull/204 into a separate feature, with a couple of notable differences:
- `TakerSignedRfqOrder` renamed to `OtcOrder`
- Allows partial fills (we just won't store partial fill amounts) if the taker submits the transaction
- Exposes three fill functions: `fillTakerSignedOtcOrder`, `fillOtcOrder`, and `fillOtcOrderWithEth`. The first is the "meta-transactions" variant that RFQM will use. The latter two might come in handy if we want to slim down RFQT.
- `fillTakerSignedOtcOrder` and `fillOtcOrder` both support ETH buys. `fillOtcOrderWithEth` supports ETH sells.

## Checklist:


mzhu25:
unit tests coming soon

alexkroeger:
@mzhu25 

There was some discussion on the maker team regarding changing the `TakerSigned` order to use a `txOriginSlot` in addition to `txOriginNonce`, so the storage would look like:

```
mapping(address => mapping(uint256 => uint256)) txOriginToSlotToNonce;
```

While this entails a small gas hit, we think it'll make using the `originRegistry` much easier and thus ease integration.

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
```diff
@@ -0,0 +1,334 @@
...
+                orderHash,
+                msg.sender,
+                order.taker
+            ).rrevert();
+        }
+        LibSignature.Signature memory nullSignature;
+        return _fillOtcOrderPrivate(
+            order,
+            makerSignature,
+            nullSignature,
```
alexkroeger, line 118: do we incur any gas penalty from doing this?

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
```diff
@@ -0,0 +1,334 @@
...
+                LibNativeOrdersRichErrors.OrderNotSignedByMakerError(
+                    orderInfo.orderHash,
+                    makerSigner,
+                    order.maker
+                ).rrevert();
+            }
+        }
+
+        // If msg.sender is not the taker, validate the taker signature.
+        if (order.taker != msg.sender) {
```
alexkroeger, line 210: we typically leave order.taker as the null address currently. this gives us the flexibility to use the FlashWallet or Multiplex. Could we change this to 

```
(order.taker != msg.sender or order.taker == null_address)
```

and change the send to address to if (order.taker == null_address) then msg.sender else order.taker?

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
mzhu25: not really

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
```diff
@@ -0,0 +1,334 @@
...
+    /// @param takerSignature The order signature from the taker.
+    /// @param takerTokenFillAmount Maximum taker token amount to fill this
+    ///        order with.
+    /// @return takerTokenFilledAmount How much maker token was filled.
+    /// @return makerTokenFilledAmount How much maker token was filled.
+    function fillTakerSignedOtcOrder(
+        LibNativeOrder.OtcOrder memory order,
+        LibSignature.Signature memory makerSignature,
+        LibSignature.Signature memory takerSignature,
+        uint128 takerTokenFillAmount
```
alexkroeger, line 137: Partial fills here are tricky. The taker then isn't guaranteed any particular amount by signing, the `txOrigin` completely dictates the amount.

alexkroeger: Not a security flaw but I think worth discussing.

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
```diff
@@ -0,0 +1,334 @@
...
+
+    /// @dev Fill a OTC order for up to `takerTokenFillAmount` taker tokens.
+    ///      "Meta-transaction" variant, requires order to be signed by both
+    ///      maker and taker.
+    /// @param order The OTC order.
+    /// @param makerSignature The order signature from the maker.
+    /// @param takerSignature The order signature from the taker.
+    /// @param takerTokenFillAmount Maximum taker token amount to fill this
+    ///        order with.
+    /// @return takerTokenFilledAmount How much maker token was filled.
```
alexkroeger, line 131:
```suggestion
    /// @return takerTokenFilledAmount How much taker token was filled.
```

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
```diff
@@ -0,0 +1,334 @@
...
+        _registerFeatureFunction(this.getOtcOrderHash.selector);
+        return LibMigrate.MIGRATE_SUCCESS;
+    }
+
+    /// @dev Fill a OTC order for up to `takerTokenFillAmount` taker tokens.
+    /// @param order The OTC order.
+    /// @param makerSignature The order signature from the maker.
+    /// @param takerTokenFillAmount Maximum taker token amount to fill this
+    ///        order with.
+    /// @return takerTokenFilledAmount How much maker token was filled.
```
alexkroeger, line 95:
```suggestion
    /// @return takerTokenFilledAmount How much taker token was filled.
```

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
mzhu25: discussed offline; decided to remove partial fills for this function

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
```diff
@@ -0,0 +1,443 @@
...
+        );
+    }
+
+    /// @dev Fill an OTC order whose taker token is WETH for up
+    ///      to `msg.value`.
+    /// @param order The OTC order.
+    /// @param makerSignature The order signature from the maker.
+    /// @return takerTokenFilledAmount How much taker token was filled.
+    /// @return makerTokenFilledAmount How much maker token was filled.
+    function fillOtcOrderWithEth(
```
PirosB3, line 122: When would we ever use this function?
- In RFQM, the taker does not have opportunity to send ETH because they are not submitting the trade
- In RFQT, we usually use the VIP+Fallback route when interacting with `fillOtcOrder`. Is that route `payable`?, also, If the taker was to send ETH with the VIP fallback route and the order reverts and falls back to Uni, what would happen to the ETH?

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
```diff
@@ -0,0 +1,443 @@
...
+        // compute order hash.
+        orderInfo.orderHash = getOtcOrderHash(order);
+
+        LibOtcOrdersStorage.Storage storage stor =
+            LibOtcOrdersStorage.getStorage();
+
+        // check tx origin nonce
+        uint256 lastNonce = stor.txOriginNonces
+            [order.txOrigin]
+            [order.txOriginNonceBucket];
```
PirosB3, line 374: It looks like we support `(2 ** 256) -1` buckets. Is there any value in restricting these buckets to a fixed number? example: 100?

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
```diff
@@ -0,0 +1,443 @@
...
+        LibNativeOrder.OtcOrder memory order,
+        LibSignature.Signature memory makerSignature,
+        uint128 takerTokenFillAmount,
+        bool unwrapWeth
+    )
+        public
+        override
+        returns (uint128 takerTokenFilledAmount, uint128 makerTokenFilledAmount)
+    {
+        if (!_isSenderValidTaker(order.taker)) {
```
PirosB3, line 98: Isn't this check done twice? once here, and then another time in the `_fillOtcOrderPrivate`? Is there a clean way to avoid running the function twice?

PirosB3: (I'm mentioning this since I assume it consumes gas, however I'm unsure what gas savings a refactor will have)

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
```diff
@@ -0,0 +1,443 @@
...
+                ).rrevert();
+            }
+        }
+
+        address taker = msg.sender;
+        // If msg.sender is not the taker, validate the taker signature.
+        if (!_isSenderValidTaker(order.taker)) {
+            address takerSigner = LibSignature.getSignerOfHash(orderInfo.orderHash, takerSignature);
+            if (takerSigner != order.taker) {
+                LibNativeOrdersRichErrors.OrderNotSignedByTakerError(
```
PirosB3, line 242: is it safe to `return` here too? or is this the equivalent of throwing an exception?

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
```diff
@@ -0,0 +1,443 @@
...
+                    makerSigner,
+                    order.maker
+                ).rrevert();
+            }
+        }
+
+        address taker = msg.sender;
+        // If msg.sender is not the taker, validate the taker signature.
+        if (!_isSenderValidTaker(order.taker)) {
+            address takerSigner = LibSignature.getSignerOfHash(orderInfo.orderHash, takerSignature);
```
PirosB3, line 240: If an RFQT order has no taker at all, and someone uses the `fillTakerSignedOtcOrder` with a null-signed taker order. Would this pass the checks?

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
```diff
@@ -0,0 +1,443 @@
...
+
+        // Settle between the maker and taker.
+        (takerTokenFilledAmount, makerTokenFilledAmount) = _settleOtcOrder(
+            order,
+            taker,
+            takerTokenFillAmount,
+            wethOptions
+        );
+
+        emit OtcOrderFilled(
```
PirosB3, line 259: Could we avoid emitting events, and simply rely on the ERC20 transfers? would that save gas?

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
```diff
@@ -0,0 +1,443 @@
...
+            );
+        }
+
+        if (wethOptions == WethOptions.UnwrapWeth) {
+            require(
+                order.makerToken == WETH,
+                "OtcOrdersFeature/INVALID_UNWRAP_WETH"
+            );
+            // Transfer maker tokens in
+            _transferERC20Tokens(
```
PirosB3, line 335: Isn't this going to be more expensive? because we need to first transfer the tokens to this contract, then unwrap, then send?

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
```diff
@@ -0,0 +1,443 @@
...
+    /// @dev Initialize and register this feature.
+    ///      Should be delegatecalled by `Migrate.migrate()`.
+    /// @return success `LibMigrate.SUCCESS` on success.
+    function migrate()
+        external
+        returns (bytes4 success)
+    {
+        _registerFeatureFunction(this.fillOtcOrder.selector);
+        _registerFeatureFunction(this.fillTakerSignedOtcOrder.selector);
+        _registerFeatureFunction(this.getOtcOrderInfo.selector);
```
dorothy-zbornak, line 72: missing `fillOtcOrderWithEth`

dorothy-zbornak (approved):
nicely done :+1:

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
```diff
@@ -0,0 +1,443 @@
...
+        }
+
+        {
+            LibNativeOrdersStorage.Storage storage stor =
+                LibNativeOrdersStorage.getStorage();
+
+            // Must be fillable by the tx.origin.
+            if (
+                order.txOrigin != tx.origin &&
+                !stor.originRegistry[order.txOrigin][tx.origin]
```
PirosB3, line 214: How do we add data to the `originRegistry`? Is there a getter/setter?

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
```diff
@@ -0,0 +1,443 @@
...
+        override
+        view
+        returns (uint256 minNonce)
+    {
+        LibOtcOrdersStorage.Storage storage stor =
+            LibOtcOrdersStorage.getStorage();
+        uint256 lastNonce = stor.txOriginNonces
+            [txOrigin]
+            [nonceBucket];
+        return lastNonce.safeAdd(1);
```
PirosB3, line 421: What happens if this overflows?

contracts/zero-ex/contracts/src/features/interfaces/IOtcOrdersFeature.sol
```diff
@@ -0,0 +1,133 @@
...
+interface IOtcOrdersFeature {
+
+    /// @dev Options for handling ETH/WETH conversion
+    /// @param LeaveAsWeth Neither unwrap nor wrap.
+    /// @param WrapEth Wrap attached ETH.
+    /// @param UnwrapWeth Unwrap WETH before transferring
+    ///        to taker.
+    enum WethOptions {
+        LeaveAsWeth,
+        WrapEth,
```
PirosB3, line 37: During the fill of a quote, should we have error checks such as

```
- if not selling ETH, raise an error if WrapEth is set
- if not buy ETH, raise an error if UnwrapWeth is set

**Review by PirosB3 — CHANGES_REQUESTED** [review]

File: contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol [hunk 17]
```diff
@@ -0,0 +1,443 @@
...
+    }
+
+    /// @dev Initialize and register this feature.
+    ///      Should be delegatecalled by `Migrate.migrate()`.
+    /// @return success `LibMigrate.SUCCESS` on success.
+    function migrate()
+        external
+        returns (bytes4 success)
+    {
+        _registerFeatureFunction(this.fillOtcOrder.selector);
```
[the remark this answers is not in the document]
PirosB3: Discussed with Kroeger, the code will not be migrated to mainnet until we have had enough time to gather feedback from market makers

contracts/zero-ex/contracts/src/features/libs/LibNativeOrder.sol
```diff
@@ -222,6 +260,55 @@ library LibNativeOrder {
...
+            mstore(add(mem, 0xC0), and(ADDRESS_MASK, mload(add(order, 0xA0))))
+            // order.txOrigin;
+            mstore(add(mem, 0xE0), and(ADDRESS_MASK, mload(add(order, 0xC0))))
+            // order.expiry;
+            mstore(add(mem, 0x100), and(UINT_64_MASK, mload(add(order, 0xE0))))
+            // order.txOriginNonceBucket;
+            mstore(add(mem, 0x120), mload(add(order, 0x140)))
+            // order.txOriginNonce;
+            mstore(add(mem, 0x140), mload(add(order, 0x160)))
+            structHash := keccak256(mem, 0x160)
```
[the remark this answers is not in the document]
mzhu25: yeah just caught that lol

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
```diff
@@ -0,0 +1,443 @@
...
+        if (orderInfo.status != LibNativeOrder.OrderStatus.FILLABLE) {
+            LibNativeOrdersRichErrors.OrderNotFillableError(
+                orderInfo.orderHash,
+                uint8(orderInfo.status)
+            ).rrevert();
+        }
+
+        {
+            LibNativeOrdersStorage.Storage storage stor =
+                LibNativeOrdersStorage.getStorage();
```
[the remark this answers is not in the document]
mzhu25: 🤫

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
```diff
@@ -0,0 +1,443 @@
...
+    /// @param order The OTC order.
+    /// @param makerSignature The order signature from the maker.
+    /// @param takerTokenFillAmount Maximum taker token amount to fill this
+    ///        order with.
+    /// @param unwrapWeth Whether or not to unwrap bought WETH into ETH
+    ///        before transferring it to the taker. Should be set to false
+    ///        if the maker token is not WETH.
+    /// @return takerTokenFilledAmount How much taker token was filled.
+    /// @return makerTokenFilledAmount How much maker token was filled.
+    function fillOtcOrder(
```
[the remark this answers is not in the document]
mzhu25: yes

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
mzhu25: no

contracts/zero-ex/contracts/src/features/interfaces/IOtcOrdersFeature.sol
mzhu25: we do

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
mzhu25: it'll revert. I'll change this function to return `lastNonce` instead 👍

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
mzhu25: It's the same `originRegistry` as RFQ orders

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
mzhu25: cc @alexkroeger 👀

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
PirosB3: Thanks!

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
mzhu25: but no, I think @sui414 would kill us

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
mzhu25: this throws

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
mzhu25: EIP-2929 made warm SLOADs cheap

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
mzhu25: more expensive than what? yes it's more expensive than thus using WETH, but sometimes people want to receive ETH. This is the cheapest way to do that

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
mzhu25: If someone is filling a RFQt order with no fallback (e.g. for OTC) and they want to use ETH instead of WETH

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
mzhu25: yes

PirosB3 (approved):

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
```diff
@@ -0,0 +1,467 @@
...
+    }
+
+    /// @dev Fill an OTC order. Private variant.
+    /// @param order The OTC order.
+    /// @param makerSignature The order signature from the maker.
+    /// @param takerSignature The order signature from the taker.
+    ///        Ignored if msg.sender == order.taker.
+    /// @param takerTokenFillAmount Maximum taker token amount to
+    ///        fill this order with.
+    /// @return takerTokenFilledAmount How much maker token was filled.
```
alexkroeger, line 198:
```suggestion
    /// @return takerTokenFilledAmount How much taker token was filled.
```

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
```diff
@@ -0,0 +1,467 @@
...
+
+        // Clamp the taker token fill amount to the fillable amount.
+        takerTokenFilledAmount = LibSafeMathV06.min128(
+            takerTokenFillAmount,
+            order.takerAmount
+        );
+        // Compute the maker token amount.
+        // This should never overflow because the values are all clamped to
+        // (2^128-1).
+        makerTokenFilledAmount = uint128(LibMathV06.getPartialAmountFloor(
```
alexkroeger, line 318: curious if this calculation is expensive at all, for taker signed flow we know `makerTokenFilledAmount` = `order.makerAmount` if we can identify it at this point

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
```diff
@@ -0,0 +1,467 @@
...
+        // Transfer ETH to recipient
+        (bool success, bytes memory revertData) =
+            recipient.call{value: amount}("");
+        // Revert on failure
+        if (!success) {
+            revertData.rrevert();
+        }
+    }
+
+    function _isSenderValidTaker(address orderTaker)
```
alexkroeger, line 460: nice refactor 👍

contracts/zero-ex/contracts/src/features/libs/LibNativeOrder.sol
```diff
@@ -69,13 +69,31 @@ library LibNativeOrder {
...
+    /// @dev An OTC limit order.
+    struct OtcOrder {
+        IERC20TokenV06 makerToken;
+        IERC20TokenV06 takerToken;
+        uint128 makerAmount;
+        uint128 takerAmount;
+        address maker;
+        address taker;
+        address txOrigin;
+        uint256 expiryAndNonce; // [uint64 expiry, uint64 nonceBucket, uint128 nonce]
```
alexkroeger, line 81:
```suggestion
        uint256 expiryAndNonce; // [uint128 nonce, uint64 nonceBucket, uint64 expiry]
```

contracts/zero-ex/contracts/src/storage/LibOtcOrdersStorage.sol
```diff
@@ -0,0 +1,45 @@
...
+
+import "./LibStorage.sol";
+
+
+/// @dev Storage helpers for `OtcOrdersFeature`.
+library LibOtcOrdersStorage {
+
+    /// @dev Storage bucket for this feature.
+    struct Storage {
+        // tx origin => nonce buckets => min nonce
```
alexkroeger, line 31:
```suggestion
        // tx origin => nonce buckets => last used nonce
```

alexkroeger (approved):
Looks good 👍 

I like the wrap/unwrap options for direct use.

contracts/zero-ex/contracts/src/features/libs/LibNativeOrder.sol
mzhu25: no this is right, evm is big-endian

contracts/zero-ex/contracts/src/features/OtcOrdersFeature.sol
mzhu25: good call 👍

PR #250, opened 2021-05-24
AS refactors (1/X) [TKR-32]

dorothy-zbornak:
TKR-32

## Description

This is phase 1 of the AS refactors. This phase focuses on the sampling part of AS. The main goals/highlights are:

- Smash the bloated, monolithic `DexOrderSampler/SamplerOperations` class that had to know how every source worked.
- All source-specific logic should live inside the that source's integration file(s), abstracted behind a common source sampler interface. This includes off-chain (subgraph) queries, pooling, configurations, etc.
- Source sampler classes should be standalone, functionally independent. If we want to call/test them on their own, we could.
- But we still have a convenient meta-sampler (`SassySampler`) class that exposes `getMedianSellRate()`, `getSellSamples()`, etc, functions. This is `DexOrderSampler`'s spiritual successor but 1000x simpler because it has no clue what the sub-samplers it holds actually do, nor does it have to do any coordinating between them.
- Use a network abstraction (`Chain`) and contract helpers (`ContractHelper`) to execute eth_calls that will be:
  - automagically batched together into a single RPC call to a forwarding contract (`CallDispatcher`).
  - automagically cached and results reused (opt-in)
  - automagically broken up across multiple simultaneous RPC calls (if they cannot fit into one)
  - automagically only send contracts/code that are actually being used by the call (eventually*). 
    - This is accomplished by breaking up the `ERC20BridgeSampler` contract into independent, standalone contracts per source, one for buys and one for sells. I only had time to do this for UniswapV3 in this pass, but it serves to illustrate the point. 
- Reduce the huge number of exported types by keeping definitions close to the place where they're actually used. I already pruned a bunch of dead code through this process.
- Misc improvements (check my code comments).
- Head's up: AS (`SwapQuoter`) needs to be created via `SwapQuoter.createAsync()` (not a constructor) because things need to do stuff on init now.

We also now on the way to a more intuitive folder hierarchy:
```
src
├── artifacts.ts
├── constants.ts
├── errors.ts
├── globals.d.ts
├── index.ts
├── network
│   ├── chain.ts
│   ├── constants.ts
│   ├── native_order_utils.ts
│   ├── network_utils.ts
│   ├── samplers
│   │   ├── balancer.ts
│   │   ├── balancer_v2.ts
│   │   ├── bancor.ts
│   │   ├── curve.ts
│   │   ├── dodo_v1.ts
│   │   ├── dodo_v2.ts
│   │   ├── index.ts
│   │   ├── kyber_dmm.ts
│   │   ├── kyber.ts
│   │   ├── lido.ts
│   │   ├── liquidity_provider.ts
│   │   ├── maker_psm.ts
│   │   ├── mooniswap.ts
│   │   ├── mstable.ts
│   │   ├── oasis.ts
│   │   ├── shell.ts
│   │   ├── smoothy.ts
│   │   ├── uniswap_v1.ts
│   │   ├── uniswap_v2.ts
│   │   ├── uniswap_v3.ts
│   │   └── utils
│   │       ├── balancer_sor_v2.ts
│   │       ├── bancor_service.ts
│   │       └── pools_cache.ts
│   ├── sassy_sampler.ts
│   ├── source_filters.ts
│   ├── source_sampler.ts
│   ├── token_adjacency_graph_builder.ts
│   ├── tokens.ts
│   ├── two_hop_sampler.ts
│   ├── types.ts
│   └── utils.ts
├── quote_consumers
│   ├── exchange_proxy_swap_quote_consumer.ts
│   ├── multiplex_encoders.ts
│   ├── quote_consumer_utils.ts
│   └── swap_quote_consumer.ts
├── swap_quoter.ts
├── types.ts
├── utils
│   ├── affiliate_fee_utils.ts
│   ├── alt_mm_implementation_utils.ts
│   ├── assert.ts
│   ├── market_operation_utils
│   │   ├── comparison_price.ts
│   │   ├── constants.ts
│   │   ├── fills.ts
│   │   ├── index.ts
│   │   ├── multihop_utils.ts
│   │   ├── orders.ts
│   │   ├── path_optimizer.ts
│   │   ├── path.ts
│   │   ├── rate_utils.ts
│   │   ├── sampler_utils.ts
│   │   └── types.ts
│   ├── protocol_fee_utils.ts
│   ├── quote_report_generator.ts
│   ├── quote_requestor.ts
│   ├── quote_simulation.ts
│   ├── rfq_maker_blacklist.ts
│   ├── rfqt_mocker.ts
│   └── utils.ts
└── wrappers.ts
```

Next phase(s):
- Break up `ERC20BridgeSampler` in the fashion of `UniswapV3SellSampler` and `UniswapV3BuySampler`.
- Clean up Curve some more.
- Give the quote serialization side of AS a similar treatment. Right now they're all still a big mess in `orders.ts` and `ExchangeProxyQuoteConsumer`.
 
## Checklist:


packages/asset-swapper/src/network/chain.ts
```diff
@@ -0,0 +1,314 @@
...
+    cacheTimeMs: number;
+}
+
+export interface Chain {
+    chainId: ChainId;
+    provider: SupportedProvider;
+    ethCall(opts: ChainEthCallOpts): Promise<Bytes>;
+}
+
+export class LiveChain implements Chain {
```
dorothy-zbornak, line 67: This is the magic auto call batching class.

packages/asset-swapper/src/network/chain.ts
```diff
@@ -0,0 +1,314 @@
...
+
+    public get provider(): SupportedProvider {
+        return this._w3.getProvider();
+    }
+
+    public get chainId(): ChainId {
+        return this._chainId;
+    }
+
+    public async ethCall(opts: ChainEthCallOpts): Promise<Bytes> {
```
dorothy-zbornak, line 102: People call this function to queue up a call that'll get batched up and executed after every tick

packages/asset-swapper/src/network/chain.ts
```diff
@@ -0,0 +1,314 @@
...
+        let c: QueuedEthCall | undefined;
+        const p = new Promise<Bytes>((accept, reject) => {
```
dorothy-zbornak, line 104: This promise gets resolved after the call finally executes.

packages/asset-swapper/src/network/chain.ts
```diff
@@ -0,0 +1,314 @@
...
+        }
+        const callInfos = calls.map(c =>
+            ({
+                data: c.opts.data,
+                to: c.opts.to,
+                gas: new BigNumber(c.opts.gas || 0),
+                value: c.opts.value || ZERO_AMOUNT,
+            }),
+        );
+        return {
```
dorothy-zbornak, line 233: Batched calls go through the new `CallDispatcher`.

packages/asset-swapper/src/network/chain.ts
```diff
@@ -0,0 +1,314 @@
...
+            if (a && b && a.code !== b.code) {
+                return false;
+            }
+        }
+    }
+    return true;
+}
+
+function hashCall(c: ChainEthCallOpts): string {
+    return crypto.createHash('sha1').update(JSON.stringify(c)).digest('hex');
```
dorothy-zbornak, line 303: Supposedly this is the fastest collision-resistant, out-of-the-box hash function.

packages/asset-swapper/src/network/two_hop_sampler.ts
```diff
@@ -0,0 +1,185 @@
...
+import { ContractHelper, createContractWrapperAndHelper } from './utils';
+import { ZERO_AMOUNT, MAX_UINT256 } from './constants';
+
+export interface TwoHopFillData extends FillData {
+    firstHop: DexSample;
+    secondHop: DexSample;
+    intermediateToken: Address;
+}
+
+export class TwoHopSampler {
```
dorothy-zbornak, line 16: This sampler class is not like the other samplers. Totally different interface and needs to be called separately in `SassySampler`.

packages/asset-swapper/contracts/src/ERC20BridgeSampler.sol
```diff
@@ -61,14 +58,11 @@ contract ERC20BridgeSampler is
...
     MooniswapSampler,
     MultiBridgeSampler,
-    NativeOrderSampler,
     ShellSampler,
     SmoothySampler,
     TwoHopSampler,
     UniswapSampler,
-    UniswapV2Sampler,
-    UniswapV3Sampler,
-    UtilitySampler
```
dorothy-zbornak, line 71: `NativeOrderSampler` and `UtilitySampler` are now standalone contracts: `NativeOrderUtils` and `NetworkUtils`.

packages/asset-swapper/contracts/src/TwoHopSampler.sol
```diff
@@ -27,96 +27,100 @@ contract TwoHopSampler {
     using LibBytesV06 for bytes;
```
dorothy-zbornak, line 27: Should be the same logic as before, just tweaked the inputs/outputs and did some light refactors.

packages/asset-swapper/contracts/src/uniswapv3/UniswapV3BuySampler.sol
```diff
@@ -0,0 +1,82 @@
+// SPDX-License-Identifier: Apache-2.0
```
dorothy-zbornak, line 1: Buy-only sampler contract for UniswapV3.

packages/asset-swapper/src/network/chain.ts
```diff
@@ -0,0 +1,314 @@
...
+        nextQueue = [];
+        const batch = [];
+        for (const c of _queue) {
+            if (canBatchCallWith(c, batch)) {
+                batch.push(c);
+            } else {
+                nextQueue.push(c);
+            }
+        }
+        yield batch;
```
dorothy-zbornak, line 261: We can also arbitrarily split calls into smaller RPC chunks with some "is this batch too big now" heuristic here.

packages/asset-swapper/src/network/constants.ts
```diff
@@ -0,0 +1,12 @@
+import { BigNumber } from '@0x/utils';
+
```
dorothy-zbornak, line 2: A bunch of these are duped. Can clean up in another pass.

packages/asset-swapper/src/network/native_order_utils.ts
```diff
@@ -0,0 +1,47 @@
...
+import { getContractAddressesForChainOrThrow } from '@0x/contract-addresses';
+
+import { SignedNativeOrder } from '../types';
+import { NativeOrderUtilsContract } from '../wrappers';
+
+import { Chain } from './chain';
+import { Address } from './types';
+import { ContractHelper, createContractWrapperAndHelper } from './utils';
+
+export class NativeOrderUtils {
```
dorothy-zbornak, line 12: We can break up on-chain helpers into distinct classes/helpers that aren't aware of each other now because `ethCall()` auto-batches for us.

packages/asset-swapper/src/network/samplers/balancer.ts
```diff
@@ -0,0 +1,250 @@
+import { ChainId } from '@0x/contract-addresses';
```
dorothy-zbornak, line 1: Balancer is an example of a source that that relies on off-chain queries and caching, but now it's neatly scoped to the balancer integration so no else one has to know or worry about it.

packages/asset-swapper/src/network/samplers/balancer.ts
```diff
@@ -0,0 +1,250 @@
...
+type SellContract = ERC20BridgeSamplerContract;
+type BuyContract = ERC20BridgeSamplerContract;
+type SellContractSellFunction = SellContract['sampleSellsFromBalancer'];
+type BuyContractBuyFunction = BuyContract['sampleBuysFromBalancer'];
+type SellContractSellFunctionParameters = Parameters<SellContractSellFunction>;
+type BuyContractBuyFunctionParameters = Parameters<BuyContractBuyFunction>;
+type SellContractSellFunctionReturn = UnwrapContractFunctionReturnType<ReturnType<SellContractSellFunction>>;
+type BuyContractBuyFunctionReturn = UnwrapContractFunctionReturnType<ReturnType<BuyContractBuyFunction>>;
+type SamplerSellEthCall = SamplerEthCall<BalancerFillData,SellContractSellFunctionParameters,SellContractSellFunctionReturn>;
+type SamplerBuyEthCall = SamplerEthCall<BalancerFillData,BuyContractBuyFunctionParameters,BuyContractBuyFunctionReturn>;
```
dorothy-zbornak, line 153: This is bit of unfortunately required boilerplate for each source to make the rest of the file more legible and nudge generic type inference along.

dorothy-zbornak: It's pretty easy to copy & paste & update for new sources though.

packages/asset-swapper/src/network/samplers/balancer.ts
```diff
@@ -0,0 +1,250 @@
...
+
+    protected constructor(chain: Chain, public readonly fork: ERC20BridgeSource, private readonly _cache: PoolsCache) {
+        super({
+            chain,
+            sellSamplerContractArtifactName: 'ERC20BridgeSampler',
+            buySamplerContractArtifactName: 'ERC20BridgeSampler',
+            sellSamplerContractType: ERC20BridgeSamplerContract,
+            buySamplerContractType: ERC20BridgeSamplerContract,
+            sellContractSellFunctionName: 'sampleSellsFromBalancer',
+            buyContractBuyFunctionName: 'sampleBuysFromBalancer',
```
dorothy-zbornak, line 190: If anyone can figure out how to infer these values from the prior `type` boilerplate, that'd be cool.

packages/asset-swapper/src/network/samplers/curve.ts
```diff
@@ -0,0 +1,709 @@
+import { ChainId } from '@0x/contract-addresses';
```
dorothy-zbornak, line 1: Curve is still heinous but at least it's only heinous to other curves now. We can refactor all these configs out into separate files under a curve subdir.

packages/asset-swapper/src/network/samplers/uniswap_v3.ts
```diff
@@ -0,0 +1,122 @@
...
+);
+
+export interface UniswapV3FillData extends FillData {
+    tokenAddressPath: Address[];
+    router: Address;
+    uniswapPath: Bytes;
+}
+
+type SellContract = UniswapV3SellSamplerContract;
+type BuyContract = UniswapV3BuySamplerContract;
```
dorothy-zbornak, line 40: This is the only source I "fully" upgraded. I.e., it has separate, standalone sampler contracts for buys and sells (not just reusing `ERC20BridgeSampler`).

packages/asset-swapper/src/network/samplers/utils/pools_cache.ts
```diff
@@ -27,8 +32,17 @@ export abstract class PoolsCache {
         makerToken: string,
         timeoutMs: number = DEFAULT_TIMEOUT_MS,
     ): Promise<Pool[]> {
-        const timeout = new Promise<Pool[]>(resolve => setTimeout(resolve, timeoutMs, []));
-        return Promise.race([this._getAndSaveFreshPoolsForPairAsync(takerToken, makerToken), timeout]);
+        // Only allow one outstanding refresh per pair at a time.
```
dorothy-zbornak, line 35: idk if this makes a difference but seemed reasonable and wasn't particularly hard?

packages/asset-swapper/src/network/samplers/utils/pools_cache.ts
```diff
@@ -45,6 +59,9 @@ export abstract class PoolsCache {
             return undefined;
         }
         if (PoolsCache._isExpired(value)) {
+            // Auto-refresh the pools if expired.
+            this.getFreshPoolsForPairAsync(takerToken, makerToken)
+                .catch(err => console.error(err));
```
dorothy-zbornak, line 64: Previously we would manually reach into the sampler and refresh the pools for the polled tokens in `MarketOperationUtils` (yuck). Now we just auto-refresh every time pairs are touched.

packages/asset-swapper/src/network/sassy_sampler.ts
```diff
@@ -0,0 +1,295 @@
...
+    sources: ERC20BridgeSource[];
+    tokenAdjacencyGraph: TokenAdjacencyGraph;
+    liquidityProviderRegistry?: LiquidityProviderRegistry;
+}
+
+type SassySamplerCreateOpts = Partial<SassySamplerCreateFullOpts> & { chain: Chain };
+
+const DEFAULT_SOURCES = SourceFilters.all().exclude([ERC20BridgeSource.Native, ERC20BridgeSource.MultiHop]).sources;
+
+export class SassySampler {
```
dorothy-zbornak, line 46: I honestly couldn't think of a good name.

packages/asset-swapper/src/network/sassy_sampler.ts
```diff
@@ -0,0 +1,295 @@
...
+        return getIntermediateTokens(takerToken, makerToken, this._tokenAdjacencyGraph);
+    }
+}
+
+interface CreateSourceSamplerOpts {
+    chain: Chain;
+    liquidityProviderRegistry?: LiquidityProviderRegistry;
+}
+
+async function createSourceSamplerAsync(source: ERC20BridgeSource, opts: CreateSourceSamplerOpts): Promise<SourceSampler> {
```
dorothy-zbornak, line 209: When you add a new source, don't forget to add to this create function.

packages/asset-swapper/src/network/source_sampler.ts
```diff
@@ -0,0 +1,216 @@
...
+
+export interface SamplerEthCall<TFillData, TSamplerFunctionArgs extends any[], TSamplerFunctionReturn> {
+    args: TSamplerFunctionArgs;
+    getDexSamplesFromResult(result: TSamplerFunctionReturn): DexSample<TFillData>[];
+}
+
+export type UnwrapContractFunctionReturnType<T> = T extends ContractFunctionObj<infer U> ? U : T;
+
+// Base class for a standard sampler with on-chain quote functions.
+export abstract class OnChainSourceSampler<
```
dorothy-zbornak, line 59: All individual source samplers currently inherit from this class for convenience, but it isn't required. In the future we could have (e.g.) totally off-chain samplers that just inherit from the less opinionated `SourceSamplerBase` and do everything off-chain.

packages/asset-swapper/src/network/utils.ts
```diff
@@ -0,0 +1,211 @@
...
+            decodeOutput: (fnName: string, data: string) => fastAbi.decodeOutput(fnName, data),
+        },
+    );
+    return [
+        wrapper,
+        new ContractHelper(chain, wrapper),
+    ];
+}
+
+export class ContractHelper<TBaseContract extends GeneratedContract> {
```
dorothy-zbornak, line 98: Any time you want to do batching-compatible calls on a contract (wrapper), you should use it with a `ContractHelper` instance. It'll automatically include the correct overrides for the eth_call and makes things typesafe.

packages/asset-swapper/src/network/utils.ts
```diff
@@ -0,0 +1,211 @@
...
+        [ChainId.Kovan]: defaultValue,
+        [ChainId.Ganache]: defaultValue,
+        [ChainId.BSC]: defaultValue,
+        [ChainId.Polygon]: defaultValue,
+        [ChainId.PolygonMumbai]: defaultValue,
+        ...(rest || {}),
+    };
+}
+
+export function createContractWrapperAndHelper<TContract extends GeneratedContract>(
```
dorothy-zbornak, line 70: This function creates both a contract wrapper and a contract helper (see below) instance for a contract you want to be compatible with batch calls and overrides.

dorothy-zbornak: The `OnChainSampler` base class inherited by all the sources abstracts all this stuff away from you though if you're just adding a new source.

packages/asset-swapper/src/network/utils.ts
```diff
@@ -0,0 +1,211 @@
...
+    artifactName: string,
+    address?: Address,
+): [TContract, ContractHelper<TContract>] {
+    const fastAbi = new FastABI(
+        contractType.ABI() as MethodAbi[],
+        { BigNumber },
+    );
+    const artifact = (artifacts as ArtifactsMap)[artifactName];
+    const wrapper = new contractType(
+        address || getDeterministicContractAddressFromArtifact(artifact),
```
dorothy-zbornak, line 82: Undeployed contracts will get a deterministic address based on its bytecode. This keeps the batcher from uploading multiple copies of the same contract, since they'll all collapse into a single override address.

packages/asset-swapper/src/quote_consumers/exchange_proxy_swap_quote_consumer.ts
```diff
@@ -549,13 +575,14 @@ export class ExchangeProxySwapQuoteConsumer implements SwapQuoteConsumerBase {
...
                     });
                     break switch_statement;
-                case ERC20BridgeSource.UniswapV3:
-                    const fillData = (order as OptimizedMarketBridgeOrder<FinalUniswapV3FillData>).fillData;
-                    wrappedBatchCalls.push({
-                        selector: this._exchangeProxy.getSelector('sellTokenForTokenToUniswapV3'),
-                        sellAmount: order.takerAmount,
-                        data: fillData.uniswapPath,
-                    });
+                case ERC20BridgeSource.UniswapV3: {
```
dorothy-zbornak, line 578: Just wrapped in braces.

packages/asset-swapper/src/network/types.ts
```diff
@@ -0,0 +1,107 @@
...
+export interface FillData {}
+
+export interface DexSample<TFillData extends FillData = FillData> {
+    source: ERC20BridgeSource;
+    fillData: TFillData;
+    input: BigNumber;
+    output: BigNumber;
+}
+
+export interface MultiHopCallInfo {
```
dorothy-zbornak, line 73: I kept the confusing mix of `TwoHop` and `MultiHop` naming. This is because some stuff is technically compatible with a an any-hop system and we may be heading in that direction.

packages/asset-swapper/contracts/src/TwoHopSampler.sol
```diff
@@ -27,96 +27,100 @@ contract TwoHopSampler {
...
-            uint256 buyAmount
-        )
+        returns (TwoHopResult memory result)
     {
         uint256 intermediateAssetAmount = 0;
         for (uint256 i = 0; i != firstHopCalls.length; ++i) {
-            firstHopCalls[i].writeUint256(firstHopCalls[i].length - 32, sellAmount);
-            (bool didSucceed, bytes memory returnData) = address(this).call(firstHopCalls[i]);
+            bytes memory data = firstHopCalls[i].data;
+            data.writeUint256(data.length - 32, sellAmount);
```
dorothy-zbornak, line 53: I forgot what a completely insane and fragile hack this two-hop stuff was. I'm thinking every sampler should expose a new function like `sampleMultiHopSell(bytes arbitraryData, uint256 takerAmount)` so we aren't making the huge assumption that they will always have amounts encoded as the final word.

packages/asset-swapper/contracts/src/CallDispatcher.sol
```diff
@@ -0,0 +1,51 @@
...
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+*/
+
+pragma solidity ^0.6;
+pragma experimental ABIEncoderV2;
+
+
+contract CallDispatcher {
```
dorothy-zbornak, line 24: This is the contract that executes auto-batched calls.

packages/asset-swapper/src/network/chain.ts
```diff
@@ -0,0 +1,314 @@
...
+        let c: QueuedEthCall | undefined;
+        const p = new Promise<Bytes>((accept, reject) => {
+            // This executes right away so `c` will be defined in this fn.
+            c = {
+                opts,
+                accept,
+                reject,
+            };
+        });
+        if (opts.maxCacheAgeMs) {
```
dorothy-zbornak, line 112: I don't have anything set up to use caching rn, but soon?

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -80,85 +72,46 @@ export class SwapQuoter {
      * Instantiates a new SwapQuoter instance
      * @param   supportedProvider   The Provider instance you would like to use for interacting with the Ethereum network.
      * @param   orderbook           An object that conforms to Orderbook, see type for definition.
-     * @param   options             Initialization options for the SwapQuoter. See type definition for details.
+     * @param   opts                Initialization opts for the SwapQuoter. See type definition for details.
      *
      * @return  An instance of SwapQuoter
      */
-    constructor(supportedProvider: SupportedProvider, orderbook: Orderbook, options: Partial<SwapQuoterOpts> = {}) {
+    public static async createAsync(supportedProvider: SupportedProvider, orderbook: Orderbook, opts: Partial<SwapQuoterOpts> = {}): Promise<SwapQuoter> {
```
dorothy-zbornak, line 79: This is how you create `SwapQuoter` now.

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -80,85 +72,46 @@ export class SwapQuoter {
...
+        const chain = await LiveChain.createAsync({ provider: providerUtils.standardizeOrThrow(supportedProvider) });
```
dorothy-zbornak, line 80: `LiveChain` will automatically fetch the chainId so we can almost get rid of it from `SwapQuoterOpts`. We can't today because the `ExchangeProxyConsumer` still uses it.

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -80,85 +72,46 @@ export class SwapQuoter {
...
-        };
         this._protocolFeeUtils = ProtocolFeeUtils.getInstance(
             constants.PROTOCOL_FEE_UTILS_POLLING_INTERVAL_IN_MS,
-            options.ethGasStationUrl,
-        );
-        // Allow the sampler bytecode to be overwritten using geths override functionality
-        const samplerBytecode = _.get(artifacts.ERC20BridgeSampler, 'compilerOutput.evm.deployedBytecode.object');
-        // Allow address of the Sampler to be overridden, i.e in Ganache where overrides do not work
-        const samplerAddress = (options.samplerOverrides && options.samplerOverrides.to) || SAMPLER_ADDRESS;
-        const defaultCodeOverrides = samplerBytecode
```
dorothy-zbornak, line 119: No more external overrides. Can API tests recover?

packages/asset-swapper/src/network/chain.ts
```diff
@@ -0,0 +1,314 @@
...
+        if (opts.immediate) {
+            // Do not add to queue. Dispatch immediately.
+            this._executeAsync([c!]);
+        } else {
+            // Just queue up and batch dispatch later.
+            this._queue.push(c!);
+        }
+        // Cache both success and failure results.
+        p.then(resultData => this._cacheCallResult(c!.opts, resultData))
+            .catch(error => this._cacheCallResult(c!.opts, undefined, error));
```
dorothy-zbornak, line 129: Maybe only do this if `maxCacheAgeMs > 0`?

packages/asset-swapper/src/network/source_sampler.ts
```diff
@@ -0,0 +1,216 @@
...
+
+    protected abstract _getSellQuoteCallsAsync(
+        tokenAddressPath: string[],
+        takerFillAmounts: BigNumber[],
+    ): Promise<SamplerEthCall<TFillData, TSellSamplerFunctionArgs, TSellSamplerFunctionReturn>[]>;
+
+    protected abstract _getBuyQuoteCallsAsync(
+        tokenAddressPath: string[],
+        takerFillAmounts: BigNumber[],
+    ): Promise<SamplerEthCall<TFillData, TBuySamplerFunctionArgs, TBuySamplerFunctionReturn>[]>;
```
dorothy-zbornak, line 197: Most source samplers just need to override these two functions, and maybe `canConvertTokens()`.

packages/asset-swapper/src/network/types.ts
```diff
@@ -0,0 +1,107 @@
+import { BigNumber } from '@0x/utils';
+
+import { ChainEthCallOverrides } from './chain';
+
+export type Bytes = string;
+export type Address = Bytes;
```
dorothy-zbornak, line 6: Introducing these cosmetic aliases for clarity.

packages/asset-swapper/src/network/sassy_sampler.ts
kimpers: We can probably use TS to enforce that all `ERC20BridgeSource` values are covered here which would be very neat 🤔

packages/asset-swapper/src/network/utils.ts
kimpers: Might want to put some nice jsdoc comment here so that it's easier to remember 😃

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -194,11 +147,11 @@ export class SwapQuoter {
             }
         });
 
-        const opts = { ...constants.DEFAULT_SWAP_QUOTE_REQUEST_OPTS, ...options };
+        const _opts = { ...constants.DEFAULT_SWAP_QUOTE_REQUEST_OPTS, ...opts };
```
kimpers, line 150: I know this is a preexisting condition but could we name this better? `opts` and `_opts` is very confusing. Maybe `optsWithDefaults` or smth?

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -322,29 +268,29 @@ export class SwapQuoter {
...
         assetFillAmount: BigNumber,
         marketOperation: MarketOperation,
-        options: Partial<SwapQuoteRequestOpts>,
+        opts: Partial<SwapQuoteRequestOpts>,
     ): Promise<SwapQuote> {
         assert.isETHAddressHex('makerToken', makerToken);
         assert.isETHAddressHex('takerToken', takerToken);
         assert.isBigNumber('assetFillAmount', assetFillAmount);
-        const opts = _.merge({}, constants.DEFAULT_SWAP_QUOTE_REQUEST_OPTS, options);
+        const _opts = _.merge({}, constants.DEFAULT_SWAP_QUOTE_REQUEST_OPTS, opts);
```
kimpers, line 276: Same confusion here 😅

packages/asset-swapper/test/dex_sampler_test.ts
```diff
@@ -1,657 +1,657 @@
...
+//                     dexOrderSampler.getLimitOrderFillableMakerAmounts(ORDERS, exchangeProxyAddress),
+//                     dexOrderSampler.getLimitOrderFillableTakerAmounts(ORDERS, exchangeProxyAddress),
+//                 );
+//                 expect(fillableMakerAmounts).to.deep.eq(expectedFillableMakerAmounts);
+//                 expect(fillableTakerAmounts).to.deep.eq(expectedFillableTakerAmounts);
+//             });
+//         });
+//     });
+// });
+// // tslint:disable-next-line: max-file-line-count
```
kimpers, line 657: Who needs tests anyway 😂 🙈

kimpers (commented):
This is one massive refactor 🤯 Looks really good! Great job! 🙏 🧙

packages/asset-swapper/src/network/utils.ts
dorothy-zbornak: But if you're looking at this class you're probably aware. The warning really needs to be near where someone would be tempted to call `callAsync()`. Maybe in `NetworkUtils` or `NativeOrderUtils`. For dealing with regular samplers it's all abstracted away atm.

packages/asset-swapper/package.json
```diff
@@ -89,7 +89,8 @@
         "graphql": "^15.4.0",
         "graphql-request": "^3.4.0",
         "heartbeats": "^5.0.1",
-        "lodash": "^4.17.11"
+        "lodash": "^4.17.11",
+        "node-fetch": "^2.6.1"
```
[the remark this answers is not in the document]
dorothy-zbornak: Sounds good. I was actually confused as to how we were using `fetch()` without any dep so I added this in. I'm still confused as to why it only started being an issue for me during this refactor.

packages/asset-swapper/src/network/chain.ts
```diff
@@ -0,0 +1,314 @@
...
+                data: c.opts.data,
+                to: c.opts.to,
+                gas: new BigNumber(c.opts.gas || 0),
+                value: c.opts.value || ZERO_AMOUNT,
+            }),
+        );
+        return {
+            gas: calls.reduce((s, c) => (c.opts.gas || DEFAULT_CALL_GAS_LIMIT) + s, 0) || DEFAULT_CALL_GAS_LIMIT,
+            from: DEFAULT_CALLER_ADDRESS,
+            gasPrice: BigNumber.sum(...calls.map(c => c.opts.gasPrice || 0)),
```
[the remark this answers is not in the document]
dorothy-zbornak: Lol nice catch. Luckily nothing currently cares about gas price.

packages/asset-swapper/test/dex_sampler_test.ts
dorothy-zbornak: Oh I actually meant to delete this file lol. Yeah we need to do a testing sprint when the dust settles. So much of the code is changing so rapidly rn, and we were already extremely reliant on simbot, that I figure we're better off without them for the moment.

packages/asset-swapper/src/network/sassy_sampler.ts
dorothy-zbornak: Oh good idea. Maybe if I refactor so that this function returns a map of samplers.

packages/asset-swapper/src/network/chain.ts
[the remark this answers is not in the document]
dekz: Did notice issues with `gasPrice` and `eth_call` when using Geth? There was some combo which sucked for me (maybe it was gasPrice, gas and from)

packages/asset-swapper/src/network/chain.ts
```diff
@@ -0,0 +1,319 @@
...
+        }
+        yield batch;
+    }
+}
+
+function resolveQueuedCall(c: QueuedEthCall, r: DispatchedCallResult): void {
+    if (r.success) {
+        c.accept(r.resultData);
+    } else {
+        c.reject(tryDecodeStringRevertErrorResult(r.resultData) || `EVM call reverted`);
```
dekz, line 269: Maybe reject it with the data?
```suggestion
        c.reject(tryDecodeStringRevertErrorResult(r.resultData) || `EVM call reverted: ${r.resultData}`);
```

packages/asset-swapper/src/network/samplers/curve.ts
```diff
@@ -0,0 +1,709 @@
...
+        CurveFillData
+    >
+{
+    public static async createAsync(
+        chain: Chain,
+        fork: ERC20BridgeSource,
+    ): Promise<CurveSampler> {
+        const curveInfos = CURVELIKE_INFOS_BY_CHAIN_ID[chain.chainId];
+        if (!curveInfos) {
+            throw new Error(`No curve configs for chain ${chain.chainId}`);
```
dekz, line 609: Is it bad do have sitting around with empty CurveInfos to act upon?

packages/asset-swapper/src/network/samplers/dodo_v2.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    protected constructor(chain: Chain, private readonly _factories: Address[]) {
+        super({
+            chain,
+            sellSamplerContractArtifactName: 'ERC20BridgeSampler',
+            buySamplerContractArtifactName: 'ERC20BridgeSampler',
+            sellSamplerContractType: ERC20BridgeSamplerContract,
+            buySamplerContractType: ERC20BridgeSamplerContract,
+            sellContractSellFunctionName: 'sampleSellsFromDODOV2',
```
dekz, line 74: Can we make some magic assumptions here with defaults?

```
sellContractSellFunctionName: `sampleSellsFrom${this.source}`
sellContractBuyFunctionName: `sampleBuysFrom${this.source}`
```

Too much of an assumption?

packages/asset-swapper/src/network/samplers/dodo_v2.ts
```diff
@@ -0,0 +1,134 @@
...
+                    makerToken,
+                    takerFillAmounts,
+                ],
+                getDexSamplesFromResult: ([isSellBase, poolAddress, samples]) => {
+                    return takerFillAmounts.map((a, i) => ({
+                        source: ERC20BridgeSource.DodoV2,
+                        fillData: { poolAddress, isSellBase },
+                        input: a,
+                        output: samples[i],
+                    }));
```
dekz, line 103: Maybe we can have some helper magic defaults, which can be overridden in special cases 
```suggestion
                    return this.buildSamples(takerFillAmounts, samples, { poolAddress, isSellBase });
```

Does anything even really change over the sources here, maybe it's always just `fillData` construction ?

If we modify SwapRevertSampler to return the required encoded `fillData` then there might be even more cleanup available to us here.

packages/asset-swapper/src/network/samplers/uniswap_v3.ts
dekz: Do you think the Buy/Sell split is still impactful when SwapRevertSampler re-uses the sell for ApproximateBuys ?

packages/asset-swapper/src/network/samplers/uniswap_v3.ts
```diff
@@ -0,0 +1,122 @@
...
+type SellContractSellFunction = SellContract['sampleSells'];
+type BuyContractBuyFunction = BuyContract['sampleBuys'];
+type SellContractSellFunctionParameters = Parameters<SellContractSellFunction>;
+type BuyContractBuyFunctionParameters = Parameters<BuyContractBuyFunction>;
+type SellContractSellFunctionReturn = UnwrapContractFunctionReturnType<ReturnType<SellContractSellFunction>>;
+type BuyContractBuyFunctionReturn = UnwrapContractFunctionReturnType<ReturnType<BuyContractBuyFunction>>;
+type SamplerSellEthCall = SamplerEthCall<UniswapV3FillData,SellContractSellFunctionParameters,SellContractSellFunctionReturn>;
+type SamplerBuyEthCall = SamplerEthCall<UniswapV3FillData,BuyContractBuyFunctionParameters,BuyContractBuyFunctionReturn>;
```
dekz, line 48: Nuking this boilerplate would be pretty good 😂

dekz (commented):
Gave it a first pass, gonna need to set aside some more time haha. I love the co-location of the config with the source.

I need to get a mental model of how many RPCs could feasibly occur (Alchemy is fine with compute units, Infura is rq/day, rate limits on both). 

The potential future dynamic "Sampler/s" is interesting, especially in conjunction with "Deregistering" sources, e.g Eth2Dai would eventually totally disappear (not sampled and code for it not supplied).

Need to cut some of this typescript type boilerplate though

packages/asset-swapper/src/network/chain.ts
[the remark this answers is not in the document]
dorothy-zbornak: Seemed fine for me

packages/asset-swapper/src/network/samplers/curve.ts
dorothy-zbornak: Not bad, just pointless. I'm pretty inconsistent on checking for valid configs in these samplers.

packages/asset-swapper/src/network/samplers/dodo_v2.ts
dorothy-zbornak: Yeah but it breaks the type safety (key in the contract wrapper)

packages/asset-swapper/src/network/samplers/dodo_v2.ts
dorothy-zbornak: Idk I was also thinking of just fetching reserve amounts for uni2 and computing here but I guess that breaks with revert sampling

dorothy-zbornak: Some (uni3) have different fill data per sample

packages/asset-swapper/src/network/samplers/uniswap_v3.ts
dorothy-zbornak: I tried so hard but ultimately gave up. If only TS had preprocessors.

packages/asset-swapper/src/network/chain.ts
dorothy-zbornak: Yeah unless we maybe snapped sample amounts.

packages/asset-swapper/src/network/samplers/uniswap_v3.ts
dorothy-zbornak: Yeah I probably won't keep it for swap revert.

packages/asset-swapper/src/network/utils.ts
mzhu25: nice

packages/asset-swapper/src/network/sassy_sampler.ts
```diff
@@ -0,0 +1,295 @@
...
+    public async getMedianSellRateAsync(
+        sources: ERC20BridgeSource[],
+        takerToken: Address,
+        makerToken: Address,
+        takerAmount: BigNumber,
+    ): Promise<BigNumber> {
+        if (takerToken.toLowerCase() === makerToken.toLowerCase()) {
+            return new BigNumber(1);
+        }
+        const ssamples = (
```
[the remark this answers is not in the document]
mzhu25: or was this intentional? 🤔

mzhu25 (approved):
whew that was a doozy but definitely a huge improvement! thanks for undertaking this 😁

packages/asset-swapper/src/network/samplers/bancor.ts
```diff
@@ -0,0 +1,125 @@
...
+
+import { ERC20BridgeSamplerContract } from '../../wrappers';
+
+import { Chain } from '../chain';
+import { NULL_ADDRESS } from '../constants';
+import { OnChainSourceSampler, UnwrapContractFunctionReturnType, SamplerEthCall } from '../source_sampler';
+import { Address, ERC20BridgeSource, FillData } from "../types";
+import { valueByChainId } from '../utils';
+
+import { BancorService } from './utils/bancor_service';
```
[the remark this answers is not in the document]
dorothy-zbornak: no, good idea

packages/asset-swapper/src/network/sassy_sampler.ts
```diff
@@ -0,0 +1,295 @@
...
+    const tokens = [takerToken, makerToken].map(t => t.toLowerCase());
+    const takerIntermediateTokens = tokenAdjacencyGraph[tokens[0]] || tokenAdjacencyGraph.default;
+    const makerIntermediateTokens = tokenAdjacencyGraph[tokens[1]] || tokenAdjacencyGraph.default;
+    // Filter intermediate tokens common to both.
+    const intermediateTokens = takerIntermediateTokens.filter(t => makerIntermediateTokens.includes(t));
+    // Select unique ones that aren't the maker or taker token.
+    return Object.keys(Object.assign(
+        {},
+        ...intermediateTokens.map(t => ({[t]: true})),
+    )).filter(t => !tokens.includes(t));
```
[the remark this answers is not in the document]
dorothy-zbornak: beautiful

packages/asset-swapper/src/network/source_sampler.ts
```diff
@@ -0,0 +1,214 @@
...
+    sellContractSellFunctionName: TSellSamplerFunctionName,
+    buyContractBuyFunctionName: TBuySamplerFunctionName,
+}
+
+export interface SamplerEthCall<TFillData, TSamplerFunctionArgs extends any[], TSamplerFunctionReturn> {
+    args: TSamplerFunctionArgs;
+    getDexSamplesFromResult(result: TSamplerFunctionReturn): DexSample<TFillData>[];
+}
+
+export type UnwrapContractFunctionReturnType<T> = T extends ContractFunctionObj<infer U> ? U : T;
```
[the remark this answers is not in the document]
dorothy-zbornak: what about just
```ts
export type UnwrapContractFunctionReturnType<T> = T extends ContractFunctionObj<infer U> ? U : never;`
```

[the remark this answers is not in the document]
mzhu25: yeah that's good too

packages/asset-swapper/src/network/source_sampler.ts
```diff
@@ -0,0 +1,214 @@
...
+    protected readonly _sellContractFunction: TSellSamplerFunction;
+    protected readonly _buyContractFunction: TBuySamplerFunction;
+
+    protected constructor(opts: TOpts) {
+        super();
+        this._chain = opts.chain;
+        [this._sellContract, this._sellContractHelper] = createContractWrapperAndHelper(
+            opts.chain,
+            opts.sellSamplerContractType,
+            opts.sellSamplerContractArtifactName,
```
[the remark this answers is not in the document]
dorothy-zbornak: It doesn't exist yet at this point, but maybe something like `` `/^(.+)Wrapper$/\`.exec(opts.sellSamplerContractType.name)[1]``?

packages/asset-swapper/src/network/source_sampler.ts
```diff
@@ -0,0 +1,214 @@
...
+    sellSamplerContractType: ContractWrapperType<TSellSamplerContract>,
+    buySamplerContractType: ContractWrapperType<TBuySamplerContract>,
+    sellContractSellFunctionName: TSellSamplerFunctionName,
+    buyContractBuyFunctionName: TBuySamplerFunctionName,
+}
+
+export interface SamplerEthCall<TFillData, TSamplerFunctionArgs extends any[], TSamplerFunctionReturn> {
+    args: TSamplerFunctionArgs;
+    getDexSamplesFromResult(result: TSamplerFunctionReturn): DexSample<TFillData>[];
+}
```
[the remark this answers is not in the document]
dorothy-zbornak: MY HERO

packages/asset-swapper/src/network/source_sampler.ts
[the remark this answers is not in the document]
mzhu25: I mean like add `contractName: string` to the `ContractWrapperType<T>` interface, since abi-gen gives each contract wrapper class a `public static contractName`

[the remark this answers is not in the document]
dorothy-zbornak: ahh yeah nice

packages/asset-swapper/src/network/source_sampler.ts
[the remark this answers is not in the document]
dorothy-zbornak: OK this turned out to break type safety in some places so I had to go with something more convoluted but it's the same boilerplate reduction so I'm happy.

PR #260, opened 2021-05-31
feat: Lido StETH deposit integration [TKR-90]

kimpers:
## Description
**Simbot vs prod for ETH, WETH, DAI, STETH**
image

## Checklist:


contracts/zero-ex/contracts/src/transformers/bridges/mixins/MixinLido.sol
```diff
@@ -0,0 +1,75 @@
...
+        bytes memory bridgeData
+    )
+        internal
+        returns (uint256 boughtAmount)
+    {
+        (ILido lido) = abi.decode(bridgeData, (ILido));
+        if (address(buyToken) == address(lido)) {
+            if (address(sellToken) == ETH_ADDRESS) {
+                // TODO(kimpers): do we want to use a referral address here
+                boughtAmount = lido.getPooledEthByShares(lido.submit{ value: sellAmount}(address(0)));
```
kimpers, line 67: Do we think we want to tag this with a referral address for tracking? 🤔

contracts/zero-ex/contracts/src/transformers/bridges/mixins/MixinLido.sol
```diff
@@ -0,0 +1,70 @@
...
+        bytes memory bridgeData
+    )
+        internal
+        returns (uint256 boughtAmount)
+    {
+        (ILido lido) = abi.decode(bridgeData, (ILido));
+        if (address(sellToken) == address(WETH) && address(buyToken) == address(lido)) {
+            WETH.withdraw(sellAmount);
+            // TODO(kimpers): do we want to use a referral address here
+            boughtAmount = lido.getPooledEthByShares(lido.submit{ value: sellAmount}(address(0)));
```
kimpers, line 66: Do we want to set an affiliate address here for tracking purposes? 🤔

dorothy-zbornak: I vote no. We don't do this anywhere else. It might potentially add gas as well?

contracts/zero-ex/contracts/src/transformers/bridges/mixins/MixinLido.sol
```diff
@@ -0,0 +1,70 @@
...
+        }
+
```
dorothy-zbornak, line 68: Should we revert otherwise?

packages/asset-swapper/contracts/src/LidoSampler.sol
```diff
@@ -0,0 +1,95 @@
...
+        uint256 numSamples = takerTokenAmounts.length;
+        makerTokenAmounts = new uint256[](numSamples);
+        if (takerToken != lidoInfo.wethToken || makerToken != address(lidoInfo.stEthToken)) {
+            // Return 0 values if not selling WETH for stETH
+            return makerTokenAmounts;
+        }
+
+        for (uint256 i = 0; i < numSamples; i++) {
+            // Minting stETH is always 1:1
+            makerTokenAmounts[i] = takerTokenAmounts[i];
```
dorothy-zbornak, line 59: Oh interesting. Does lido reward LPs purely with eth2 validator rewards?

dorothy-zbornak (approved):
:+1: :eye: :lips: :eye: :+1:

contracts/zero-ex/contracts/src/transformers/bridges/mixins/MixinLido.sol
kimpers: Makes sense, will add 👍

packages/asset-swapper/contracts/src/LidoSampler.sol
kimpers: Actually, this is quite complex. When you deposit ETH for stETH you get a certain amount of shares, at the time of deposit this translates to 1:1 ETH to stETH, but as the validators collect more staking rewards your shares will be worth more stETH. The `balanceOf` function dynamically translates shares -> stETH 😅 

For LPing on Curve, you also receive their governance token Lido token.

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -1451,6 +1468,15 @@ export const DEFAULT_GAS_SCHEDULE: Required<FeeSchedule> = {
...
         return gas;
     },
+    [ERC20BridgeSource.Lido]: (fillData?: FillData) => {
+        const takerToken = (fillData as LidoFillData).takerToken;
+        // WETH avoids one wrapping and so it's slightly cheaper
+        if (takerToken === MAINNET_TOKENS.WETH) {
+            return 226e3;
+        }
+
+        return 255e3;
```
dekz, line 1478: IMHO this should be handled in `exchangeProxyOverhead` rather than here. It works out to be the same and the intent is the same, but the WETH MixinLido behaves the same regardless. It get's WETH and it withdraws into ETH. Regardless of whether the trade is ETH or WETH.

It's the `exchangeProxyOverhead` which introduces the ETH->WETH deposit overhead.

The way I think about it is, from the perspective of the `BridgeAdapter`, given the choice of two dexes, what is the cost difference? ETH->WETH deposit has already happened so the only option is to take the WETH and withdraw it.

This will be captured in SwapRevertSampling so it makes no matter, just thought I'd drop how I think about it

packages/asset-swapper/contracts/src/LidoSampler.sol
dekz: FWIW I think `return takerTokenAmounts;` here would accomplish the same thing, without having to iterate through the for loop.

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -1392,6 +1431,15 @@ export class SamplerOperations {
                             ...intermediateTokens.map(t => [takerToken, t, makerToken]),
                         ].map(path => this.getUniswapV3SellQuotes(router, quoter, path, takerFillAmounts));
                     }
+                    case ERC20BridgeSource.Lido: {
+                        const lidoInfo = LIDO_INFO_BY_CHAIN[this.chainId];
+
+                        if (lidoInfo.stEthToken === NULL_ADDRESS) {
+                            return [];
```
dekz, line 1438: I would check that takerToken == WETH and makerToken == stETH here. That way we can avoid hitting the sampler whenever these tokens aren't involved. Similar to Curve, where we only check certain curves if they involve those tokens.

packages/asset-swapper/contracts/src/LidoSampler.sol
```diff
@@ -0,0 +1,95 @@
...
+    )
+        public
+        pure
+        returns (uint256[] memory makerTokenAmounts)
+    {
+        _assertValidPair(makerToken, takerToken);
+
+        uint256 numSamples = takerTokenAmounts.length;
+        makerTokenAmounts = new uint256[](numSamples);
+        if (takerToken != lidoInfo.wethToken || makerToken != address(lidoInfo.stEthToken)) {
```
dekz, line 52: I would look at performing this validation earlier. It's a non-zero cost to encode, execute and decode from the sampler.

PR #267, opened 2021-06-28
feat: Balancer V2 Polygon

romainbutteaud:
## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/market_operation_utils/constants.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/constants.ts
@@ -146,6 +146,7 @@ export const SELL_SOURCE_FILTER_BY_CHAIN_ID = valueByChainId<SourceFilters>(
             ERC20BridgeSource.Polydex,
             ERC20BridgeSource.ApeSwap,
             ERC20BridgeSource.FirebirdOneSwap,
+            ERC20BridgeSource.BalancerV2,
         ]),
     },
     new SourceFilters([]),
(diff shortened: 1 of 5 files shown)


romainbutteaud:
Simbot is: `occur-crowded-toys`

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -1303,7 +1306,14 @@ export const LIDO_INFO_BY_CHAIN = valueByChainId<LidoInfo>(
...
 export const BALANCER_TOP_POOLS_FETCHED = 250;
 export const BALANCER_MAX_POOLS_FETCHED = 3;
-export const BALANCER_V2_SUBGRAPH_URL = 'https://api.thegraph.com/subgraphs/name/balancer-labs/balancer-v2';
+
+export const BALANCER_V2_SUBGRAPH_URL_BY_CHAIN = valueByChainId<string>(
+    {
+        [ChainId.Mainnet]: 'https://api.thegraph.com/subgraphs/name/balancer-labs/balancer-v2',
+        [ChainId.Polygon]: 'https://api.thegraph.com/subgraphs/name/balancer-labs/balancer-polygon-v2',
+    },
+    NULL_ADDRESS,
```
kimpers, line 1315: I pretty sure `NULL_ADDRESS` is `0x000....` here, I don't think we can use that as a fallback when it's expecting an URL. This will probably cause issues on chains that don't have Balancer v2.

We could default to `null` but then you'd need to handle the cases where there shouldn't be any cache.

kimpers (changes requested):

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
dekz: agree with kim, might spew errors on BSC

romainbutteaud: made `Mainnet` URL by default for now

dekz (approved):

kimpers (approved):
LGTM 👍

PR #269, opened 2021-07-01
fix: KyberDMM pick best pools, Polygon KyberDMM

romainbutteaud:
## Checklist:


dekz (approved):

packages/asset-swapper/contracts/src/KyberDmmSampler.sol
```diff
@@ -20,12 +20,25 @@
...
 interface IKyberDmmFactory {
 
-    function getPoolAtIndex(address token0, address token1, uint256 index)
+    function getPools(address token0, address token1)
         external
         view
-        returns (address);
+        returns (address[] memory _tokenPools);
+
+    function getPoolsLength(address token0, address token1)
```
dorothy-zbornak, line 38: Used?

packages/asset-swapper/contracts/src/KyberDmmSampler.sol
```diff
@@ -140,17 +153,25 @@ contract KyberDmmSampler
...
+                factory.getPools
                     {gas: KYBER_DMM_CALL_GAS}
-                    (path[i], path[i + 1], 0)
-                returns (address pool)
+                    (path[i], path[i + 1])
+                returns (address[] memory allPools)
             {
-                pools[i] = pool;
+                uint256 maxSupply = 0;
+                for (uint256 j = 0; j < allPools.length; j++) {
```
dorothy-zbornak, line 168: What happens if `allPools.length == 0`? We just have an invalid pool (null) at that index? Should we revert?

packages/asset-swapper/contracts/src/KyberDmmSampler.sol
romainbutteaud: removing it

packages/asset-swapper/contracts/src/KyberDmmSampler.sol
romainbutteaud: fixed it, added a revert if no pools are found

PR #272, opened 2021-07-05
fix: PLP fallback

dekz:
## Description

Fixes issue where PLP was considered a non-sturdy source, when a fallback was added PLP would be removed from the optimal path.

[Simbot](https://metabase.spaceship.0x.org/dashboard/186?run_id=jacob-polygon-plp-3&adjusted_amount_bought_win_tolerance__bps_=0.5)

## Checklist:

### DIFF
--- a/packages/asset-swapper/test/path_test.ts
+++ b/packages/asset-swapper/test/path_test.ts
@@ -20,7 +20,7 @@ const createFill = (
         sourcePathId: source,
     } as Fill);
 
-describe.only('Path', () => {
+describe('Path', () => {
     it('Adds a fallback', () => {
         const targetInput = new BigNumber(100);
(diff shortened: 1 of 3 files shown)


packages/asset-swapper/test/path_test.ts
```diff
@@ -0,0 +1,86 @@
...
+    ({
+        source,
+        input,
+        output,
+        adjustedOutput: output,
+        flags: BigInt(0),
+        sourcePathId: source,
+    } as Fill);
+
+describe.only('Path', () => {
```
mzhu25, line 23:
```suggestion
describe('Path', () => {
```

mzhu25 (approved):

PR #275, opened 2021-07-06
chore: emit a log when a quote is returned that is between 99-100% of request

phil-ociraptor:
## Description

For better visibility, emit a log when a request is returned between 99% and 100% of the requested amount.

## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/quote_requestor.ts
+++ b/packages/asset-swapper/src/utils/quote_requestor.ts
@@ -29,6 +29,7 @@ import { RfqMakerBlacklist } from './rfq_maker_blacklist';
 
 const MAKER_TIMEOUT_STREAK_LENGTH = 10;
 const MAKER_TIMEOUT_BLACKLIST_DURATION_MINUTES = 10;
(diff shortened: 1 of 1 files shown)


packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -563,6 +564,21 @@ export class QuoteRequestor {
                 this._warningLogger(order, 'Expiry too soon in RFQ-T firm quote, filtering out');
                 return false;
             } else {
+                const fillRatio = order.takerAmount.div(assetFillAmount);
```
PirosB3, line 567: why not `order.takerAmount < assetFillAmount`?

PirosB3 (approved):

packages/asset-swapper/src/utils/quote_requestor.ts
phil-ociraptor: discussed offline - mostly to find intent when a MM wants to send 100%, but for some reason sent slightly less

PR #284, opened 2021-07-15
feat: ACryptoS StableSwap

aandreakis:
## Checklist:

### DIFF
--- a/packages/asset-swapper/CHANGELOG.json
+++ b/packages/asset-swapper/CHANGELOG.json
@@ -1,5 +1,14 @@
 [
     {
+        "version": "16.23.0",
+        "changes": [
+            {
+                "note": "StableSwap",
+                "pr": 284
+            }
+        ]
+    },
+    {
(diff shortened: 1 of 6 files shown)


romainbutteaud:
I think we should name it `Acryptos` instead of `StableSwap` because StableSwap is the generic name of that stable  swap implementation (Curve / Shell / Saddle...) see: https://curve.fi/files/stableswap-paper.pdf

dorothy-zbornak (approved):
looks good! :+1: :eye: :lips: :eye: :+1:

aandreakis:
> I think we should name it `Acryptos` instead of `StableSwap` because StableSwap is the generic name of that stable swap implementation (Curve / Shell / Saddle...) see: https://curve.fi/files/stableswap-paper.pdf

Sounds good. Renaming to `ACryptoS`.

romainbutteaud (approved):
🚀

PR #285, opened 2021-07-15
FirebirdSwap

romainbutteaud:
## Checklist:


contracts/zero-ex/contracts/src/transformers/bridges/mixins/MixinFirebird.sol
```diff
@@ -0,0 +1,104 @@
...
+        FirebirdBridgeData memory data = abi.decode(bridgeData, (FirebirdBridgeData));
+        FirebirdRouter router = FirebirdRouter(data.router);
+
+        IERC20TokenV06 pool;
+        address _pool = data.pool;
+        {
+            assembly {pool := _pool}
+        }
+        IERC20TokenV06[] memory path = new IERC20TokenV06[](1);
+        path[0] = pool;
```
aandreakis, line 81: Happy for feedback if there is a better way to turn `FirebirdBridgeData.pool` of type `address` to a single element array of type `IERC20TokenV06`.

dekz: I believe this "pattern" was only done in UniswapV3 because Solidity probably freaked out about decoding into an array of `IERC20TokenV06`

In this case they're not ERC20Tokens but pool addresses, so I would make the following changes.

```
        (FirebirdRouter router, address[] memory pools) = abi.decode(bridgeData, (FirebirdRouter, address[]));
```

```
    function swapExactTokensForTokens(
        IERC20TokenV06 sellToken,
        IERC20TokenV06 buyToken,
        uint amountIn,
        uint amountOutMin,
        address[] calldata path,
        address to,
        uint deadline
    ) external returns (uint[] memory amounts);
```

Then in `orders.ts` the birdge data encoder  would be `address,address[]`

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
```diff
@@ -475,6 +483,10 @@ export const BRIDGE_ENCODERS: {
     ]),
     [ERC20BridgeSource.KyberDmm]: AbiEncoder.create('(address,address[],address[])'),
     [ERC20BridgeSource.Lido]: AbiEncoder.create('(address)'),
+    [ERC20BridgeSource.Firebird]: AbiEncoder.create([
+        { name: 'router', type: 'address' },
+        { name: 'pool', type: 'address' },
+    ]),
```
dekz, line 489: From my previous comment this would be 
```suggestion
    [ERC20BridgeSource.Firebird]: AbiEncoder.create([
        { name: 'router', type: 'address' },
        { name: 'pools', type: 'address[]' },
    ]),
```

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
```diff
@@ -318,6 +321,11 @@ export function createBridgeDataForBridgeOrder(order: OptimizedMarketBridgeOrder
             const lidoFillData = (order as OptimizedMarketBridgeOrder<LidoFillData>).fillData;
             bridgeData = encoder.encode([lidoFillData.stEthTokenAddress]);
             break;
+        case ERC20BridgeSource.Firebird:
+            const firebirdFillData = (order as OptimizedMarketBridgeOrder<FirebirdFillData>).fillData;
+            bridgeData = encoder.encode([firebirdFillData.router, firebirdFillData.pool]);
```
dekz, line 326: From my previous comment this would be
```suggestion
            bridgeData = encoder.encode([firebirdFillData.router, [firebirdFillData.pool]]);
```

contracts/zero-ex/contracts/src/transformers/bridges/mixins/MixinFirebird.sol
aandreakis: Sounds good. Handled via commit 25434c0
thx

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
aandreakis: handled in 25434c0

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
aandreakis: handled in 25434c0

aandreakis:
Failed test seems to be flaky and unrelated to this PR.
```
 |   1) Pools Caches for Balancer-based sampling
 |        BalancerPoolsCache
 |          fetches pools:
 | 
 |       Failed to find any pools for 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48 and 0x6b175474e89094c44da98b954eedeac495271d0f
 |       + expected - actual
```

Test passes locally and also passed in the previous commits. The latest commit just removed a whitespace and should not cause this :)

packages/asset-swapper/contracts/src/FirebirdSampler.sol
```diff
@@ -0,0 +1,111 @@
...
+    ///         amount.
+    function sampleBuysFromFirebird(
+        address pool,
+        address takerToken,
+        address makerToken,
+        uint256[] memory takerTokenAmounts
+    )
+        public
+        view
+        returns (uint256[] memory makerTokenAmounts)
```
dorothy-zbornak, line 87: swap "maker" and "taker" variables for buys.

dorothy-zbornak (approved):
one small nit but otherwise looks great!

packages/asset-swapper/contracts/src/FirebirdSampler.sol
aandreakis: good catch. thank you.
Fixed via d08b58e

PR #286, opened 2021-07-16
Fix AS refactor bugs

dorothy-zbornak:
## Description
Fix bugs and performance issues discovered by throwing the AS refactor under prod load.

- Use more accurate gas limits for all operations so calls can be more intelligently split across RPCs.
- Improve efficiency of UniswapV3 sampling solidity.
- Cache basic utility call results (isAddressContract, getTokenDecimals, etc).
- Cache prices for 10s (`getMedianSellRate()`).
- Put a 500ms time limit on `findOptimalPathAsync()` (probably doesn't help that much).
- `SassySampler` now assigns a "liquidity score" to each source and pair combo, ranging from 0 - 1.0. This is also the chance a source will be queried in a call to `getSellQuotes()` or `getTwoHopSellQuotes()`, etc (1.0 = 100%). Each time a source returns zero liquidity its score is halved. If it returns non-zero liquidity, its score is restored to 1.0. Additionally, every 10 seconds all liquidity scores are "healed" by 1%. So after ~16m a liquidity source with a score of 0 will be back to 1.0.
- Make source sampling calls return empty if the eth_call fails (instead of blowing up).
- Flush the ethCall queue if it gets "full", in addition to when the tick timeout expires. 

## Checklist:


packages/asset-swapper/src/network/chain.ts
```diff
@@ -283,8 +291,9 @@ function canBatchCallWith(ethCall: QueuedEthCall, batch: QueuedEthCall[]): boole
     if (batch.length === 0) {
         return true;
     }
+    const batchGas = batch.reduce((a, v) => a + (v.opts.gas || DEFAULT_CALL_GAS_LIMIT), 0);
     // TODO: have the sources provide realistic gas limits and split based on that.
-    if (batch.length >= 48) {
+    if ((ethCall.opts.gas || DEFAULT_CALL_GAS_LIMIT) + batchGas >= 384e6) {
```
dorothy-zbornak, line 296: Tried a bunch of values and this seemed to have the best performance characteristics. Might change in the future when we add more accurate gas limits to all the other sources.

packages/asset-swapper/src/network/network_utils.ts
```diff
@@ -7,6 +7,8 @@ import { ContractHelper, createContractWrapperAndHelper } from './utils';
 export class NetworkUtils {
     private readonly _networkUtilsContractHelper: ContractHelper<NetworkUtilsContract>;
     private readonly _networkUtilsContract: NetworkUtilsContract;
+    private readonly _tokenDecimalsCache: { [tokenAddress: string]: number } = {};
+    private readonly _isAddressContractCache: { [address: string]: boolean } = {};
```
dorothy-zbornak, line 11: TODO: Have these expire in case someone passes in an undeployed contract.

packages/asset-swapper/src/utils/quote_simulation.ts
```diff
@@ -172,7 +172,7 @@ export function fillQuoteOrders(
                     fo.totalOrderInput,
                     fo.totalOrderInputFee,
                 );
-                const filledOutput = subFill.output.times(filledInput.div(subFill.input));
+                const filledOutput = subFill.output.times(filledInput.div(BigNumber.max(1, subFill.input)));
```
dorothy-zbornak, line 175: Fixes the odd NaN error when quantities are very small. There's still another one out there I'm trying to track down.

mzhu25 (approved):
works for me, just one q

packages/asset-swapper/src/network/chain.ts
```diff
@@ -267,14 +273,16 @@ function resolveQueuedCall(c: QueuedEthCall, r: DispatchedCallResult): void {
...
+        c.reject(new Error(tryDecodeStringRevertErrorResult(r.resultData) || `EVM call reverted: ${r.resultData}`));
     }
 }
 
 function tryDecodeStringRevertErrorResult(rawResultData: Bytes): string | undefined {
     if (rawResultData.startsWith('0x08c379a0')) {
         const strLen: number = new BigNumber(hexUtils.slice(rawResultData, 4, 36)).toNumber();
-        return Buffer.from(hexUtils.slice(rawResultData, 68, 68 + strLen).slice(2), 'hex').toString();
+        return Buffer.from(hexUtils.slice(rawResultData, 68, 68 + strLen).slice(2), 'hex')
+            .filter(b => b === 0)
```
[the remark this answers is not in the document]
dorothy-zbornak: Sometimes revert strings get right-padded with 0s so they print like "ERROR\u0000\u0000". This (inelegantly) filters them out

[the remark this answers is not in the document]
dorothy-zbornak: OK I just looked at it and realized it's backwards ty :facepalm:

packages/asset-swapper/src/network/sassy_sampler.ts
```diff
@@ -205,6 +244,52 @@ export class SassySampler {
...
+    }
+
+    private _healLiquidityScores(healAmount: number): void {
+        for (const source in this._liquidityScores) {
+            const scores = this._liquidityScores[source as ERC20BridgeSource];
+            for (const pathId in scores) {
+                scores[pathId] = Math.min(scores[pathId] + healAmount, 1);
+            }
+        }
+    }
```
kimpers, line 288: Awesome! 🙏

kimpers (approved):
This is great! Extra props for the liquidity scoring and overall the red diff 😄

PR #293, opened 2021-07-23
feat: Aave aToken deposit/withdrawal [TKR-111]

kimpers:
## Description
Deposit token -> aToken and withdraw aToken -> token at 1:1 rate.

## Checklist:


packages/asset-swapper/src/utils/market_operation_utils/aave_reserves_cache.ts
```diff
@@ -0,0 +1,115 @@
+import { logUtils } from '@0x/utils';
+import { gql, request } from 'graphql-request';
+
+import { constants } from '../../constants';
+
+const RESERVES_GQL_QUERY = gql`
+    {
+        reserves(where: { isActive: true, isFrozen: false }, orderBy: totalLiquidity, orderDirection: desc) {
```
kimpers, line 8: **TODO** specify number of pools to fetch?

contracts/zero-ex/contracts/src/transformers/bridges/mixins/MixinAaveV2.sol
```diff
@@ -0,0 +1,93 @@
...
+
+        if (address(buyToken) == aToken) {
+            lendingPool.deposit(address(sellToken), sellAmount, address(this), 0);
+            // 1:1 mapping token -> aToken and have the same number of decimals as the underlying token
+            return sellAmount;
+        } else if (address(sellToken) == aToken) {
+            return lendingPool.withdraw(address(buyToken), sellAmount, address(this));
+        }
+
+        revert("MixinLido/UNSUPPORTED_TOKEN_PAIR");
```
romainbutteaud, line 91: small typo

romainbutteaud:
awesome job! 🚀

romainbutteaud (approved):

packages/asset-swapper/src/utils/market_operation_utils/aave_reserves_cache.ts
```diff
@@ -0,0 +1,115 @@
...
+                id
+                underlyingAssetAddress
+            }
+            pool {
+                id
+                lendingPool
+            }
+            decimals
+            symbol
+        }
```
dorothy-zbornak, line 25: We don't actually use a lot of these? I think you can leave out fields even if they're used in the `where` and the `orderBy` clauses. `id`s are also optional afaik

packages/asset-swapper/src/utils/market_operation_utils/aave_reserves_cache.ts
```diff
@@ -0,0 +1,115 @@
...
+        reserves(where: { isActive: true, isFrozen: false }, orderBy: totalLiquidity, orderDirection: desc) {
+            id
+            name
+            isActive
+            isFrozen
+            underlyingAsset
+            totalLiquidity
+            aToken {
+                id
+                underlyingAssetAddress
```
dorothy-zbornak, line 17: Is there a difference between this and the outer `underlyingAsset`?

packages/asset-swapper/contracts/src/AaveV2Sampler.sol
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    function sampleSellsFromAaveV2(
+        AaveInfo memory aaveInfo,
+        address takerToken,
+        address makerToken,
+        uint256[] memory takerTokenAmounts
+    )
+        public
+        pure
```
dorothy-zbornak, line 40: Ideally we would just do all this in AS and skip the eth_call but I can tend to that.

dorothy-zbornak (approved):
:+1: :eye: :lips: :eye: :+1:

packages/asset-swapper/src/utils/market_operation_utils/aave_reserves_cache.ts
kimpers: IIRC 100 is the default limit, which is more than enough but it's probably better to be explicit :)

packages/asset-swapper/src/utils/market_operation_utils/aave_reserves_cache.ts
kimpers: Nope 🙈

packages/asset-swapper/src/utils/market_operation_utils/aave_reserves_cache.ts
kimpers: Yeah, I used some of this extra info for debugging during implementation, and grabbing id's everywhere is an old habit, because caching in heavier GQL clients like Apollo break if you don't fetch ids :feelsgood: 

Will clean it up 👍

packages/asset-swapper/contracts/src/AaveV2Sampler.sol
kimpers: Good point! This file entirely unnecessary, move this logic into AS instead 🙏

contracts/zero-ex/contracts/src/transformers/bridges/mixins/MixinAaveV2.sol
kimpers: Thanks! 🙏  Copy-pasta is hard 🙈

packages/asset-swapper/contracts/src/AaveV2Sampler.sol
kimpers: Moved to local sampling in f26497f815f306c211685d66bf348ef164b529df 👍

packages/asset-swapper/src/utils/market_operation_utils/aave_reserves_cache.ts
kimpers: Removed most of the unused fields from the data fetching and caching. Prefer to keep the ids tho as that is good GQL hygiene, even if our GQL client doesn't do any fancy query deduping/batching or response catching 😄

PR #295, opened 2021-07-28
feat: initial integration of new router

kimpers:
## Description

Initial integration of the new router with RFQ/Limit order support and basic VIP handling

## Checklist:

### DIFF
--- a/packages/asset-swapper/package.json
+++ b/packages/asset-swapper/package.json
@@ -67,6 +67,7 @@
         "@0x/dev-utils": "^4.2.9",
         "@0x/json-schemas": "^6.3.0",
         "@0x/protocol-utils": "^1.9.1",
+        "@0x/neon-router": "^0.1.3",
         "@0x/quote-server": "^6.0.6",
         "@0x/types": "^3.3.4",
         "@0x/typescript-typings": "^5.2.1",
@@ -89,8 +90,7 @@
         "graphql": "^15.4.0",
         "graphql-request": "^3.4.0",
         "heartbeats": "^5.0.1",
-        "lodash": "^4.17.11",
-        "neon-router": "link:~/Development/Router/neon-router"
+        "lodash": "^4.17.11"
     },
     "devDependencies": {
         "@0x/abi-gen": "^5.6.2",
--- a/packages/asset-swapper/package.json
+++ b/packages/asset-swapper/package.json
@@ -89,7 +89,8 @@
         "graphql": "^15.4.0",
         "graphql-request": "^3.4.0",
         "heartbeats": "^5.0.1",
-        "lodash": "^4.17.11"
+        "lodash": "^4.17.11",
+        "neon-router": "link:~/Development/Router/neon-router"
     },
     "devDependencies": {
         "@0x/abi-gen": "^5.6.2",
--- a/packages/asset-swapper/src/utils/market_operation_utils/fills.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/fills.ts
@@ -71,7 +71,7 @@ function hasLiquidity(fills: Fill[]): boolean {
     return true;
 }
 
-function nativeOrdersToFills(
+export function nativeOrdersToFills(
     side: MarketOperation,
     orders: NativeOrderWithFillableAmounts[],
     targetInput: BigNumber = POSITIVE_INF,
--- a/packages/asset-swapper/src/utils/market_operation_utils/fills.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/fills.ts
@@ -71,6 +71,24 @@ function hasLiquidity(fills: Fill[]): boolean {
     return true;
 }
 
+export function ethToOutputAmount({
+    input,
+    output,
+    ethAmount,
+    inputAmountPerEth,
+    outputAmountPerEth,
+}: {
+    input: BigNumber;
+    output: BigNumber;
+    inputAmountPerEth: BigNumber;
+    outputAmountPerEth: BigNumber;
+    ethAmount: BigNumber | number;
+}): BigNumber {
+    return !outputAmountPerEth.isZero()
+        ? outputAmountPerEth.times(ethAmount)
+        : inputAmountPerEth.times(ethAmount).times(output.dividedToIntegerBy(input));
+}
+
 export function nativeOrdersToFills(
     side: MarketOperation,
     orders: NativeOrderWithFillableAmounts[],
@@ -89,9 +107,13 @@ export function nativeOrdersToFills(
         const input = side === MarketOperation.Sell ? takerAmount : makerAmount;
         const output = side === MarketOperation.Sell ? makerAmount : takerAmount;
         const fee = fees[ERC20BridgeSource.Native] === undefined ? 0 : fees[ERC20BridgeSource.Native]!(o);
-        const outputPenalty = !outputAmountPerEth.isZero()
-            ? outputAmountPerEth.times(fee)
-            : inputAmountPerEth.times(fee).times(output.dividedToIntegerBy(input));
+        const outputPenalty = ethToOutputAmount({
+            input,
+            output,
+            inputAmountPerEth,
+            outputAmountPerEth,
+            ethAmount: fee,
+        });
         // targetInput can be less than the order size
         // whilst the penalty is constant, it affects the adjusted output
         // only up until the target has been exhausted.
@@ -156,9 +178,13 @@ export function dexSamplesToFills(
         let penalty = ZERO_AMOUNT;
         if (i === 0) {
             // Only the first fill in a DEX path incurs a penalty.
-            penalty = !outputAmountPerEth.isZero()
-                ? outputAmountPerEth.times(fee)
-                : inputAmountPerEth.times(fee).times(output.dividedToIntegerBy(input));
+            penalty = ethToOutputAmount({
+                input,
+                output,
+                inputAmountPerEth,
+                outputAmountPerEth,
+                ethAmount: fee,
+            });
         }
         const adjustedOutput = side === MarketOperation.Sell ? output.minus(penalty) : output.plus(penalty);
 
--- a/packages/asset-swapper/src/utils/market_operation_utils/fills.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/fills.ts
@@ -132,7 +132,7 @@ function nativeOrdersToFills(
     return fills;
 }
 
-function dexSamplesToFills(
+export function dexSamplesToFills(
     side: MarketOperation,
     samples: DexSample[],
     outputAmountPerEth: BigNumber,
--- a/packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
@@ -45,7 +45,7 @@ function calculateOuputFee(
 ): BigNumber {
     if (isDexSample(sampleOrNativeOrder)) {
         const { source, fillData } = sampleOrNativeOrder;
-        const fee = fees[source] === undefined ? 0 : fees[source]!(fillData) || 0;
+        const fee = fees[source]?.(fillData) || 0;
         const outputFee = !outputAmountPerEth.isZero()
             ? outputAmountPerEth.times(fee)
             : inputAmountPerEth
@@ -55,8 +55,7 @@ function calculateOuputFee(
         return outputFee;
     } else {
         const { input, output } = nativeOrderToNormalizedAmounts(side, sampleOrNativeOrder);
-        const fee =
-            fees[ERC20BridgeSource.Native] === undefined ? 0 : fees[ERC20BridgeSource.Native]!(sampleOrNativeOrder);
+        const fee = fees[ERC20BridgeSource.Native]?.(sampleOrNativeOrder) || 0;
         const outputFee = !outputAmountPerEth.isZero()
             ? outputAmountPerEth.times(fee)
             : inputAmountPerEth.times(fee).times(output.dividedToIntegerBy(input));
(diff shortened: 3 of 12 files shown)


packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -9,17 +11,161 @@ import { ERC20BridgeSource, Fill } from './types';
...
+        } else {
+            sampleIdLookup[key] = ++sampleIdCounter;
+            return sampleIdLookup[key];
+        }
+    };
+
+    const adjustedParsedFills = allFills.map(fills => {
+        const _adjustedFills: FillWithOutputFee[] = [];
+        // NOTE: Output fee is negative for sells and positive for buys
+        const outputFee = fills[0].adjustedOutput.minus(fills[0].output);
```
kimpers, line 47: **TODO:** Need to fix this now that we're not using swap revert anymore?

packages/asset-swapper/package.json
```diff
@@ -89,7 +89,8 @@
         "graphql": "^15.4.0",
         "graphql-request": "^3.4.0",
         "heartbeats": "^5.0.1",
-        "lodash": "^4.17.11"
+        "lodash": "^4.17.11",
+        "neon-router": "link:~/Development/Router/neon-router"
```
kimpers, line 93: **TODO:** Publish the router and update this

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -1,20 +1,340 @@
...
+        const outputs = [
+            0,
+            normalizedOrderOutput
+                .dividedBy(2)
+                .integerValue()
+                .toNumber(),
+            normalizedOrderOutput.integerValue().toNumber(),
+        ];
+        // NOTE: same fee no matter if full or partial fill
+        const fee = calculateOuputFee(side, nativeOrder, opts.outputAmountPerEth, opts.inputAmountPerEth, fees)
```
dorothy-zbornak, line 155: Is the router OK with cases where fee > sample?

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -1,20 +1,340 @@
...
+            inputs,
+            outputs,
+            outputFees,
+        };
+
+        samplesAndNativeOrdersWithResults.push([nativeOrder]);
+        serializedPaths.push(serializedPath);
+    }
+
+    const rustArgs: OptimizerCapture = {
```
dorothy-zbornak, line 174: Oh I did not expect you guys to keep using my data dump hack input types and names.

dorothy-zbornak: Is that why we're still passing in path IDs?

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -1,20 +1,340 @@
...
+    const totalRoutedAmount = BigNumber.sum(...allSourcesRustRoute);
+
+    const scale = input.dividedBy(totalRoutedAmount);
+    for (const [routeInput, routeSamplesAndNativeOrders] of routesAndSamples) {
+        if (!routeInput || !routeSamplesAndNativeOrders) {
+            continue;
+        }
+        // TODO(kimpers): [TKR-241] amounts are sometimes clipped in the router due to precisions loss for number/f64
+        // we can work around it by scaling it and rounding up. However now we end up with a total amount of a couple base units too much
+        const rustInput = new BigNumber(routeInput).multipliedBy(scale).integerValue(BigNumber.ROUND_CEIL);
```
dorothy-zbornak, line 202: We might want to `min()` this with the sample's `input` as well.

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -1,20 +1,340 @@
...
+        const fee =
+            fees[ERC20BridgeSource.Native] === undefined ? 0 : fees[ERC20BridgeSource.Native]!(sampleOrNativeOrder);
+        const outputFee = !outputAmountPerEth.isZero()
+            ? outputAmountPerEth.times(fee)
+            : inputAmountPerEth.times(fee).times(output.dividedToIntegerBy(input));
+        return outputFee;
+    }
+}
+
+function createPathFromSamples(
```
dorothy-zbornak, line 67: This function feels like it has a deceptively simple name considering all it does.

dorothy-zbornak (commented):
some thoughts, still looking through it :eyeglasses:

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -1,20 +1,340 @@
...
+    fees: FeeSchedule,
+): BigNumber {
+    if (isDexSample(sampleOrNativeOrder)) {
+        const { source, fillData } = sampleOrNativeOrder;
+        const fee = fees[source] === undefined ? 0 : fees[source]!(fillData) || 0;
+        const outputFee = !outputAmountPerEth.isZero()
+            ? outputAmountPerEth.times(fee)
+            : inputAmountPerEth
+                  .times(fee)
+                  .times(sampleOrNativeOrder.output.dividedToIntegerBy(sampleOrNativeOrder.input));
```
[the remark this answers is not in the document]
kimpers: Agreed, too much copy-pasta 😅

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
```diff
@@ -71,6 +71,24 @@ function hasLiquidity(fills: Fill[]): boolean {
     return true;
 }
 
+export function ethToOutputAmount({
```
kimpers, line 74: Not sure if this is the best place to put this logic but I didn't want to create yet another file for this general util 🙈

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
dorothy-zbornak: yeah that's fine

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -1,20 +1,358 @@
...
+            continue;
+        }
+
+        // NOTE: For DexSamples only
+        let fill = createFill(current);
+        const routeSamples = routeSamplesAndNativeOrders as Array<DexSample<FillData>>;
+        // Descend to approach a closer fill for fillData which may not be consistent
+        // throughout the path (UniswapV3) and for a closer guesstimate at
+        // gas used
+        for (let k = routeSamples.length - 1; k >= 0; k--) {
```
dorothy-zbornak, line 243: I feel like so much of this would break unexpectedly or silently if there was a bug in the router because there's a lot of assumptions going on with the consistency of the router's outputs. Maybe we could pepper in asserts?

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -1,20 +1,358 @@
...
+                        .minus(rightPrice)
+                        .dividedBy(left.input.minus(right.input))
+                        .times(rustInput.minus(right.input))
+                        .plus(rightPrice);
+                    const output = scaledPrice
+                        .times(rustInput)
+                        .decimalPlaces(0, side === MarketOperation.Sell ? BigNumber.ROUND_FLOOR : BigNumber.ROUND_CEIL);
+                    fill = createFill({
+                        ...right, // default to the greater (for gas used)
+                        input: rustInput,
```
dorothy-zbornak, line 265: shouldn't we interpolate this as well?

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -1,20 +1,358 @@
...
+                    const leftPrice = left.output.dividedBy(left.input);
+                    const rightPrice = right.output.dividedBy(right.input);
+                    const scaledPrice = leftPrice
+                        .minus(rightPrice)
+                        .dividedBy(left.input.minus(right.input))
+                        .times(rustInput.minus(right.input))
+                        .plus(rightPrice);
+                    const output = scaledPrice
+                        .times(rustInput)
+                        .decimalPlaces(0, side === MarketOperation.Sell ? BigNumber.ROUND_FLOOR : BigNumber.ROUND_CEIL);
```
dorothy-zbornak, line 262: I am not entirely sure what we're doing here. Can we get a comment on the formula?

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -1,20 +1,358 @@
...
+                const left = routeSamples[k];
+                const right = routeSamples[k + 1];
+                if (left && right) {
+                    const leftPrice = left.output.dividedBy(left.input);
+                    const rightPrice = right.output.dividedBy(right.input);
+                    const scaledPrice = leftPrice
+                        .minus(rightPrice)
+                        .dividedBy(left.input.minus(right.input))
+                        .times(rustInput.minus(right.input))
+                        .plus(rightPrice);
```
dorothy-zbornak, line 259: Should this be
```suggestion
                        .plus(leftPrice);
```

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
dorothy-zbornak: Is this all equivalent to...?
```ts
                    const stride = right.input.minus(left.input);
                    const rateBetween = right.output.minus(left.output).dividedBy(stride);
                    const output = rustInput
                        .minus(left.input)
                        .dividedBy(stride)
                        .times(rateBetween)
                        .plus(left.input)
                        .decimalPlaces(...);
```

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -1,20 +1,358 @@
...
+                    });
+                } else {
+                    fill = createFill(left || right);
+                }
+                break;
+            }
+        }
+
+        // HACK: Handle the case where the router can under quote the input
+        // Set the first fill just a tad higher
```
dorothy-zbornak, line 276: Doesn't it make more sense to inflate the last fill? But also clipping it to the max input for that fill? And should it be spread over all the sources? Or just the last? I guess there is no canonical ordering of fills coming from the router, right?

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
kimpers: I could be wrong but I don't think that is exactly equivalent 😅 

I don't know why we use price here to interpolate (shamelessly stole it from Jacobs PoC that used the midpoint price), I have cleaned this up and refactored to do interpolate on the input and output amounts instead with some useful comments. Lmkwyt!

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
kimpers: The ordering that the router returns is the same as the input order so there's no significance to that. Is there any advantage to doing it on the last fill? (This is another snippet that I blatantly stole), it seems easier to just do it on the first? 🤔 

It would be better to spread this over all the sources but it's probably not worth the effort. The router already tries to do that before returning the results but the `f64/number` precision loss makes it sometimes slightly short anyway.

Once we refactor the router to use a higher precision representation of amounts we should be able to remove this hack entirely!

kimpers: On second thought I wonder if we can just remove this code entirely, I already added [scaling on the input](https://github.com/0xProject/protocol/pull/295/files#diff-899962c6423923a5840eea991afd577aa723139f14440279faffba0b8726ab36R237-R242) before we get here to combat the precision loss. I think we're double adjusting right now 🤔

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
dorothy-zbornak: I just want to be sure we're interpolating using the price *between* the two samples, and not the difference between each sample's price (from 0).

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
dorothy-zbornak: Yeah, I'm crazy

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
dorothy-zbornak: OK I forgot we're creating fills and not samples (which are cumulative). But it still feels like if we're going to be inflating a fill it would be safer using the worst rate (last fill) vs best (first fill).

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -1,20 +1,377 @@
...
+const interpolateOutputFromSamples = (
+    left: { input: BigNumber; output: BigNumber },
+    right: { input: BigNumber; output: BigNumber },
+    targetInput: BigNumber,
+): BigNumber =>
+    left.output.plus(
+        right.output
+            .minus(left.output)
+            .dividedBy(right.input.minus(left.input))
+            .times(targetInput.minus(left.input)),
```
dorothy-zbornak, line 84: :+1:

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
kimpers: Yeah right, but keep in mind that these amounts stem from rounding/precision errors so they will be in the order of magnitude of a couple thousand wei at the high end 🐛

dorothy-zbornak (approved):
Do itt

PR #296, opened 2021-07-29
AS: Fix fill amount rounding error when converting to orders

dorothy-zbornak:
## Description
Paths don't often produce integer `output` values. Previously we were using this value directly to populate the `maxTakerFillAmount` in an FQT order. But for buys this can sometimes result in selling too little and causing an underbought error. This PR rounds up or down the path values to prevent this issue.

## Checklist:


dekz:
seems fine to me 😂

dekz (approved):

PR #300, opened 2021-08-05
fix: Adds "MetricsProxy" that allows us to proxy Prometheus metrics across…

PirosB3:
Adds a MetricsProxy interface object that serves as a proxy between AssetSwapper metrics and 0x API metrics. The terminology used in the proxy is pretty opaque, however all of our metrics target Prometheus 100%


packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -367,13 +367,14 @@ export class SwapQuoter {
         // pass the QuoteRequestor on if rfqt enabled
         if (calcOpts.rfqt !== undefined) {
             calcOpts.rfqt.quoteRequestor = new QuoteRequestor(
-                rfqtOptions ? rfqtOptions.makerAssetOfferings || {} : {},
```
PirosB3, line 370: Minor refactoring, can omit if you prefer

phil-ociraptor (commented):
I like this approach!

PR #303, opened 2021-08-06
Update TransformERC20Feature

mzhu25:
## Checklist:

### DIFF
--- a/contracts/zero-ex/contracts/src/features/TransformERC20Feature.sol
+++ b/contracts/zero-ex/contracts/src/features/TransformERC20Feature.sol
@@ -22,6 +22,7 @@ pragma experimental ABIEncoderV2;
 
 import "@0x/contracts-erc20/contracts/src/v06/IERC20TokenV06.sol";
 import "@0x/contracts-erc20/contracts/src/v06/LibERC20TokenV06.sol";
+import "@0x/contracts-utils/contracts/src/v06/LibBytesV06.sol";
 import "@0x/contracts-utils/contracts/src/v06/errors/LibRichErrorsV06.sol";
 import "@0x/contracts-utils/contracts/src/v06/LibSafeMathV06.sol";
 import "../errors/LibTransformERC20RichErrors.sol";
@@ -60,8 +61,6 @@ contract TransformERC20Feature is
     /// @dev Version of this feature.
     uint256 public immutable override FEATURE_VERSION = _encodeVersion(1, 4, 0);
 
-    constructor() public {}
-
     /// @dev Initialize and register this feature.
     ///      Should be delegatecalled by `Migrate.migrate()`.
     /// @param transformerDeployer The trusted deployer for transformers.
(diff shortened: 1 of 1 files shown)


contracts/zero-ex/contracts/src/features/TransformERC20Feature.sol
```diff
@@ -244,6 +281,12 @@ contract TransformERC20Feature is
                     args.recipient
                 );
             }
+            // Transfer output tokens from wallet to recipient
+            outputTokenAmount = _executeOutputTokenTransfer(
+                args.outputToken, 
+                state.wallet, 
+                args.recipient
+            );
         }
```
dorothy-zbornak, line 290: wonder why this stuff is in its own scope

contracts/zero-ex/contracts/src/features/TransformERC20Feature.sol
```diff
@@ -244,6 +281,12 @@ contract TransformERC20Feature is
                     args.recipient
                 );
             }
+            // Transfer output tokens from wallet to recipient
+            outputTokenAmount = _executeOutputTokenTransfer(
```
dorothy-zbornak, line 285: So this would replace `PayTakerTransformer`? That transformer transfers more than the output token currently, like leftover taker token (buys) or unspent protocol fees so we would still need it. Maybe we could pass in additional tokens to transfer out.

Also does the fact that we set `minOutputTokenAmount = 0` [in multiplex](https://github.com/0xProject/protocol/blob/feat/otc/spot-check/contracts/zero-ex/contracts/src/features/multiplex/MultiplexTransformERC20.sol#L47) mean none of this matters in the context of multiplex?

dorothy-zbornak: A big part of me really wants to just add `outputToken, minBuyAmount` to the transformer context and do the check in (the last) FQT and not be too worried about the return value of `transformERC20()` being gamable because it wouldn't be worth tricking it anymore with the check in FQT.

mzhu25: it wouldn't necessarily replace `PayTakerTransformer` (though I had the same thought). It would mean that `PayTakerTransformer` would no longer handle the `outputToken` in particular. It matters to Multiplex insofar as Multiplex can now trust the `outputTokenAmount` returned by `_tranformERC20`

mzhu25: I'm not really seeing the benefit of doing that in FQT. I started doing that, but calculating the desired `minBuyAmount` in `MultiplexTransformERC20` is kinda annoying

dorothy-zbornak: Yeah I guess we would always want `minBuyAmount: 0` for a transformERC20 in a batch fills which would disable any check in the FQT. :thinking:

PR #305, opened 2021-08-10
fix: Add ethers dependency to `0x/contracts-erc20`

rhinodavid:
## Description

Here's the problem: `0x/abi-gen`, which is used to build the contracts in `0x/contracts-erc20`, produces code specific to `ethers` v4. [An example from the TS template:](https://github.com/0xProject/tools/blob/5d17d8f4581e6a9f066d51530910e9b708cde661/abi-gen/templates/TypeScript/contract.handlebars#L151)

```
const iface = new ethers.utils.Interface(abi);
const deployInfo = iface.deployFunction;
const txData = deployInfo.encode(bytecode, [{{> params inputs=ctor.inputs}}]);
```

`deployFunction` is present in [ethers v4](https://docs.ethers.io/v4/api-advanced.html#id3), but v5 breaks this.

`ethers` is currently not a dependency of `@0x/contracts-erc20`. Therefore, if a package depends on `@0x/contracts-erc20`, to resolve `ethers` node will look in `@0x/contracts-erc20`'s `node_modules` and not find it, and node will [start traversing up the file tree to find `ethers`](https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders). If the package depending on `@0x/contracts-erc20` also depends on `ethers` v5, then that version of `ethers` will be used to process the generated contract abi. Since `deployFunction` is `undefined` in v5, the program will fail when it tries to access `deployInfo.encode`.

Add `ethers` v4 as an explicit dependency of `@0x/contracts-erc20` so that contracts will always require v4.
## Checklist:

### DIFF
--- a/contracts/erc20/package.json
+++ b/contracts/erc20/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@0x/contracts-erc20",
-    "version": "3.3.14",
+    "version": "3.3.15",
     "engines": {
         "node": ">=6.12"
     },
@@ -82,7 +82,8 @@
         "typescript": "4.2.2"
     },
     "dependencies": {
-        "@0x/base-contract": "^6.4.0"
+        "@0x/base-contract": "^6.4.0",
+        "ethers": "~4.0.4"
     },
     "publishConfig": {
         "access": "public"
--- a/contracts/erc20/package.json
+++ b/contracts/erc20/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@0x/contracts-erc20",
-    "version": "3.3.15",
+    "version": "3.3.14",
     "engines": {
         "node": ">=6.12"
     },
--- a/contracts/erc20/CHANGELOG.json
+++ b/contracts/erc20/CHANGELOG.json
@@ -1,5 +1,14 @@
 [
     {
+        "version": "3.3.15",
+        "changes": [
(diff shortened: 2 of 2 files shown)


contracts/erc20/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@0x/contracts-erc20",
-    "version": "3.3.14",
+    "version": "3.3.15",
```
mzhu25, line 3: no need to manually update this, instead you should add an entry to `contracts/erc20/CHANGELOG.json` like:
```
    {
        "version": "3.3.15",
        "changes": [
            {
                "note": "Add ethers to dependencies",
                "pr": 305
            }
        ]
    },
```
(note that you don't need to put the `timestamp`)
The GitHub `publish` action will handle the rest 🙂

mzhu25 (commented):
thanks for fixing this!

contracts/erc20/package.json
rhinodavid: yall are fancy. updated.

dorothy-zbornak (approved):

dorothy-zbornak:
Technically we should add `ethers` as a dep to all packages that export abi-gen wrappers. But I can take care of that later since we'll probably upgrade to ethers 5 for 1559.

PR #307, opened 2021-08-14
fix: fallback path include unused fills on a used source

dekz:
## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/market_operation_utils/index.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/index.ts
@@ -38,7 +38,7 @@ import {
 import { createFills } from './fills';
 import { getBestTwoHopQuote } from './multihop_utils';
 import { createOrdersFromTwoHopSample } from './orders';
-import { PathPenaltyOpts } from './path';
+import { Path, PathPenaltyOpts } from './path';
 import { fillsToSortedPaths, findOptimalPathAsync } from './path_optimizer';
 import { DexOrderSampler, getSampleAmounts } from './sampler';
 import { SourceFilters } from './source_filters';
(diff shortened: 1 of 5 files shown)


packages/asset-swapper/test/market_operation_utils_test.ts
```diff
@@ -1166,28 +1166,6 @@ describe('MarketOperationUtils tests', () => {
                 expect(orderSources.sort()).to.deep.eq(expectedSources.sort());
             });
 
-            it('fallback orders use different sources', async () => {
```
dekz, line 1169: We don't use a different source. We can pad more to the currently used source or add a different source. Tests in `path_test.ts`

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -727,6 +705,44 @@ export class MarketOperationUtils {
             }),
         );
     }
+
+    // tslint:disable-next-line: prefer-function-over-method
```
kimpers, line 709: I vote we just get rid of this annoying linting rule

kimpers (commented):
Post merge approval ✅  😂

packages/asset-swapper/src/utils/market_operation_utils/path.ts
```diff
@@ -99,14 +99,18 @@ export class Path {
...
+
         this.fills = [
             // Append all of the native fills first
             ...nativeFills.filter(f => f !== lastNativeFillIfExists),
             // Add the other fills that are not native in the optimal path
             ...otherFills,
-            // Add the fallbacks to the end that aren't already included
-            ...fallback.fills.filter(f => !otherSourcePathIds.includes(f.sourcePathId)),
+            // Add the fills to the end that aren't already included
+            ...fallback.fills.filter(f => !otherFillIds.includes(fillToFillId(f))),
```
dorothy-zbornak, line 113: Can this cause non-contiguous sources to occur? Like if the current path is `[Native, Uniswap(0)]` and we create a fallback of `[Balancer, Uniswap(0), Uniswap(1)]`, do we wind up with `[Native, Uniswap(0), Balancer, Uniswap(1)]`?

PR #308, opened 2021-08-17
fix: added prometheus changes on expiration

PirosB3:
## Checklist:


packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -177,6 +177,17 @@ export class QuoteRequestor {
         }
     }
 
+    public static getExpirationContext(
```
rhinodavid, line 180: I think my slight preference would be to just make this `durationUntilExpirationS (expirationTimeS): /* seconds until expiration */ BigNumber`.. then when you use it:
```
if (durationUntilExpirationS(expirationTime) < this._expiryBuffer) { ... }
```
I think this is slightly better because (1) it's not as complex and (2) the function is pure at that point. But it's only a very slight preference; otherwise LGTM.

rhinodavid (approved):

packages/asset-swapper/src/utils/quote_requestor.ts
PirosB3: This is great feedback! will adjust

rhinodavid (approved):

PR #317, opened 2021-09-03
fix: Exclusive API keys for select integrations

PirosB3:
## Description

We want to create API keys that can exclusively ping a select group of market makers. This provides a better overall user experience when integrating with 0x API and Matcha.


packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -367,6 +367,43 @@ export class QuoteRequestor {
...
+                  (mm: string): TypedMakerUrl => {
+                      return { pairType: RfqPairType.Alt, url: mm };
+                  },
+              )
+            : [];
+
+        let typedMakerUrls = standardUrls.concat(altUrls);
+
+        // If there is a whitelist, filter out maker URLs so that only
+        // they only match the whitelist.
```
phil-ociraptor, line 399: nit: typo

```suggestion
        // If there is a whitelist, only allow approved maker URLs 
```

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -367,6 +367,43 @@ export class QuoteRequestor {
...
+    /**
+     * Gets both standard RFQ makers and "alternative" RFQ makers and combines them together
+     * in a single configuration map. If an integration key whitelist is present, it will be used
+     * to filter a specific makers.
+     *
+     * @param options the RfqmRequestOptions passed in
+     * @param assetOfferings the RFQM or RFQT maker offerings
+     * @returns a list of TypedMakerUrl instances
+     */
+    public static getTypedMakerUrls(
```
phil-ociraptor, line 379: hmm, this function is doing two things - I overlooked the second part when reading it the first time:

1. Combines regular and alt asset offerings into one typed array
2. Filters out maker urls unless they are approved

I would either rename it to something like `getTypedMakerUrlsAndWhitelist` or break into two functions: `getTypedMakerUrls` and `selectWhitelistedMakerUrls`

PirosB3:
Thank you for the feedback @phil-ociraptor ! I addressed

phil-ociraptor (approved):

PR #323, opened 2021-09-14
fix: Remove API key whitelist field

PirosB3:
Tuck the integrator key whitelist into it's own type

### DIFF
--- a/packages/asset-swapper/src/utils/quote_requestor.ts
+++ b/packages/asset-swapper/src/utils/quote_requestor.ts
@@ -65,8 +65,8 @@ export interface MetricsProxy {
 
     /**
      * Logs the outcome of a network (HTTP) interaction with a market maker.
-     * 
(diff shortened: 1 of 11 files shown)


packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -214,7 +214,7 @@ export class QuoteRequestor {
      * @returns a list of TypedMakerUrl instances
      */
     public static getTypedMakerUrlsAndWhitelist(
```
rhinodavid, line 216: what does "typed" mean here?

rhinodavid (approved):

packages/asset-swapper/src/utils/quote_requestor.ts
PirosB3: Discussed in "person": no need to adjust things here

PR #335, opened 2021-09-29
fix: removing Clipper custom integration (to add it later as a real PLP)

romainbutteaud:
## Description

- Removing Clipper, will then add the new contract in our infra plp registry   

## Checklist:


contracts/zero-ex/contracts/src/transformers/bridges/BridgeProtocols.sol
```diff
@@ -49,5 +49,4 @@ library BridgeProtocols {
     uint128 internal constant KYBERDMM    = 19;
     uint128 internal constant CURVEV2     = 20;
     uint128 internal constant LIDO        = 21;
-    uint128 internal constant CLIPPER     = 22;
```
dorothy-zbornak, line 52: Let's just keep this but comment that it's not used.

packages/protocol-utils/src/transformer_utils.ts
```diff
@@ -131,7 +131,6 @@ export enum BridgeProtocol {
     KyberDmm,
     CurveV2,
     Lido,
-    Clipper,
```
dorothy-zbornak, line 134: Same.

dorothy-zbornak (approved):

PR #349, opened 2021-10-21
[WIP] Create Polygon Univ2 VIP feature

cece-z:
## Description
Create Polygon Univ2 VIP feature

## Checklist:


contracts/zero-ex/contracts/src/features/UniswapV2Feature.sol
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    function sellToUniswapV2(
+        IERC20TokenV06[] calldata tokens,
+        uint256 sellAmount,
+        uint256 minBuyAmount,
+        ProtocolFork fork
+    )
+        external
+        payable
```
dorothy-zbornak, line 64: Payable implies that this function can also sell ETH, right? But I don't see that case being handled.

dorothy-zbornak: Following the assembly version of this feature:
- if we detect `0xeee...` for the first token, it's an ETH sell and we need to wrap ETH->WETH and transfer from ourselves.
- If we detect `0xeee...` for the last token, it's an ETH buy and we need to make ourselves the final recipient in the pair swap, unwrap the bought WETH->ETH, then transfer that to `msg.sender`.

contracts/zero-ex/contracts/src/features/UniswapV2ForkFeature.sol
```diff
@@ -0,0 +1,205 @@
...
+        uint[] memory amounts = _getAmountsOut(factory, sellAmount, path);
+        buyAmount = amounts[amounts.length - 1];
+        require(buyAmount >= minBuyAmount, 'UniswapV2ForkFeature/INSUFFICIENT_OUTPUT_AMOUNT');
+        _transferERC20TokensFrom(
+            path[0],
+            msg.sender,
+            _pairFor(factory, path[0], path[1]),
+            amounts[0]
+        );
+        _swap(amounts, path, path[path.length - 1], factory);
```
dorothy-zbornak, line 88: What if `_swap()` computed `amounts` on the fly? iteratively? I think we can avoid having to do a `_pairFor()` 2+ times per pool if we do it that way.

dorothy-zbornak: Also, why is `_to` being set to a token?

Remeber we want the first recipient to be:
1. The next pool if `path.length > 2`
2. The taker if `!unwrapWeth`
3. Otherwise `this`

contracts/zero-ex/contracts/src/features/UniswapV2ForkFeature.sol
```diff
@@ -0,0 +1,205 @@
...
+
+        if (unwrapWeth) {
+            IEtherTokenV06(WETH_ADDRESS).withdraw(buyAmount);
```
dorothy-zbornak, line 91: This gives `this` a balance of `buyAmount` ETH but we never send it to the taker?

contracts/zero-ex/contracts/src/features/UniswapV2ForkFeature.sol
```diff
@@ -0,0 +1,205 @@
...
+        internal
+        pure
+        returns (address pair)
+    {
+        (IERC20TokenV06 token0, IERC20TokenV06 token1) = _sortTokens(tokenA, tokenB);
+        pair = address(uint(keccak256(abi.encodePacked(
+            hex'ff',
+            factory,
+            keccak256(abi.encodePacked(address(token0), address(token1))),
+            hex'96e8ac4277198ff8b6f785478aa9a39f403cb768dd02cbee326c3e7da348845f' // init code hash TODO(Cece): how can I find the init code for each fork
```
dorothy-zbornak, line 181: Usually their router contract will have the init code hash somewhere in their code.

contracts/zero-ex/contracts/src/features/UniswapV2ForkFeature.sol
cece-z: getting I stack too deep when I put all together, I have to split the swap into two, lmk if that make sense.

cece-z: addressed the recipient issue. thx

contracts/zero-ex/contracts/src/features/UniswapV2ForkFeature.sol
```diff
@@ -0,0 +1,200 @@
...
+        returns (uint[] memory amounts)
+    {
+        require(tokens.length >= 2, 'UniswapV2ForkFeature/INVALID_PATH');
+
+        amounts = new uint[](tokens.length);
+        amounts[0] = sellAmount;
+
+        for (uint i; i < tokens.length - 1; i++) {
+            (IERC20TokenV06 tokenA, IERC20TokenV06 tokenB, IERC20TokenV06 tokenC) =
+                (tokens[i], tokens[i + 1], tokens[i + 2]);
```
dorothy-zbornak, line 173: wont' this revert for paths < 3?

contracts/zero-ex/contracts/src/features/UniswapV2ForkFeature.sol
```diff
@@ -0,0 +1,200 @@
...
+        address to
+    )
+        internal
+        returns (uint amountOut)
+    {
+        IUniswapV2ForkPair pairCurrent = _pairFor(tokenA, tokenB);
+
+        (IERC20TokenV06 token0,) = _sortTokens(tokenA, tokenB);
+        (uint reserve0, uint reserve1,) = pairCurrent.getReserves();
+        (uint reserveIn, uint reserveOut) = tokenA == token0 ? (reserve0, reserve1) : (reserve1, reserve0);
```
dorothy-zbornak, line 193: save some stack space
```suggestion
        (uint256 reserveIn, uint256 reserveOut) = pairCurrent.getReserves();
        (reserveIn, reserveOut) = tokenA == token0 ? (reserveIn, reserveOut) : (reserveOut, reserveIn);
```

contracts/zero-ex/contracts/src/features/UniswapV2ForkFeature.sol
```diff
@@ -0,0 +1,200 @@
...
+        IERC20TokenV06 tokenA,
+        IERC20TokenV06 tokenB,
+        address to
+    )
+        internal
+        returns (uint amountOut)
+    {
+        IUniswapV2ForkPair pairCurrent = _pairFor(tokenA, tokenB);
+
+        (IERC20TokenV06 token0,) = _sortTokens(tokenA, tokenB);
```
dorothy-zbornak, line 191: could just check `tokenA < tokenB` instead of `tokenA == token0` for less gas

contracts/zero-ex/contracts/src/features/UniswapV2ForkFeature.sol
```diff
@@ -0,0 +1,200 @@
...
+        (uint reserve0, uint reserve1,) = pairCurrent.getReserves();
+        (uint reserveIn, uint reserveOut) = tokenA == token0 ? (reserve0, reserve1) : (reserve1, reserve0);
+
+        amountOut = _getAmountOut(amountIn, reserveIn, reserveOut);
+        (uint amount0Out, uint amount1Out) = tokenA == token0 ? (uint(0), amountOut) : (amountOut, uint(0));
+
+        pairCurrent.swap(amount0Out, amount1Out, to, new bytes(0));
```
dorothy-zbornak, line 198: shorthand
```suggestion
        pairCurrent.swap(amount0Out, amount1Out, to, "");
```

contracts/zero-ex/contracts/src/features/UniswapV2ForkFeature.sol
```diff
@@ -0,0 +1,200 @@
...
+        amounts[0] = sellAmount;
+
+        for (uint i; i < tokens.length - 1; i++) {
+            (IERC20TokenV06 tokenA, IERC20TokenV06 tokenB, IERC20TokenV06 tokenC) =
+                (tokens[i], tokens[i + 1], tokens[i + 2]);
+
+            address to = i < tokens.length - 2 ? address(_pairFor(tokenB, tokenC)) : _to;
+            amounts[i + 1] = _swapInternal(amounts[i], tokenA, tokenB, to);
+        }
+    }
```
dorothy-zbornak, line 178: something like
```suggestion
    function _swap(
        IERC20TokenV06[] memory tokens,
        uint256 sellAmount,
        address recipient
    )
        internal
        returns (uint[] memory amounts)
    {
        require(tokens.length >= 2, 'UniswapV2ForkFeature/INVALID_PATH');
    
        IUniswapV2ForkPair currentPair = _pairFor(
            _normalizeToken(tokens[0]),
            _normalizeToken(tokens[1])
        );

        if (tokens[0] == ETH_ADDRESS) {
            _transferERC20Tokens(WETH, currentPair, sellAmount);
        } else {
            _transferERC20TokensFrom(WETH, msg.sender, currentPair, sellAmount);
        }

        for (uint256 i; i < tokens.length - 1; i++) {
            IUniswapV2ForkPair nextPair;
            if (i < tokens.length - 2) {
                nextPair = _pairFor(
                    _normalizeToken(tokens[i + 1]),
                    _normalizeToken(tokens[i + 2])
                );
            }
            sellAmount = _swapInternal(
                _normalizeToken(tokens[i]),
                _normalizeToken(tokens[i + 1]),
                currentPair,
                sellAmount, // sell last bought amount
                address(nextPair) ? address(nextPair) : recipient
            );
            currentPair = nextPair;
        }
        return amount;
    }
```

where `_normalizeToken()` just converts ETH_ADDRESS to WETH

contracts/zero-ex/contracts/src/features/UniswapV2ForkFeature.sol
```diff
@@ -0,0 +1,200 @@
...
+        IERC20TokenV06 tokenA,
+        IERC20TokenV06 tokenB
+    )
+        internal
+        pure
+        returns (IERC20TokenV06 token0, IERC20TokenV06 token1)
+    {
+        require(tokenA != tokenB, 'UniswapV2ForkFeature/IDENTICAL_ADDRESSES');
+        (token0, token1) = tokenA < tokenB ? (tokenA, tokenB) : (tokenB, tokenA);
+        require(token0 != IERC20TokenV06(0), 'UniswapV2ForkFeature/ZERO_ADDRESS');
```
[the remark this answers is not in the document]
cece-z: when front end creates a transaction but forget to set pramas. but I guess it little bit overkill. have removed.

contracts/zero-ex/contracts/src/features/UniswapV2ForkFeature.sol
```diff
@@ -0,0 +1,200 @@
...
+    address private immutable UNI_FACTORY_ADDRESS;
+    /// @dev UniswapV2Fork init code of the factory.
+    bytes32 private immutable UNI_POOL_INIT_CODE_HASH;
+
+    address private constant ETH_ADDRESS = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
+
+    constructor(
+        IEtherTokenV06 weth,
+        address uniV2ForkFactory,
+        bytes32 poolInitCodeHash
```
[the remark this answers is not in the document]
cece-z: shouldn't this contract be instantiated for each fork? regardless to whether we do it in abstract way or not?

contracts/zero-ex/contracts/src/features/UniswapV2ForkFeature.sol
cece-z: I think they are the same? https://github.com/crytic/evm-opcodes

contracts/zero-ex/contracts/src/features/UniswapV2ForkFeature.sol
cece-z: do we need to do `_normalizeToken`  every time, shouldn't we revert once we see a ETH in the middle of the path? I pushed the change without doing so, PTAL, thx!

dorothy-zbornak: ETH/WETH is often in the middle of a path. A->WETH->B is a common hop. We just treat intermediate ETHs as WETH

contracts/zero-ex/contracts/src/features/UniswapV2ForkFeature.sol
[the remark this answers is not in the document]
dorothy-zbornak: No, the original UniswapV2Feature supports multiple forks.

contracts/zero-ex/contracts/src/features/UniswapV2ForkFeature.sol
dorothy-zbornak: I mean if you just replace the `tokenA == token0` with `tokenA < tokenB` you don't have to call `_sortTokens()` to compute `token0` at all.

dorothy-zbornak:
I'm thinking something like
```solidity
abstract contract UniswapV2ForkFeature {
    ...
    function _getForkDetails(fork)
        internal
        pure
        virtual
        returns (address factoryAddress, bytes32 initCodeHash);
    
    function _pairFor(ProtocolFork fork, ...) ... {
        (address factoryAddress, bytes32 initCodeHash) = _getForkDetails(fork);
        (IERC20TokenV06 token0, IERC20TokenV06 token1) = _sortTokens(tokenA, tokenB);
        return IUniswapV2ForkPair(address(uint(keccak256(abi.encodePacked(
            hex'ff',
            factoryAddress,
            keccak256(abi.encodePacked(address(token0), address(token1))),
            initCodeHash
        )))));
    }
    ...
}

contract PolygonUniswapV2ForkFeature is UniswapV2ForkFeature {
    address private constant SUSHISWAP_FACTORY_ADDRESS = 0x...;
    bytes32 private constant SUSHISWAP_INIT_CODE_HASH = 0x...;
    
    function _getForkDetails(fork)
        internal
        pure
        override
        returns (address factoryAddress, bytes32 initCodeHash)
    {
        if (fork == Protocol.SushiSwap) {
            return (SUSHISWAP_FACTORY_ADDRESS, SUSHISWAP_INIT_CODE_HASH);
        } else if (fork == ...) {
            ...
        }
        revert('unsupported uniswap2 fork');
    }
}
```
And we deploy `PolygonUniswapV2ForkFeature` on polygon, `BSCUniswapV2Feature` on BSC, etc.

Also need to add the `fork` param to sellToUniswapV2Fork()`

contracts/zero-ex/contracts/src/features/UniswapV2ForkFeature.sol
cece-z: i guess here is what i confused about, I understand that we could have intermediate WETH, but how we get an intermediate ETH on Uni V2 forks, which they doesnt support? Is it from our centralized service?

dorothy-zbornak: yeah we won't ever get ETH from a uniswap swap, only WETH, so whether we encounter `WETH` or `ETH_ADDDRESS` in `tokens` we will treat it as `WETH` here.

cece-z: sorry, I guess I didnt ask it right, how can we ever encounter `ETH_ADDDRESS` in the **MIDDLE** of `tokens`? Shouldn't we expect the `tokens` is prepared so that it does not include ETH in the middle prior to calling Univ2 feature?

dorothy-zbornak: Yes, it won't ever happen, but the function will treat it as WETH for simplicity.

PR #361, opened 2021-11-02
Feat/extended quote report

Ktl-XV:
## Description
This PR adds an Extended Quote report to `asset-swapper`. This is always computed (instead of only on RFQ trades) to allow for better data analysis.

A "new" quote report was used to have both the regular and extended running during the transition time.

## Testing instructions
Calling `getOptimizerResultAsync` should return the additional (optional) field `extendedQuoteReport` 

See https://github.com/0xProject/0x-api/pull/724

## Types of changes

## Checklist:


packages/asset-swapper/src/utils/quote_report_generator.ts
```diff
@@ -116,6 +137,85 @@ export function generateQuoteReport(
...
+
+    // Dex Quotes
+    sourcesConsidered.push(
+        ..._.flatten(
+            quotes.dexQuotes.map(dex =>
+                dex
+                    .filter(quote => isDexSampleForTotalAmount(quote, marketOperation, amount))
+                    .map(quote => dexSampleToReportSource(quote, marketOperation)),
+            ),
+        ),
```
phil-ociraptor, line 182: Can this be simplified using `flatMap`? 
```
        ...quotes.dexQuotes.flatMap(dex =>
            dex
                .filter(quote => isDexSampleForTotalAmount(quote, marketOperation, amount))
                .map(quote => dexSampleToReportSource(quote, marketOperation)),
        ),

File: packages/asset-swapper/src/utils/quote_report_generator.ts [hunk 2]
```diff
@@ -116,6 +137,85 @@ export function generateQuoteReport(
     };
 }
 
+/**
+ * Generates a report of sources considered while computing the optimized swap quote, the sources ultimately included in the computed quote, and all considered quotes.
```
phil-ociraptor, line 141: can you specify in this comment how this method differs from `generateQuoteReport`?

packages/asset-swapper/src/utils/quote_report_generator.ts
```diff
@@ -116,6 +137,85 @@ export function generateQuoteReport(
...
+ */
+export function generateExtendedQuoteReport(
```
phil-ociraptor, line 143: now that we have `generateExtendedQuoteReport`, is it possible for us to remove `generateQuoteReport`?

phil-ociraptor (commented):
Overall looks great from my side, just some minor comments

packages/asset-swapper/src/utils/quote_report_generator.ts
Ktl-XV: I didn't know about `flatMap` sounds much better, but currently the whole `protocol` repo uses `es2017` and flatMap was introduced in `es2019`. I would prefer not to change it in this PR.

packages/asset-swapper/src/utils/quote_report_generator.ts
Ktl-XV: Not yet, we want a period where both work in parallel to compare across versions

phil-ociraptor (approved):

PR #364, opened 2021-11-07
feat: expose advanced bring your own signature function for limit orders

kimpers:
## Checklist:

### DIFF
--- a/packages/protocol-utils/package.json
+++ b/packages/protocol-utils/package.json
@@ -59,6 +59,7 @@
         "tslint": "5.11.0",
         "typedoc": "~0.16.11",
         "typescript": "4.2.2",
+        "chai": "^4.0.1",
         "web3-provider-engine": "14.0.6"
     },
     "dependencies": {
@@ -69,7 +70,6 @@
         "@0x/subproviders": "^6.6.0",
         "@0x/utils": "^6.4.4",
         "@0x/web3-wrapper": "^7.6.0",
-        "chai": "^4.0.1",
         "ethereumjs-util": "^7.0.10",
         "ethers": "~4.0.4",
         "lodash": "^4.17.11"
--- a/packages/protocol-utils/src/orders.ts
+++ b/packages/protocol-utils/src/orders.ts
@@ -86,6 +86,10 @@ export interface OtcOrderInfo {
     orderHash: string;
 }
 
+type EIP712SignCallbackFunction = (data: EIP712TypedData, signer: string) => Promise<string>;
+type EthSignCallbackFunction = (data: string, signer: string) => Promise<string>;
+type SignatureCallbackFunction = EIP712TypedData | EthSignCallbackFunction;
+
 export abstract class OrderBase {
     public makerToken: string;
     public takerToken: string;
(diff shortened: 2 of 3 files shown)


packages/protocol-utils/src/orders.ts
```diff
@@ -129,6 +135,35 @@ export abstract class OrderBase {
...
+        signFn: SignatureCallbackFunction,
+        type: SignatureType,
+        signer: string = this.maker,
+    ): Promise<Signature> {
+        switch (type) {
+            case SignatureType.EIP712:
+                return eip712SignTypedDataWithSignatureFnAsync(
+                    this.getEIP712TypedData(),
+                    signer,
+                    (signFn as unknown) as EIP712SignCallbackFunction,
```
kimpers, line 154: This is pretty horrible but I couldn't figure out a good way to handle this with proper typing. Open to any suggestions for improvements 😅

packages/protocol-utils/package.json
```diff
@@ -69,7 +70,6 @@
         "@0x/subproviders": "^6.6.0",
         "@0x/utils": "^6.4.4",
         "@0x/web3-wrapper": "^7.6.0",
-        "chai": "^4.0.1",
```
dorothy-zbornak, line 72: :grimacing:

packages/protocol-utils/src/signature_utils.ts
```diff
@@ -85,6 +85,44 @@ export async function eip712SignTypedDataWithProviderAsync(
...
+ * Sign a hash with the EthSign signature type using a compatible signature function
+ */
+export async function ethSignHashWithSignatureFnAsync(
+    hash: string,
+    signer: string,
+    signMessageAsync: (hash: string, signer: string) => Promise<string>,
+): Promise<Signature> {
+    if (typeof hash !== 'string') {
+        throw new Error(`Invalid message parameter for eth sign hash`);
+    }
```
dorothy-zbornak, line 98: Are these checks necessary?

packages/protocol-utils/src/orders.ts
dorothy-zbornak: wow it really won't let you just do a direct cast?

packages/protocol-utils/src/orders.ts
```diff
@@ -129,6 +135,35 @@ export abstract class OrderBase {
         }
     }
 
+    /*
+     * This is an advanced order signature function that should only be used if you know what you're doing.
+     * It allows you to bring your own wallet signature callback function, this can be used to sign orders
+     * with a wallet that is otherwise unsupported by Web3Wrapper.
+     * The signature function needs to implement correct signing of the data and return the results as a string.
+     */
+    public async _getSignatureWithSignFunctionAsync(
```
dorothy-zbornak, line 144: why _ prefix?

dorothy-zbornak, line 144: I don't really get the value this function adds. Why not just export `parseRpcSignature()`? Sure we call either `getEIP712TypedData()` or `getHash()` based on `type`, but it seems like the caller already has to know in advance which one they're working with because they have to pass in the function that accepts it. Or is the prescription that the callback should handle both cases? In which case shouldn't it be `type SignatureCallbackFunction = (dataOrHash: EIP712TypedData | string, signer: string) => Promise<string>;`?

kimpers: It doesn't really add much value tbh! I was initially gonna do a big refactor and create a minimal dependency file for this logic but then I got lazy and appreciative of the convenience of the `LimitOrder` object. 

But yeah, maybe I should give that another stab and realize the original vision 🤓  We are getting so rekt on bundle size because we don't use `import type {} from '@0x/...` when we only need the types 🙈 

I'll fix this tomorrow 👍

packages/protocol-utils/src/orders.ts
kimpers: nah TS was completely freaking out over the lack of overlap 🙈

packages/protocol-utils/src/orders.ts
kimpers: to deter people from using this function 😂

packages/protocol-utils/src/signature_utils.ts
kimpers: With the current typing madness yeah, it would not stop you from passing `SignatureType.EIP712` and a string, or vice versa

kimpers:
I ended up porting all the required code for a minimal limit order implementation to Matcha so this is no longer needed 🎉  Closing the PR 🍻

PR #365, opened 2021-11-09
Add ETH leak modifier to mtx functions

dorothy-zbornak:
## Description

Add a check to ensure mtxs do not cause the EP's ETH balance to go down.

## Checklist:

### DIFF
--- a/contracts/zero-ex/CHANGELOG.json
+++ b/contracts/zero-ex/CHANGELOG.json
@@ -1,5 +1,14 @@
 [
     {
+        "version": "0.29.4",
+        "changes": [
+            {
+                "note": "Prevent EP ETH balance from reducing when executin mtxs",
+                "pr": 365
+            }
+        ]
+    },
+    {
         "version": "0.29.3",
         "changes": [
             {
--- a/contracts/zero-ex/contracts/src/features/MetaTransactionsFeature.sol
+++ b/contracts/zero-ex/contracts/src/features/MetaTransactionsFeature.sol
@@ -78,7 +78,7 @@ contract MetaTransactionsFeature is
     /// @dev Name of this feature.
     string public constant override FEATURE_NAME = "MetaTransactions";
     /// @dev Version of this feature.
-    uint256 public immutable override FEATURE_VERSION = _encodeVersion(1, 2, 0);
+    uint256 public immutable override FEATURE_VERSION = _encodeVersion(1, 2, 1);
     /// @dev EIP712 typehash of the `MetaTransactionData` struct.
     bytes32 public immutable MTX_EIP712_TYPEHASH = keccak256(
         "MetaTransactionData("
@@ -103,6 +103,17 @@ contract MetaTransactionsFeature is
         if (remainingBalance > 0) {
             msg.sender.transfer(remainingBalance);
         }
+    }
+
+    /// @dev Ensures that the ETH balance of `this` does not go below the
+    ///      initial ETH balance before the call (excluding ETH attached to the call).
+    modifier doesNotReduceEthBalance() {
+        uint256 initialBalance = address(this).balance - msg.value;
+        _;
+        require(
+            initialBalance <= address(this).balance,
+            "MetaTransactionsFeature/ETH_LEAK"
+        );
     }
 
     constructor(address zeroExAddress)
@@ -140,6 +151,7 @@ contract MetaTransactionsFeature is
         payable
         override
         nonReentrant(REENTRANCY_MTX)
+        doesNotReduceEthBalance
         refundsAttachedEth
         returns (bytes memory returnResult)
     {
@@ -164,6 +176,7 @@ contract MetaTransactionsFeature is
         payable
         override
         nonReentrant(REENTRANCY_MTX)
+        doesNotReduceEthBalance
         refundsAttachedEth
         returns (bytes[] memory returnResults)
     {
--- a/contracts/zero-ex/contracts/test/TestMetaTransactionsTransformERC20Feature.sol
+++ b/contracts/zero-ex/contracts/test/TestMetaTransactionsTransformERC20Feature.sol
@@ -46,6 +46,10 @@ contract TestMetaTransactionsTransformERC20Feature is
         payable
         returns (uint256 outputTokenAmount)
     {
+        if (msg.value == 555) {
+            tx.origin.transfer(1);
+        }
+
         if (msg.value == 666) {
             revert('FAIL');
         }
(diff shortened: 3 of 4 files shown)


dextracker (approved):
Looks good to me!

mzhu25 (approved):
yup

shawnxin89 (approved):
LGTM

ytrezq:
@dorothy-zbornak @mzhu25 @dextracker hello, sorry, but why is it a problem ?
Isn’t 0x supposed to be a non‑custodial exchange ? Also it’s not like there’s a trade pressure when money is sent to 0x anyway. What can be sure is there’s something like months without any attempt to get it back from depositors.

dorothy-zbornak:
> @dorothy-zbornak @mzhu25 @dextracker hello, sorry, but why is it a problem ? Isn’t 0x supposed to be a non‑custodial exchange ? Also it’s not like there’s a trade pressure when money is sent to 0x anyway. What can be sure is there’s something like months without any attempt to get it back from depositors.

You're right that it isn't custodial, which is why there's no reason to allow a user to withdraw more ETH than they passed in. Limiting behavior to the intended use case is good security hygiene imo.

ytrezq:
@dorothy-zbornak Ok, but then I understand why preveting someone from depositing Ether which by the way you aren t doing. But preventing to get it back?
There s no reason the main exchange address should have a balance of anything.

dorothy-zbornak:
> 

I doubt anyone has ever recovered *their own* ETH via this mechanism. It's too exotic and any public attempt could easily get frontrunned. So I don't agree with the argument that this was a realistic recovery method for users. By plugging this leak, we can, and have, recovered funds for users that have requested it. That's currently the most reliable avenue. If you can think of a way to prevent people from transferring tokens and ETH directly into the Exchange Proxy, we're all ears.

ytrezq:
@dorothy-zbornak attempts to recover the money can t be frontrunned because of the signature. Bots only detects transactions that don t require modification of the input data.

Then, well, I recognize this won t work for tokens, but as Ether implies calling, simply check the balance before is equal to the balance prior, and not just below.

dorothy-zbornak:
> @dorothy-zbornak attempts to recover the money can t be frontrunned because of the signature. Bots only detects transactions that don t require modification of the input data. Also, as developpers, you tend to not read the requests on the official forum. Or have spambots which are too touth by error on anything https://www.reddit.com/r/0xProject/comments/r5c109/please_review_my_temporary_ban_on_the_official/. Just notice the current token balance of the contract since November.
> 
> Then, well, I recognize this won t work for tokens, but as Ether implies calling, simply check the balance before is equal to the balance prior, and not just below.

It *can*, and eventually will, be frontrunned because multiple people know how to craft this transaction at this point. So I don't believe it's a sustainable solution, even in the short term.

I don't really follow your suggestion with the ether. 
People tend to send ether to the contract as a single transaction. Perhaps you can elaborate.

ytrezq:
@dorothy-zbornak Openethereum/parity tracing is single threaded, with the signatures, it s hard to determine if the the transaction  is about getting the Ethers in the specific way explained.

In your, fix, instead of checking bellow or equal `<=`, just require balance to be equal to the balance prior the transaction `==`

dorothy-zbornak:
> @dorothy-zbornak Openethereum/parity tracing is single threaded, with the signatures, it s hard to determine if the the transaction is about getting the Ethers in the specific way explained.
> 
> In your, fix, instead of checking bellow or equal `<=`, just require balance to be equal to the balance prior the transaction `==`

Yes, an unaware bot wouldn't catch it today, but it's trivial to write one that can generate a proper signature and pull the ETH before someone tries to legitimately recover it. You've suggested that you've already done this in other channels. Bottom line, it's really hard to argue that the old behavior would be regularly used legitimately or in a way that is good for the protocol.

As for the `==` check, it's something probably fine, but like I said, this is not how ETH winds up in the contract.

ytrezq:
@dorothy-zbornak then this isn t frontrunning per se but being the first to take the money. The mechanism is not that well known and there can be months with Ether staying on the contract untouched. It s hard to detect if a transaction is about rescuing Ethers in terms of gas price as currently written (frontrunning require adjusting the priority fee).

Also, as developpers, you tend to not read the requests on the official forum. Or have spambots which are too touth by error on anything https://www.reddit.com/r/0xProject/comments/r5c109/please_review_my_temporary_ban_on_the_official/. So a general mechanism for recovering is suitable even if it can be frontrunnned which by the way is the choice of some erc20 token projects.

Concerning, the Ethers being sent to the contract, do you mean without calling a specific function? Then just check `msg.value==0` on fallback too.

dorothy-zbornak:
> @dorothy-zbornak then this isn t frontrunning per se but being the first to take the money. The mechanism is not that well known and there can be months with Ether staying on the contract untouched. It s hard to detect if a transaction is about rescuing Ethers in terms of gas price as currently written (frontrunning require adjusting the priority fee).
> 
> Also, as developpers, you tend to not read the requests on the official forum. Or have spambots which are too touth by error on anything https://www.reddit.com/r/0xProject/comments/r5c109/please_review_my_temporary_ban_on_the_official/. So a general mechanism for recovering is suitable even if it can be frontrunnned which by the way is the choice of some erc20 token projects.
> 
> Concerning, the Ethers being sent to the contract, do you mean without calling a specific function? Then just check `msg.value==0` on fallback too.

If you are convinced that there is value to the protocol in adding a free-for-all mechanism to withdraw tokens from the Exchange Proxy, I encourage you to create a proposal on the forums with proper justification and see if users agree. It is out of the scope of this particular feature.

We can't practically block ETH in the fallback because the EP sometimes needs to make ETH swaps with other protocols (including WETH).

ytrezq:
For starting a proposal, the problem is I m censored by spambots and that mods aren t replying https://www.reddit.com/r/0xProject/comments/r5c109/please_review_my_temporary_ban_on_the_official/.

Then, in fallback just revert on `tx.origin==msg.sender`

dorothy-zbornak:
> For starting a proposal, the problem is I m censored by spambots and that mods aren t replying https://www.reddit.com/r/0xProject/comments/r5c109/please_review_my_temporary_ban_on_the_official/.
> 
> Then, in fallback just revert on `tx.origin==msg.sender`

That's actually not a bad idea. I'm sure some of the ETH comes in through contracts as well, but this would help mitigate mistakes from EOAs. Unfortunately the receive function on the EP is not upgradable so we could only get this change in on new networks. 😩

If you want to dispute your ban, I doubt you'll get very far in a PR. The discord is probably your best bet of reaching an admin.

PR #369, opened 2021-11-17
update package versions for protocol contracts

dextracker:
## Checklist:


packages/protocol-utils/package.json
```diff
@@ -64,7 +64,7 @@
     "dependencies": {
         "@0x/assert": "^3.0.29",
         "@0x/contract-addresses": "^6.9.0",
-        "@0x/contract-wrappers": "^13.18.3",
+        "@0x/contract-wrappers": "^13.18.2",
```
dekz, line 67: Weird, doesn't lerna do this for us automatically @merklejerk

dekz: Yeah I think if you rebase you'll see these changes should've already been applied 

https://github.com/0xProject/protocol/blob/development/packages/protocol-utils/package.json#L67

PR #376, opened 2021-11-29
Sushi router celo fix

dextracker:
## Checklist:


packages/asset-swapper/CHANGELOG.json
```diff
@@ -6,8 +6,7 @@
                 "note": "Changed Sushiswap router address",
                 "pr": 373
             }
-        ],
-        "timestamp": 1637349338
+        ]
```
dekz, line 9:
```suggestion
        ],
        "timestamp": 1637349338
```

This log is append only, can't modify the previous entries.

PR #379, opened 2021-12-02
v4 ERC721 orders

mzhu25:
## Description

Introduces `ERC721Order`s to 0x V4. This PR includes the smart contracts for buying and selling ERC721 assets.
tests in #381 

## Checklist:


contracts/zero-ex/contracts/src/features/libs/LibSignature.sol
```diff
@@ -44,7 +44,8 @@ library LibSignature {
         ILLEGAL,
         INVALID,
         EIP712,
-        ETHSIGN
+        ETHSIGN,
+        PRESIGNED
```
mzhu25, line 48: This has a different value than in v3, idk if that'll be a problem

mzhu25: could pad this enum with `WALLET` and `VALIDATOR` so everything is aligned with v3

contracts/zero-ex/contracts/src/features/ERC721OrdersFeature.sol
```diff
@@ -0,0 +1,1047 @@
...
+        // Refund
+        if (ethSpent < msg.value) {
+            _transferEth(msg.sender, msg.value - ethSpent);
+        }
+    }
+
+    /// @dev Cancel a single ERC721 order. The caller must be the maker.
+    ///      Silently succeeds if the order has already been cancelled.
+    /// @param order The ERC721 order.
+    function cancelERC721Order(LibERC721Order.ERC721Order memory order)
```
mzhu25, line 153: maybe change this function signature to just:
`function cancelERC721Order(address maker, uint256 nonce)`
for gas efficiency?

contracts/zero-ex/contracts/src/features/ERC721OrdersFeature.sol
```diff
@@ -0,0 +1,1041 @@
...
+        _setOrderStatusBit(order.maker, order.nonce);
+
+        // Transfer the ERC721 asset to the buyer (`msg.sender`).
+        _transferERC721AssetFrom(
+            order.erc721Token,
+            order.maker,
+            msg.sender,
+            order.erc721TokenId
+        );
+
```
mzhu25, line 670: Thinking about throwing in a callback to `msg.sender` here (and in `_sellERC721` too), so we have a flow similar to UniV3:
```
maker -> taker transfer
taker callback
taker -> maker transfer
```
This would enable matching 0x orders with e.g. Wyvern orders without a flash loan
Would make the `ethAvailable` accounting a bit more complex though but I think we can just do something like this:
```
uint256 balanceBeforeCallback = address(this).balance;
// taker callback
ethAvailable += address(this).balance - balanceBeforeCallback;
```

contracts/zero-ex/contracts/src/vendor/IFeeRecipient.sol
```diff
@@ -0,0 +1,44 @@
...
+    ///      to implement arbitrary fee-handling logic, e.g. splitting the fee 
+    ///      between multiple parties. 
+    /// @param tokenAddress The address of the token in which the received fee is 
+    ///        denominated. `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE` indicates 
+    ///        that the fee was paid in the native token (e.g. ETH).
+    /// @param amount The amount of the given token received.
+    /// @param feeData Arbitrary data encoded in the `Fee` used by this callback.
+    /// @return success The selector of this function (0x0190805e), 
+    ///         indicating that the callback succeeded.
+    function receiveFeeCallback(
```
mzhu25, line 37: maybe change this to `receiveZeroExFeeCallback` or something?

contracts/zero-ex/contracts/src/features/ERC721OrdersFeature.sol
mzhu25: (I did)

contracts/zero-ex/contracts/src/features/ERC721OrdersFeature.sol
mzhu25: (I did this too)

contracts/zero-ex/contracts/src/features/ERC721OrdersFeature.sol
```diff
@@ -0,0 +1,1112 @@
...
+    function getERC721OrderStatus(LibERC721Order.ERC721Order memory order)
+        public
+        override
+        view
+        returns (LibERC721Order.OrderStatus status)
+    {
+        // Only sell orders with `erc721TokenId` == 0 can be property
+        // orders. 
+        if (order.erc721TokenProperties.length > 0 && 
+                (order.direction == LibERC721Order.TradeDirection.SELL_721 ||
```
kimpers, line 1043: should this be `||` or `&&`?

mzhu25: Ah yes `||` is correct but the comment is wrong; should read "Only buy orders..."

contracts/zero-ex/contracts/src/features/ERC721OrdersFeature.sol
```diff
@@ -0,0 +1,1112 @@
...
+                // TODO: Probably safe to just use WETH.transfer for some
+                //       small gas savings
+                // Transfer WETH to the seller.
+                _transferERC20Tokens(
+                    WETH,
+                    order.maker,
+                    order.erc20TokenAmount
+                );
+                // Pay fees using ETH. 
+                uint256 ethFees = _payFees(order, address(this), true);
```
dorothy-zbornak, line 767: `ethAvailable >= order.erc20TokenAmount` doesn't account for fees though? So it's possible this'll run out?

dorothy-zbornak: idk should `_payFees()` intelligently wrap on the fly? maybe we write a smart `transferTokens()` function that wraps on the fly and handles all the edge cases? Is that possible?

dorothy-zbornak: Or I guess we're OK with saying that if you're filling with ETH then you must supply enough to cover the fees as well. Also, I know we also discussed this before but it feels worse in person that fees can come in either ETH or WETH at the mercy of the taker. At the very least we should call out that we've accepted this edge case with a comment for the auditors.

contracts/zero-ex/contracts/src/features/ERC721OrdersFeature.sol
```diff
@@ -0,0 +1,1112 @@
...
+            } else {
+                // Transfer WETH from the buyer to the seller.
+                _transferERC20TokensFrom(
+                    order.erc20Token,
+                    msg.sender,
+                    order.maker,
+                    order.erc20TokenAmount
+                );
+                // The buyer pays fees using WETH.
+                _payFees(order, msg.sender, false);
```
dorothy-zbornak, line 779: can this be refactored to fall through to the dupe code below to reduce LoC? :sweat_smile:

dorothy-zbornak (commented):
still going...

contracts/zero-ex/contracts/src/features/interfaces/IERC721OrdersFeature.sol
```diff
@@ -0,0 +1,283 @@
...
+        bytes calldata data
+    )
+        external
+        returns (bytes4 success);
+
+    /// @dev Approves an ERC721 order hash on-chain. After pre-signing 
+    ///      a hash, the `PRESIGNED` signature type will become valid 
+    ///      for that order and maker.
+    /// @param orderHash An ERC721 order hash.
+    function preSignERC721Order(bytes32 orderHash)
```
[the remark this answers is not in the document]
mzhu25: Yeah, I originally had a registry but I guess with stuff like partybid you might not want to delegate signing powers to a single EOA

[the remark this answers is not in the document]
mzhu25: @steveklebanoff can provide color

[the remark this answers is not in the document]
dorothy-zbornak: isn't partybid always taker?

contracts/zero-ex/contracts/src/fixins/FixinCommon.sol
```diff
@@ -62,6 +62,18 @@ abstract contract FixinCommon {
         _implementation = address(this);
     }
 
+    /// @dev RichError equivalent of `require`. 
+    /// @param condition If `condition` is false, this function will revert.
+    /// @param errorData ABI-encoded error data.
+    function rrequire(bool condition, bytes memory errorData)
```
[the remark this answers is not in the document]
mzhu25: ahh 🧠

contracts/zero-ex/contracts/src/features/ERC721OrdersFeature.sol
mzhu25: Yeah, we ran this by some potential integrators and they seem to be ok with the fee recipients potentially receiving a mix of ETH/WETH. But yeah we can specifically call this out for auditors

contracts/zero-ex/contracts/src/features/ERC721OrdersFeature.sol
```diff
@@ -0,0 +1,1112 @@
...
+            );
+        }
+
+        // Validate each property
+        for (uint256 i = 0; i < order.erc721TokenProperties.length; i++) {
+            LibERC721Order.Property memory property = order.erc721TokenProperties[i];
+            // `address(0)` is interpreted as a no-op. Any token ID
+            // will satisfy a property with `propertyValidator == address(0)`.
+            if (address(property.propertyValidator) == address(0)) {
+                continue;
```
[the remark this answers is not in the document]
mzhu25: no, but I don't think it really hurts to `continue` instead of `return` here

contracts/zero-ex/contracts/src/features/ERC721OrdersFeature.sol
```diff
@@ -0,0 +1,1112 @@
...
+
+        if (takerCallbackData.length > 0) {
+            require(
+                msg.sender != address(this), 
+                "ERC721OrdersFeature::_buyERC721/CANNOT_CALLBACK_SELF"
+            );
+            uint256 ethBalanceBeforeCallback = address(this).balance;
+            // Invoke the callback
+            bytes4 callbackResult = IERC721OrderCallback(msg.sender)
+                .zeroExERC721OrderCallback(takerCallbackData);
```
dorothy-zbornak, line 725: Just calling out that we need to be mindful about this pattern in the future. If these contracts rely on checking `msg.sender == EP` to do something that costs someone money, it's a potential attack vector against them if we somehow let an arbitrary call to one of these addresses slide through. We've potentially expanded our liability here.

contracts/zero-ex/contracts/src/features/ERC721OrdersFeature.sol
```diff
@@ -0,0 +1,1112 @@
...
+                _transferERC20TokensFrom(
+                    buyOrder.erc20Token,
+                    buyOrder.maker,
+                    msg.sender,
+                    profit
+                );
+            }
+        }
+
+        // TODO: or we could emit 2 ERC721OrderFilled events
```
dorothy-zbornak, line 448: I like that. Maybe those events could also have a `matcher` param to differentiate.

contracts/zero-ex/contracts/src/features/ERC721OrdersFeature.sol
```diff
@@ -0,0 +1,1112 @@
...
+    /// @return profits The amount of profit earned by the caller
+    ///         of this function for each pair of matched orders
+    ///         (denominated in the ERC20 token of the order pair).
+    /// @return successes An array of booleans corresponding to
+    ///         whether each pair of orders was successfully matched.
+    function batchMatchERC721Orders(
+        LibERC721Order.ERC721Order[] memory sellOrders,
+        LibERC721Order.ERC721Order[] memory buyOrders,
+        LibSignature.Signature[] memory sellOrderSignatures,
+        LibSignature.Signature[] memory buyOrderSignatures
```
dorothy-zbornak, line 480: should *this* have a `revertIfIncomplete`?

contracts/zero-ex/contracts/src/features/ERC721OrdersFeature.sol
```diff
@@ -0,0 +1,1112 @@
...
+                returnData.rrevert();
+            }
+
+            rrequire(
+                ethSpent <= msg.value,
+                LibERC721OrdersRichErrors.OverspentEthError(
+                    ethSpent,
+                    msg.value   
+                )
+            );
```
dorothy-zbornak, line 248: could just move this outside the loop and check it once

contracts/zero-ex/contracts/src/features/interfaces/IERC721OrdersFeature.sol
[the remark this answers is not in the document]
dorothy-zbornak: what if we just deleted this feature :roll_eyes:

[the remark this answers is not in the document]
dorothy-zbornak: coco chanel did say before you leave the house, look in the mirror and take one thing off

contracts/zero-ex/contracts/src/features/libs/LibERC721Order.sol
```diff
@@ -0,0 +1,269 @@
...
+        //     feesHash,
+        //     order.maker,
+        //     order.taker,
+        //     order.expiry,
+        //     order.nonce
+        // ));
+        assembly {
+            let pos1 := sub(order, 32)
+            let pos2 := add(order, 160)
+            let pos3 := add(order, 192)
```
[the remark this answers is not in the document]
dorothy-zbornak: nvm I forgot yul sucks

contracts/zero-ex/contracts/src/features/interfaces/IERC721OrdersFeature.sol
```diff
@@ -0,0 +1,283 @@
...
+        IERC20TokenV06 erc20Token,
+        IERC721Token erc721Token,
+        uint256 erc20TokenAmount,
+        uint256 erc721TokenId,
+        address sellOrderMaker,
+        address buyOrderMaker,
+        uint256 sellOrderNonce,
+        uint256 buyOrderNonce,
+        address matcher,
+        uint256 profit
```
dorothy-zbornak, line 79: Is this worth logging? Matches are always deterministic so it can be computed off-chain.

contracts/zero-ex/contracts/src/features/ERC721OrdersFeature.sol
```diff
@@ -0,0 +1,1112 @@
...
+
+    // Core settlment logic for selling an ERC721 asset. 
+    // Used by `sellERC721` and `onERC721Received`. 
+    function _sellERC721(
+        LibERC721Order.ERC721Order memory order,
+        LibSignature.Signature memory signature,
+        uint256 erc721TokenId,
+        bool unwrapNativeToken,
+        address taker,
+        bool isCallback,
```
dorothy-zbornak, line 600: I'm thinking something less edge-case sounding like
```suggestion
        address owner,
```

contracts/zero-ex/contracts/src/features/ERC721OrdersFeature.sol
```diff
@@ -0,0 +1,1112 @@
...
+    )
+        private
+        returns (uint256 totalFeesPaid)
+    {
+        for (uint256 i = 0; i < order.fees.length; i++) {
+            LibERC721Order.Fee memory fee = order.fees[i];
+            
+            require(
+                fee.recipient != address(this), 
+                "ERC721OrdersFeature::_payFees/RECIPIENT_CANNOT_BE_EXCHANGE_PROXY"
```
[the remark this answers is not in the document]
mzhu25: Probably not, but I have this here for the same reason we have this check in `FixinTokenSpender` and such, i.e. we don't want to avoid rogue self-calls as much as possible

contracts/zero-ex/contracts/src/features/ERC721OrdersFeature.sol
```diff
@@ -0,0 +1,1112 @@
...
+    )
+        public
+        override
+        returns (uint256 profit)
+    {
+        _validateSellOrder(sellOrder, sellOrderSignature);
+        _validateBuyOrder(
+            buyOrder,
+            buyOrderSignature,
+            msg.sender, // TODO: maybe allow sellOrder.taker == buyOrder.maker, vice versa
```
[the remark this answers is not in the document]
mzhu25: yea had I put `msg.sender` here initially because that was the behavior in v3, but was unsure what would be better

contracts/zero-ex/contracts/src/features/ERC721OrdersFeature.sol
mzhu25: idk, I can't really think of a reason people would ever set `revertIfIncomplete` to true?

dorothy-zbornak: Yeah, I guess. I added it in my branch but you can take it out.

mzhu25:
superseded by https://github.com/0xProject/protocol/pull/429

PR #387, opened 2021-12-29
fix: don't create fills for 0 output samples and negative adjusted rate orders

kimpers:
## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
@@ -1,7 +1,7 @@
 import { assert } from '@0x/assert';
 import { ChainId } from '@0x/contract-addresses';
 import { OptimizerCapture, route, SerializedPath } from '@0x/neon-router';
-import { BigNumber } from '@0x/utils';
+import { BigNumber, hexUtils } from '@0x/utils';
 import * as _ from 'lodash';
 import { performance } from 'perf_hooks';
 
@@ -21,6 +21,8 @@ const FILL_QUOTE_TRANSFORMER_GAS_OVERHEAD = new BigNumber(150e3);
 const MIN_NUM_SAMPLE_INPUTS = 3;
 
 const isDexSample = (obj: DexSample | NativeOrderWithFillableAmounts): obj is DexSample => !!(obj as DexSample).source;
+
+type SourcePathIdMap = { [key in ERC20BridgeSource]: string };
 
 function nativeOrderToNormalizedAmounts(
     side: MarketOperation,
@@ -211,6 +213,17 @@ function findRoutesAndCreateOptimalPath(
     const adjustedFills: Fill[] = [];
     const totalRoutedAmount = BigNumber.sum(...allSourcesRustRoute);
 
+    const sourcePathIdMap = ({} as any) as SourcePathIdMap;
+    const getSourcePathId = (key: ERC20BridgeSource) => {
+        if (sourcePathIdMap[key]) {
+            return sourcePathIdMap[key];
+        }
+
+        const sourcePathId = hexUtils.random();
+        sourcePathIdMap[key] = sourcePathId;
+        return sourcePathId;
+    };
+
     const scale = input.dividedBy(totalRoutedAmount);
     for (const [routeInput, routeSamplesAndNativeOrders, outputAmount] of routesAndSamplesAndOutputs) {
         if (!routeInput || !routeSamplesAndNativeOrders || !outputAmount || !Number.isFinite(outputAmount)) {
@@ -237,7 +250,7 @@ function findRoutesAndCreateOptimalPath(
             // and nativeFill will be `undefined`
             if (nativeFill) {
                 // NOTE: For Limit/RFQ orders we are done here. No need to scale output
-                adjustedFills.push(nativeFill);
+                adjustedFills.push({ ...nativeFill, sourcePathId: getSourcePathId(ERC20BridgeSource.Native) });
             }
             continue;
         }
@@ -275,6 +288,7 @@ function findRoutesAndCreateOptimalPath(
             }
         }
 
+        const sourcePathId = getSourcePathId(current.source);
         // TODO(kimpers): remove once we have solved the rounding/precision loss issues in the Rust router
         const scaleOutput = (fillInput: BigNumber, output: BigNumber) =>
             output
@@ -288,6 +302,7 @@ function findRoutesAndCreateOptimalPath(
             adjustedOutput: scaleOutput(fill.input, fill.adjustedOutput),
             index: 0,
             parent: undefined,
+            sourcePathId,
         });
     }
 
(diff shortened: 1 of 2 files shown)


packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -77,8 +77,15 @@ function findRoutesAndCreateOptimalPath(
     fees: FeeSchedule,
     neonRouterNumSamples: number,
 ): Path | undefined {
-    const createFill = (sample: DexSample) =>
-        dexSamplesToFills(side, [sample], opts.outputAmountPerEth, opts.inputAmountPerEth, fees)[0];
+    const createFill = (sample: DexSample): Fill | undefined => {
+        const fills = dexSamplesToFills(side, [sample], opts.outputAmountPerEth, opts.inputAmountPerEth, fees);
```
dorothy-zbornak, line 81: Won't fills from the same source be disconnected from each other? `dexSamplesToFills()` generates a `sourcePathId` common to all samples passed in. But since you call this sample by sample, they'll have different `sourcePathId`s which might mess with other logic in AS ([example](https://github.com/0xProject/protocol/blob/kim/fix/fix-creating-fills-for-zero-output/packages/asset-swapper/src/utils/market_operation_utils/path.ts#L257)). `index` and `parent` fields will also be disconnected (always `0` and `undefined`). Maybe those were just used during routing though.

kimpers: `index` and `parent` are fine, they are intentionally overridden to `0` and `undefined` when creating the adjusted fills anyway.

The effect on `sourcePathId` is true, good catch! 🙏 The router doesn't care about them, but I'm not sure what the side effect of not merging fills when collapsing. Any ideas?

I think the medium-term plan is to try to move away from the fills/path abstraction into something that is leaner and closer to what we use in the Rust router (lots of unnecessary data-wrangling going on right now with unclear benefit).

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -249,36 +259,37 @@ function findRoutesAndCreateOptimalPath(
         assert.assert(routeSamples.length >= 1, 'Found no sample to use for source');
         for (let k = routeSamples.length - 1; k >= 0; k--) {
             if (k === 0) {
-                fill = createFill(routeSamples[0]);
+                fill = createFill(routeSamples[0]) ?? fill;
```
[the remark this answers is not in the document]
kimpers: It was added in TS 3.7, so fairly recently that protocol got access to it 😂

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
dorothy-zbornak: I think it might possibly cause us to swap on (for example) uniswap 13 times even though it's all the same. :shrug:

kimpers: Well that would be bad :feelsgood:

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -233,14 +253,21 @@ function findRoutesAndCreateOptimalPath(
...
                 fees,
-            )[0];
-            // NOTE: For Limit/RFQ orders we are done here. No need to scale output
-            adjustedFills.push(nativeFill);
+            )[0] as Fill | undefined;
+            // Note: If the order has an adjusted rate of less than or equal to 0 it will be skipped
+            // and nativeFill will be `undefined`
+            if (nativeFill) {
+                // NOTE: For Limit/RFQ orders we are done here. No need to scale output
+                adjustedFills.push({ ...nativeFill, sourcePathId: getSourcePathId(ERC20BridgeSource.Native) });
```
kimpers, line 261: Do we wanna do this for all native orders or do we need to separate them into RFQ and Limit?

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -212,6 +221,17 @@ function findRoutesAndCreateOptimalPath(
     const adjustedFills: Fill[] = [];
     const totalRoutedAmount = BigNumber.sum(...allSourcesRustRoute);
 
+    const sourcePathIdMap = ({} as any) as SourcePathIdMap;
+    const getSourcePathId = (key: ERC20BridgeSource) => {
```
dorothy-zbornak, line 225: We often have multiple source paths for a single source. E.g., A->B @ uniswap, A->C->B @ uniswap, etc.

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
dorothy-zbornak: For native orders, I think the sourcePathId doesn't really matter. They always get treated as separate orders when they get serialized.

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -233,14 +246,21 @@ function findRoutesAndCreateOptimalPath(
...
                 fees,
-            )[0];
-            // NOTE: For Limit/RFQ orders we are done here. No need to scale output
-            adjustedFills.push(nativeFill);
+            )[0] as Fill | undefined;
+            // Note: If the order has an adjusted rate of less than or equal to 0 it will be skipped
+            // and nativeFill will be `undefined`
+            if (nativeFill) {
+                // NOTE: For Limit/RFQ orders we are done here. No need to scale output
+                adjustedFills.push({ ...nativeFill, sourcePathId: sourcePathId ?? hexUtils.random() });
```
dorothy-zbornak, line 254: How can `sourcePathId` be nullish here? I think it's probably fine to just always do `hexUtils.random()` here though.

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -249,38 +269,40 @@ function findRoutesAndCreateOptimalPath(
...
         adjustedFills.push({
             ...fill,
             input: rustInputAdjusted,
-            output: scaleOutput(fill.output),
-            adjustedOutput: scaleOutput(fill.adjustedOutput),
+            output: scaleOutput(fill.input, fill.output),
+            adjustedOutput: scaleOutput(fill.input, fill.adjustedOutput),
             index: 0,
             parent: undefined,
+            sourcePathId: sourcePathId ?? hexUtils.random(),
```
dorothy-zbornak, line 305: Shouldn't `sourcePathId` always be defined here?

dorothy-zbornak (approved):
:+1: :eye: :lips: :eye: :+1:

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
kimpers: Yeah, it should always be defined here since we are zipping arrays of the same length, but TS doesn't know that. I figured we might as well fall back to generating a random id rather than overriding the types and risk future changes breaking that invariant 😄

PR #388, opened 2021-12-30
feat: Capture Routing timing metrics

dekz:
## Description

Allows for timing data to be collected around Routing

## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
@@ -5,14 +5,13 @@ import { BigNumber } from '@0x/utils';
 import * as _ from 'lodash';
 import { performance } from 'perf_hooks';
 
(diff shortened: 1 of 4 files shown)


packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -319,6 +307,13 @@ export function findOptimalRustPathFromSamples(
         fees,
         neonRouterNumSamples,
     );
+    // tslint:disable-next-line: no-unused-expression
+    samplerMetrics &&
+        samplerMetrics.logRouterDetails({
+            router: 'neon-router',
+            type: 'all',
+            timingMs: performance.now() - beforeTimeMs,
```
kimpers, line 315: NIT: you could use `beforeAllMs` here, then you could create `beforeTimeMs` inside the vip block and wouldn't need to reassign it. Same results but might be slightly more readable? 🤔

kimpers (approved):
NOICE 🙏

PR #397, opened 2022-01-14
feat: Improve Uniswap V3 gas schedule

kimpers:
## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/market_operation_utils/constants.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/constants.ts
@@ -2338,7 +2338,8 @@ export const DEFAULT_GAS_SCHEDULE: Required<FeeSchedule> = {
     },
     [ERC20BridgeSource.UniswapV3]: (fillData?: FillData) => {
         const uniFillData = fillData as UniswapV3FillData | FinalUniswapV3FillData;
-        // NOTE: This value was picked by just looking at commonly occurring amounts for underestimations
+        // NOTE: This base value was heuristically chosen by looking at how much it generally
+        // underestimated gas usage
         let gas = 34e3; // 34k base
         if (isFinalUniswapV3FillData(uniFillData)) {
             gas += uniFillData.gasUsed;
--- a/packages/asset-swapper/src/utils/market_operation_utils/constants.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/constants.ts
@@ -2340,7 +2340,8 @@ export const DEFAULT_GAS_SCHEDULE: Required<FeeSchedule> = {
         const uniFillData = fillData as UniswapV3FillData | FinalUniswapV3FillData;
         // NOTE: This base value was heuristically chosen by looking at how much it generally
         // underestimated gas usage
-        let gas = 34e3; // 34k base
+        const base = 34e3; // 34k base
+        let gas = base;
         if (isFinalUniswapV3FillData(uniFillData)) {
             gas += uniFillData.gasUsed;
         } else {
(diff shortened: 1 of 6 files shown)


packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -2041,11 +2044,20 @@ export const DEFAULT_GAS_SCHEDULE: Required<FeeSchedule> = {
...
+            gas += uniFillData.gasUsed;
+        } else {
+            // Base gas usage estimate on all unique paths sorted in descending order by amount
+            const sortedPathAmounts = uniFillData.pathAmounts.sort((a, b) => b.inputAmount.comparedTo(a.inputAmount));
+            const uniqPathAmounts = _.uniqBy(sortedPathAmounts, p => p.uniswapPath);
+            const totalGasUsed = uniqPathAmounts.reduce((memo, p) => memo + p.gasUsed, 0);
+            gas += totalGasUsed;
         }
+
         return gas;
```
kimpers, line 2061: I think this is fine 🤔 Any thoughts?

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -2041,11 +2044,20 @@ export const DEFAULT_GAS_SCHEDULE: Required<FeeSchedule> = {
...
+        // underestimated gas usage
+        let gas = 34e3; // 34k base
+        if (isFinalUniswapV3FillData(uniFillData)) {
+            gas += uniFillData.gasUsed;
+        } else {
+            // Base gas usage estimate on all unique paths sorted in descending order by amount
+            const sortedPathAmounts = uniFillData.pathAmounts.sort((a, b) => b.inputAmount.comparedTo(a.inputAmount));
+            const uniqPathAmounts = _.uniqBy(sortedPathAmounts, p => p.uniswapPath);
+            const totalGasUsed = uniqPathAmounts.reduce((memo, p) => memo + p.gasUsed, 0);
+            gas += totalGasUsed;
```
dekz, line 2058: I didn't fully expect an accumulator here, can you refresh my memory on this data layout?

My assumption was the following:
```
200 USDC->DAI
[
 { path: USDCDAI30bps, inputAmount: 10, gasUsed: 150000 },
 { path: USDCDAI30bps, inputAmount: 50, gasUsed: 150000 },
 { path: USDCDAI5bps, inputAmount: 100, gasUsed: 150000 },
 { path: USDCDAI5bps, inputAmount: 200, gasUsed: 300000 },
]
```

From my reading, if we sell all 200 to USDCDAI5bps, the gas cost should be 300,000. From the accumulator is it calculating 450,000?

kimpers: I had misunderstood how how this worked, didn't realize that feel level was encoded into the path, so I thougth different unique paths meant different intermediate hops 😳 Updated to not double count the costs

kimpers: I found one issue with this approach last week. If the sampler aborts early/runs out of gas then `gasUsed` would be `0`, which would lead to an invalid gas estimation. Maybe it's better to revert to the original method, and just bump the gas depending on how many ticks were crossed? 🤔

packages/asset-swapper/src/utils/market_operation_utils/types.ts
```diff
@@ -268,19 +268,29 @@ export interface HopInfo {
...
 }
 
 export interface KyberDmmFillData extends UniswapV2FillData {
     poolsPath: string[];
 }
 
-export interface FinalUniswapV3FillData extends Omit<UniswapV3FillData, 'uniswapPaths'> {
+export const isFinalUniswapV3FillData = (
+    data: UniswapV3FillData | FinalUniswapV3FillData,
+): data is FinalUniswapV3FillData => !!(data as FinalUniswapV3FillData).uniswapPath;
```
dorothy-zbornak, line 288: no closures being used so why not a regular function?

dorothy-zbornak (approved):
:clap: :clap:

dorothy-zbornak:
oh don't forget changelog

packages/asset-swapper/src/utils/market_operation_utils/types.ts
kimpers: No reason other than I like the shorter `=>` version 😂  I don't believe there is any performance difference

dorothy-zbornak: Don't think there is in this case but we lean towards normal/named functions where possible as a rule because they show up in stack traces.

kimpers: Ah, that makes sense! Ty 🙏

PR #401, opened 2022-01-21
fix: incorrect output scaling when input is less than desired amount

kimpers:
## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
@@ -275,16 +275,14 @@ function findRoutesAndCreateOptimalPath(
         }
 
         // TODO(kimpers): remove once we have solved the rounding/precision loss issues in the Rust router
-        const scaleOutput = (fillInput: BigNumber, output: BigNumber) =>
-            output
-                .dividedBy(fillInput)
-                .times(rustInputAdjusted)
-                .decimalPlaces(0, side === MarketOperation.Sell ? BigNumber.ROUND_FLOOR : BigNumber.ROUND_CEIL);
+        const maxSampledOutput = BigNumber.max(...routeSamples.map(s => s.output));
+        const scaleOutput = (output: BigNumber) => BigNumber.min(output.times(scale), maxSampledOutput);
+
         adjustedFills.push({
             ...fill,
             input: rustInputAdjusted,
-            output: scaleOutput(fill.input, fill.output),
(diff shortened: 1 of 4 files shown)


packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -275,16 +275,14 @@ function findRoutesAndCreateOptimalPath(
         }
 
         // TODO(kimpers): remove once we have solved the rounding/precision loss issues in the Rust router
-        const scaleOutput = (fillInput: BigNumber, output: BigNumber) =>
-            output
-                .dividedBy(fillInput)
-                .times(rustInputAdjusted)
-                .decimalPlaces(0, side === MarketOperation.Sell ? BigNumber.ROUND_FLOOR : BigNumber.ROUND_CEIL);
+        const maxSampledOutput = BigNumber.max(...routeSamples.map(s => s.output));
+        const scaleOutput = (output: BigNumber) => BigNumber.min(output.times(scale), maxSampledOutput);
```
dorothy-zbornak, line 279: Wouldn't we want to min against the sum of outputs? We can still use multiple sources/samples even if we don't hit our Target, right?

dorothy-zbornak (commented):
When does this come up? Don't we abort with insufficient liquidity if we can't satisfy the input amount?

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
kimpers: This is all a mess and if it weren't for the f64 precision issues we wouldn't need this duct tape.

I could be wrong but this scaling is on a per fill level, and therefore it should always be from a single source, within the sampled range of that source, correct?

I added this min to check that we don't scale the output outside of the sampled range, where the liquidity is undefined.

kimpers:
> When does this come up? Don't we abort with insufficient liquidity if we can't satisfy the input amount?

This was observed on Optimism, due to sampling running out of gas because it had to cross too many ticks on Uni V3 (only source at this time).

As it stands today we don't abort with insufficient liquidity in this scenario (but maybe we should). One of the assumptions in the router is that e.g for sells you can always sell more for the same output. So if we have the last sample that sells 10k DAI -> 2 ETH, it assumes that it can also sell 20k DAI -> 2 ETH and will therefore try to Route in the best possible way with the samples it received.

dorothy-zbornak (approved):

PR #403, opened 2022-01-24
chore: Use global sample values for reduced payloads

dekz:
## Description

[Initial Simbot](https://metabase.spaceship.0x.org/dashboard/186?run_id=jacob-global-encoding-1&adjusted_amount_bought_win_tolerance__bps_=1.0)

For reference, our Sampler bytecode itself is ~85kb.

DAI/USDC was chosen due to some of our sources are only enabled when the pools match a token pair (i.e some Curves). 
ZRX/UNI was chosen due to maximising the amount of intermediate hops.

| test case       | request size (kb) | gas used    | abi encode (ms) |
| --------------- | ----------------- | ----------- | --------------- |
| Before USDC/DAI | 423.996           | 209,835,965 | 16.98           |
| After USDC/DAI  | 308.552           | 210,212,511 | 11.98           |
| Before ZRX/UNI  | 546.492           | 124,618,456 | 18.04           |
| After ZRX/UNI   | 408.584           | 128,185,347  | 12.73           |

Request payload size reduction of ~25%

Gas minor increase. Note: We perform a few extra SLOADs and SSTOREs so this should cost more. Calldata gas is not accounted for.

ABI encoding timing reduced by ~30%. This was not exactly precise testing.

### ABI Encoding timing
```
Runs: 10000
    UniswapV2 ABI
      13 input encode
        ✓ ZeroEx - optimized (2649ms)
p25: 0.191375ms, p50: 0.196625ms, p99: 2.276709ms, p100: 12.7295ms

        ✓ ZeroEx - no optimize (569ms)
p25: 0.045791ms, p50: 0.046458ms, p99: 0.061125ms, p100: 3.882292ms

        ✓ rust (363ms)
p25: 0.029666ms, p50: 0.029834ms, p99: 0.041666ms, p100: 1.869208ms

        ✓ rust - no amounts (160ms)
p25: 0.012084ms, p50: 0.012292ms, p99: 0.014292ms, p100: 1.98975ms
```

So since we're not messing around with amounts we may see a larger decrease in ABI encode time.

TODO:
* [ ] Buys 
* [x] Confirm nothing is broken (why is gasUsed lower than expected)
* [x] Remaining liquidity source (not as impactful as Uniswap(s) and Curves)
* [x] Remove old way? (of specifying the amounts explicitly)

## Checklist:


packages/asset-swapper/contracts/src/SamplerBase.sol
```diff
@@ -0,0 +1,41 @@
...
+    /// @dev Stored Sample values for each Sampler to pull
+    uint256[] private SAMPLE_VALUES = new uint256[](0);
+
+    function getSampleValues()
+        public
+        view
+        returns (uint256[] memory sampleValues)
+    {
+        sampleValues = SAMPLE_VALUES;
+    }
```
dorothy-zbornak, line 34: what if we just did this, have all the sampler contracts inherit `SamplerBase`, and simply access `SAMPLE_VALUES` directly so we can reduce gas and bytecode further.

```suggestion
    uint256[] internal SAMPLE_VALUES;
```

dorothy-zbornak: And also mark all the non-global `sample*From*()` functions private/internal.

dorothy-zbornak (approved):
amazing, thanks for making this a reality!

packages/asset-swapper/contracts/src/SamplerBase.sol
dekz: My intuition says we still need to keep some of these public for the Approximate buys, but I haven't got there yet.

TalCrypto:
ff

elenadimitrova:
Closing. Do not delete branch.

PR #410, opened 2022-01-31
Fix ABI encoding error with two hop buys

dorothy-zbornak:
## Description

Multihop orders use `uint256(-1)` as intermediate amounts. We should not be applying slippage to those values in the consumer.

## Checklist:


kimpers (approved):
LGTM

PR #412, opened 2022-02-03
fix: slippage inconsistency when recalculated in exchange proxy quote consumer

dekz:
## Description

We had a base line amount of errors where there was inconsistent slippage between the Orders and what is recalculated in `getMaxQuoteSlippageRate`. To keep things consistent I think we just pass it around in the `worstCaseQuoteInfo` which has already been adjusted by that amount of slippage.

For example, the order would be adjusted by 0.005 slippage and the calculated slippage would be 0.005011....

This could create an inconsistency with the `sellAmount` encoded and the FQT data encoded

```
{
  "functionName": "IncompleteFillSellQuoteError",
  "functionSignature": "adc35ca6",
  "functionArguments": [
    "IncompleteFillSellQuoteError({\n  sellToken: '0x2791bca1f2de4661ed88a30c99a7a9449aa84174',\n  soldAmount: 29303021,\n  sellAmount: 29303022\n})"
  ]
}
```

## Checklist:

### DIFF
--- a/packages/asset-swapper/src/quote_consumers/exchange_proxy_swap_quote_consumer.ts
+++ b/packages/asset-swapper/src/quote_consumers/exchange_proxy_swap_quote_consumer.ts
@@ -691,7 +691,7 @@ export class ExchangeProxySwapQuoteConsumer implements SwapQuoteConsumerBase {
 
 function slipNonNativeOrders(quote: MarketSellSwapQuote | MarketBuySwapQuote): OptimizedMarketOrder[] {
     const slippage = getMaxQuoteSlippageRate(quote);
-    if (!slippage) {
+    if (slippage === 0) {
         return quote.orders;
     }
     return quote.orders.map(o => {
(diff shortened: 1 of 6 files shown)


packages/asset-swapper/src/quote_consumers/exchange_proxy_swap_quote_consumer.ts
```diff
@@ -691,7 +691,7 @@ export class ExchangeProxySwapQuoteConsumer implements SwapQuoteConsumerBase {
 
 function slipNonNativeOrders(quote: MarketSellSwapQuote | MarketBuySwapQuote): OptimizedMarketOrder[] {
     const slippage = getMaxQuoteSlippageRate(quote);
-    if (!slippage) {
```
dorothy-zbornak, line 694: What if we just made this `slippage < 1e-4`?

dekz: @kimpers was able to reproduce this with 0.05 slippage parameter. Original error was reported with 0.005 slippage.

dorothy-zbornak (approved):

PR #423, opened 2022-02-14
Feat/mobius

dextracker:
## Description

#423 add support to the celo api for the following pools:

usdc_optics_v2
dai_optics_v2
weth_optics_v2
pusdc_optics_v2
usdc_allbridge_solana
usdc_poly_optics

## Mobius pools

Mobius pools are [here](https://github.com/mobiusAMM/mobiusV1). Not sure how updated it gets but this has all of the pools on the mobius UI

## Checklist:


kimpers (approved):
LGTM, if you rebase the latest development CI should start working again 👍

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -1029,6 +1070,15 @@ const createCurveFactoryCryptoExchangePool = (info: { tokens: string[]; pool: st
     poolAddress: info.pool,
     gasSchedule: info.gasSchedule,
 });
+const MOBIUSMONEY_CELO_SHARED: CurveInfo = {
+    exchangeFunctionSelector: CurveFunctionSelectors.swap,
+    sellQuoteFunctionSelector: CurveFunctionSelectors.calculateSwap,
+    buyQuoteFunctionSelector: CurveFunctionSelectors.None,
+    metaTokens: undefined,
+    gasSchedule: 100e3,
+    poolAddress: '',
```
dekz, line 1079: I'd maybe consider defaulting here to `NULL_ADDRESS`

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -1696,6 +1746,79 @@ export const FIREBIRDONESWAP_POLYGON_INFOS: { [name: string]: CurveInfo } = {
     },
 };
 
+export const MOBIUSMONEY_CELO_INFOS: { [name: string]: CurveInfo } = {
```
dekz, line 1749: If you have a link on where you found these pools I'd suggest commenting it above so we know where to go if we get wind of a new pool.

dekz (approved):
Nice work, looks good.

I'd recommend a simbot on it with the new tokens involved since it is a Curve like source. That is, it will give us higher confidence that the indexes of the tokens are correct. This is less important on sources where the token addresses are used to query since there's less that can go wrong.

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
dextracker: yup ill link it in the description

PR #424, opened 2022-02-15
feat: Improve Uniswap V3 gas schedule redux

kimpers:
## Checklist:

### DIFF
--- a/packages/asset-swapper/contracts/src/UniswapV3Sampler.sol
+++ b/packages/asset-swapper/contracts/src/UniswapV3Sampler.sol
@@ -77,7 +77,7 @@ interface IUniswapV3Pool {
 contract UniswapV3Sampler
 {
     /// @dev Gas limit for UniswapV3 calls. This is 100% a guess.
-    uint256 constant private QUOTE_GAS = 600e3;
+    uint256 constant private QUOTE_GAS = 900e3;
 
     /// @dev Sample sell quotes from UniswapV3.
     /// @param quoter UniswapV3 Quoter contract.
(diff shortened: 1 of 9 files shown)


packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -2349,11 +2352,33 @@ export const DEFAULT_GAS_SCHEDULE: Required<FeeSchedule> = {
...
-            gas += (path.length - 2) * 32e3; // +32k for each hop.
+        const uniFillData = fillData as UniswapV3FillData | FinalUniswapV3FillData;
+        // NOTE: This base value was heuristically chosen by looking at how much it generally
+        // underestimated gas usage
+        const base = 34e3; // 34k base
+        let gas = base;
+        if (isFinalUniswapV3FillData(uniFillData)) {
+            gas += uniFillData.gasUsed;
+        } else {
+            // NOTE: We don't actually know which of the paths would be used in the router
```
dorothy-zbornak, line 2363: What if we just used the median? Maybe even quicker since it's already (probably) sorted?

packages/asset-swapper/contracts/src/UniswapV3Sampler.sol
```diff
@@ -143,21 +181,30 @@ contract UniswapV3Sampler
...
-                        topUniswapPath = _toUniswapPath(path, poolPaths[j]);
+                        uniswapPaths[i] = _toUniswapPath(path, poolPaths[j]);
+                        uniswapGasUsed[i] = gasUsed;
                     }
                 } catch {}
             }
             // Break early if we can't complete the buys.
             if (topSellAmount == 0) {
+                // HACK(kimpers): To avoid too many local variables, paths and gas used is set directly in the loop
+                // then reset if no valid valid quote was found
```
dorothy-zbornak, line 202:
```suggestion
                // then reset if no valid quote was found
```

dorothy-zbornak (approved):

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
kimpers: Great idea! We probably still need to filter out 0 values because we can end up with a bunch of trailing 0s, but inside `pathAmountsWithGasUsed` they are roughly sorted (minor price variations) 🙏

PR #425, opened 2022-02-16
fix: start native orders "samples" at 0 & don't skip neg adj orders

kimpers:
## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/market_operation_utils/fills.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/fills.ts
@@ -96,6 +96,7 @@ export function nativeOrdersToFills(
     outputAmountPerEth: BigNumber,
     inputAmountPerEth: BigNumber,
     fees: FeeSchedule,
+    filterNegativeAdjustedRateOrders: boolean = true,
 ): Fill[] {
     const sourcePathId = hexUtils.random();
     // Create a single path from all orders.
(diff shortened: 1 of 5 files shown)


packages/asset-swapper/src/utils/market_operation_utils/fills.ts
```diff
@@ -126,8 +127,8 @@ export function nativeOrdersToFills(
             side === MarketOperation.Sell ? clippedOutput.minus(outputPenalty) : clippedOutput.plus(outputPenalty);
         const adjustedRate =
             side === MarketOperation.Sell ? adjustedOutput.div(clippedInput) : clippedInput.div(adjustedOutput);
-        // Skip orders with rates that are <= 0.
-        if (adjustedRate.lte(0)) {
+        // Optionally skip orders with rates that are <= 0.
+        if (filterNegativeAdjustedRateOrders && adjustedRate.lte(0)) {
```
kimpers, line 131: If the rust router chose a native order we want to use it, even if it has negative adjusted rate. In this case it was the best option from all the bad options 🙈

mzhu25: do we ever _need_ to skip over orders with rates <= 0?

mzhu25: is the assumption here that skipping them saves us some compute time?

mzhu25: 😫 pain

mzhu25 (approved):

PR #430, opened 2022-02-23
fix: native order scaling & don't try to route 1 wei trades (it will fail)

kimpers:
## Checklist:

### DIFF
--- a/packages/asset-swapper/package.json
+++ b/packages/asset-swapper/package.json
@@ -66,7 +66,7 @@
         "@0x/contracts-zero-ex": "^0.31.0",
         "@0x/dev-utils": "^4.2.11",
         "@0x/json-schemas": "^6.4.1",
-        "@0x/neon-router": "0xProject/gitpkg-registry#0x-neon-router-v0.3.2-commit.4b8eda4-4b8eda4",
+        "@0x/neon-router": "^0.3.3",
         "@0x/protocol-utils": "^1.11.0",
         "@0x/quote-server": "^6.0.6",
         "@0x/types": "^3.3.4",
--- a/packages/asset-swapper/package.json
+++ b/packages/asset-swapper/package.json
@@ -66,7 +66,7 @@
         "@0x/contracts-zero-ex": "^0.31.0",
         "@0x/dev-utils": "^4.2.11",
         "@0x/json-schemas": "^6.4.1",
-        "@0x/neon-router": "^0.3.2",
+        "@0x/neon-router": "0xProject/gitpkg-registry#0x-neon-router-v0.3.2-commit.4b8eda4-4b8eda4",
         "@0x/protocol-utils": "^1.11.0",
         "@0x/quote-server": "^6.0.6",
         "@0x/types": "^3.3.4",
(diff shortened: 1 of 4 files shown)


packages/asset-swapper/package.json
```diff
@@ -66,7 +66,7 @@
         "@0x/contracts-zero-ex": "^0.31.0",
         "@0x/dev-utils": "^4.2.11",
         "@0x/json-schemas": "^6.4.1",
-        "@0x/neon-router": "^0.3.2",
+        "@0x/neon-router": "0xProject/gitpkg-registry#0x-neon-router-v0.3.2-commit.4b8eda4-4b8eda4",
```
kimpers, line 69: Will update this with the proper version once the code has been reviewed, and I have made a proper publish

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -75,6 +77,13 @@ function findRoutesAndCreateOptimalPath(
     fees: FeeSchedule,
     neonRouterNumSamples: number,
 ): Path | undefined {
+    // Currently the rust router is unable to handle 1 base unit sized quotes and will error out
+    // To avoid flooding the logs with these errors we just return an insufficient liquidity error
+    // which is how the JS router handles these quotes today
+    if (input.eq(ONE_BASE_UNIT)) {
```
dorothy-zbornak, line 83: But it can handle, like, 2?

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -155,12 +164,24 @@ function findRoutesAndCreateOptimalPath(
...
+        for (let i = 1; i <= 13; i++) {
+            const fraction = i / 13;
+            const currentInput = BigNumber.min(
+                normalizedOrderInput.times(scaleToInput).times(fraction),
+                normalizedOrderInput,
+            );
+            const currentOutput = BigNumber.min(
+                normalizedOrderOutput.times(scaleToInput).times(fraction),
+                normalizedOrderOutput,
+            );
```
dorothy-zbornak, line 184: the router can handle non-integers, yeah?

dorothy-zbornak (approved):

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
kimpers: yes, strangely enough 😂  

It does a sanity assertion that for every sample index i, `samples[i] <= samples[i + 1]`, which should work as long as the sampled amounts are equal or increasing. But when all samples are the same value (eg `1`) this assertion fails. I expect it's due to some f64 precision shenanigans but I haven't had time to investigate 🤷‍♂️

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
kimpers: All values are rounded to integers below, but technically the router should not care wether it's an integer or not 👍

PR #433, opened 2022-02-28
fix: routing glue optimization

shawnxin89:
## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/market_operation_utils/path.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/path.ts
@@ -122,17 +122,8 @@ export class Path {
                 ++i;
                 continue;
             }
-            // If there are contiguous bridge orders, we can batch them together.
-            // TODO jacob pretty sure this is from DFB and we can remove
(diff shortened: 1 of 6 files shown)


packages/asset-swapper/src/utils/market_operation_utils/path.ts
```diff
@@ -122,17 +122,8 @@ export class Path {
...
-            const contiguousBridgeFills = [collapsedFills[i]];
-            for (let j = i + 1; j < collapsedFills.length; ++j) {
-                if (collapsedFills[j].source === ERC20BridgeSource.Native) {
-                    break;
-                }
-                contiguousBridgeFills.push(collapsedFills[j]);
-            }
 
-            this.orders.push(createBridgeOrder(contiguousBridgeFills[0], makerToken, takerToken, opts.side));
+            this.orders.push(createBridgeOrder(collapsedFills[i], makerToken, takerToken, opts.side));
```
shawnxin89, line 126: Seems only the first element in `contiguousBridgeFills` is used

PR #437, opened 2022-03-01
feat: BTRFLY/WETH factory pool

dextracker:
## Checklist:


packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -1270,6 +1281,11 @@ export const CURVE_MAINNET_INFOS: { [name: string]: CurveInfo } = {
         pool: CURVE_POOLS.ibEURsEUR,
         gasSchedule: 176e3,
     }),
+    [CURVE_POOLS.btrflyweth]: createCurveFactoryCryptoExchangePool({
+        tokens: [MAINNET_TOKENS.WETH, MAINNET_TOKENS.BTRFLY],
+        pool: CURVE_POOLS.btrflyweth,
+        gasSchedule: 411e3,
```
dekz, line 1287: This seems a bit high, is it from the output of Simbot?

FWIW this value should be from the raw cost of `POOL.exchange_underlying`. So essentially it is the cost of switching from SourceA to SourceB. That is, it assumes we're already in a transaction (-21,000) and that we're already inside the FillQuoteTransformer (-150,000). From that perspective, what is the cost Between Curve or Uniswap. Simbot will report the entire tx gas.

I'd suggest the gas cost may be closer to `250,000` 
image

dekz (approved):
Looks good to me, only a slight comment on the `gasSchedule`

PR #440, opened 2022-03-04
feat: VIP routing in the router and don't create fallback orders for quotes with native

kimpers:
## Checklist:


packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -94,6 +95,126 @@ function findRoutesAndCreateOptimalPath(
         return fills[0];
     };
 
+    const createPathFromStrategy = (sourcesRustRoute: Float64Array, sourcesOutputAmounts: Float64Array) => {
```
kimpers, line 98: all of this will need to be rewritten when we rip out the JS router and refactor the fills/path constructs so it's probably fine to leave this a bit hacky solution here for now 🤔

packages/asset-swapper/package.json
```diff
@@ -66,7 +66,7 @@
         "@0x/contracts-zero-ex": "^0.31.1",
         "@0x/dev-utils": "^4.2.11",
         "@0x/json-schemas": "^6.4.1",
-        "@0x/neon-router": "^0.3.3",
+        "@0x/neon-router": "0xProject/gitpkg-registry#0x-neon-router-v0.3.4-commit.a21133f-6a426f3",
```
kimpers, line 69: **TODO:** update this with official release when the code has been reviewed

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -573,7 +572,10 @@ export class MarketOperationUtils {
         }
 
         // Generate a fallback path if required
-        await this._addOptionalFallbackAsync(side, inputAmount, optimalPath, dexQuotes, fills, opts, penaltyOpts);
+        // TODO(kimpers): Will experiment with disabling this and see how it affects revert rate
+        // to avoid yet another router roundtrip
+        // TODO: clean this up if we don't need it
+        // await this._addOptionalFallbackAsync(side, inputAmount, optimalPath, dexQuotes, fills, opts, penaltyOpts);
```
kimpers, line 578: We're gonna try skipping fallback orders for RFQT, as it seems like it would only marginally increase the revert rate, but decrease tx sticker shock for users, and remove one extra routing call for routes that include Native

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
kimpers, line 98: This was basically just moved up and wrapped into a function so that we can call it twice (with access to closure variables), first for all sources, then for VIP sources

All of this will need to be rewritten when we rip out the JS router and refactor the fills/path constructs so it's probably fine to leave this a bit hacky solution here for now 🤔

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -128,6 +249,7 @@ function findRoutesAndCreateOptimalPath(
                         .integerValue()
                         .toNumber(),
                 );
+                memo.isVip = vipSourcesSet.has(sample.source);
```
dorothy-zbornak, line 252: This should be the same for all samples in the path right? Maybe move it to the last arg to `reduce()` with optional chaining on element `0`?

dorothy-zbornak (approved):

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
kimpers: Nice optimization 🙏

PR #443, opened 2022-03-10
Feat/add router test data sampler

shawnxin89:
## Description

Add the option to sample test data for Rust router. The test data will be uploaded to AWS S3 when it hits 1000 samples. 

Will need to setup environment variables 

## Checklist:


packages/asset-swapper/src/utils/market_operation_utils/router_test_data_sampler.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public sampleRoute(route: OptimizerCapture): void {
+        if (Math.random() < SAMPLE_THRESHOLD) {
+            this._routes.push(route);
+
+            if (this._routes.length > UPLOAD_SIZE) {
+                const toUpload = this._routes;
+                this._routes = [];
+                const file = `${this._chainId}-${Date.now()}`;
```
kimpers, line 28: This would be the folder path, not the file right? If so the timestamp should go into the key otherwise we'll end up creating thousands of buckets if I'm not mistaken 🤔 

I would create a bucket like `0x-router-test-data` and this would be the bucket.

Then the key would be something like`${this._chainId}/sample-${Date.now()}.json`, that should put them into "subfolder" in s3 IIRC

packages/asset-swapper/src/utils/market_operation_utils/router_test_data_sampler.ts
```diff
@@ -0,0 +1,57 @@
...
+                const s3Client = new S3Client({ region: REGION });
+                // Set the parameters
+                const params = {
+                    Bucket: file, // The name of the bucket.
+                    Key: `${file}-sample.txt`, // The name of the object.
```
kimpers, line 33:
```suggestion
                    Key: `${file}-sample.json`, // The name of the object.
```

packages/asset-swapper/src/utils/market_operation_utils/router_test_data_sampler.ts
```diff
@@ -0,0 +1,57 @@
...
+                    Body: JSON.stringify(toUpload), // The content of the object.
+                };
+
+                const run = async () => {
+                    // Create an Amazon S3 bucket.
+                    await s3Client.send(new CreateBucketCommand({ Bucket: params.Bucket }));
```
kimpers, line 39: This should not be needed, we create the bucket once (which we can do from the web UI) and the reuse it.

packages/asset-swapper/src/utils/market_operation_utils/router_test_data_sampler.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    public sampleRoute(route: OptimizerCapture): void {
+        if (Math.random() < SAMPLE_THRESHOLD) {
+            this._routes.push(route);
+
+            if (this._routes.length > UPLOAD_SIZE) {
+                const toUpload = this._routes;
+                this._routes = [];
+                const file = `${this._chainId}-${Date.now()}`;
+                const s3Client = new S3Client({ region: REGION });
```
kimpers, line 29: NIT: This could be done outside the class since we're not depending on any arguments

packages/asset-swapper/src/utils/market_operation_utils/router_test_data_sampler.ts
```diff
@@ -0,0 +1,57 @@
...
+                    Bucket: file, // The name of the bucket.
+                    Key: `${file}-sample.txt`, // The name of the object.
+                    Body: JSON.stringify(toUpload), // The content of the object.
+                };
+
+                const run = async () => {
+                    // Create an Amazon S3 bucket.
+                    await s3Client.send(new CreateBucketCommand({ Bucket: params.Bucket }));
+                    // Create an object and upload it to the Amazon S3 bucket.
+                    await s3Client.send(new PutObjectCommand(params));
```
kimpers, line 41: No need to await this since we should only have one async call
```suggestion
                    s3Client.send(new PutObjectCommand(params)).catch(err => {
                    DEFAULT_WARNING_LOGGER(err, "Failed to upload routing test sample to s3");
                    })
```

packages/asset-swapper/src/utils/market_operation_utils/router_test_data_sampler.ts
```diff
@@ -0,0 +1,57 @@
...
+                };
+                run()
+                    .then()
+                    .catch();
```
kimpers, line 45: I think this `run` fn can be removed entirely if we don't call the create bucket method and just catch errors from upload

packages/asset-swapper/src/utils/market_operation_utils/router_test_data_sampler.ts
kimpers, line 41: I could be wrong here but I think you need to use `.promise()` in order for this to actually be a promise. Otherwise, I think it's expecting a callback function argument.

It's been a while since I used S3 tho so the behaviour could have changed

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -340,6 +342,12 @@ function findRoutesAndCreateOptimalPath(
     const vipSourcesRustRoute = new Float64Array(rustArgs.pathsIn.length);
     const vipSourcesOutputAmounts = new Float64Array(rustArgs.pathsIn.length);
 
+    // Get rustArgs samples for test purpose
+    if (SHOULD_ENABLE_RUST_DATA_SAMPLER) {
```
kimpers, line 346: I know it was like this in the initial spec, but I wonder if it's better to do the `Math.random() < SAMPLE_THRESHOLD` check here instead, saves us some execution for the vast majority of the cases where we do nothing
```suggestion
    if (SHOULD_ENABLE_RUST_DATA_SAMPLER && Math.random() < SAMPLE_THRESHOLD) {
```

kimpers (changes requested):
Nice work on this. Gonna be awesome to start collecting some test cases! Feel free to ignore the nits

packages/asset-swapper/src/utils/market_operation_utils/router_test_data_sampler.ts
```diff
@@ -0,0 +1,57 @@
+import { ChainId } from '@0x/contract-addresses';
+import { OptimizerCapture } from '@0x/neon-router';
+import { CreateBucketCommand, PutObjectCommand, S3Client } from '@aws-sdk/client-s3';
+
+const SAMPLE_THRESHOLD = 0.05;
```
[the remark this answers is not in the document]
shawnxin89: Good point 👍 It seems we have ~600 requests/min on mainnet. Something like 0.003 will give us ~2500 tests

packages/asset-swapper/src/utils/market_operation_utils/router_test_data_sampler.ts
```diff
@@ -0,0 +1,57 @@
...
+const UPLOAD_SIZE = 1000;
```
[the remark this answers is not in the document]
shawnxin89: I will try 500

packages/asset-swapper/src/utils/market_operation_utils/router_test_data_sampler.ts
shawnxin89: Yeah the previous approach generates a lot of buckets. I kinda forgot the number of buckets allowed is limited. Good pick 👍

packages/asset-swapper/src/utils/market_operation_utils/router_test_data_sampler.ts
shawnxin89: The return of `s3Client.send` is a promise per the docs

kimpers: awesome then they've finally moved away from callbacks as the default behavior 🎉

PR #445, opened 2022-03-14
Draft. PoC pseudo code showing general idea for resuing SOR path crea…

johngrantuk:
This is a draft PR for feedback only.
Proposal for adding Balancer Linear pool support by reusing SOR path creation logic and adding multihop support.
See more detail [here](https://www.notion.so/Linear-Pools-0x-df6772d202e848989594f882e61882e4)

## Checklist:

### DIFF
--- a/packages/asset-swapper/contracts/src/BalancerV2Sampler.sol
+++ b/packages/asset-swapper/contracts/src/BalancerV2Sampler.sol
@@ -109,6 +109,47 @@ contract BalancerV2Sampler is SamplerUtils {
         }
     }
 
+    // TODO - THIS IS PSEUDO CODE AS A DEMO
+    // Replaces amount for first step with each takerTokenAmount and calls queryBatchSwap using supplied steps
+    /// @dev Sample sell quotes from Balancer V2 supporting multihops.
+    /// @param swapSteps Array of swap steps (can be >= 1).
+    /// @param swapAssets Array of token address for swaps.
+    /// @param takerTokenAmounts Taker token sell amount for each sample.
+    function sampleMultihopSellsFromBalancerV2(
+        IBalancerV2Vault.BatchSwapStep[] memory swapSteps,
+        IAsset[] memory swapAssets,
+        uint256[] memory takerTokenAmounts
+    )
+        public
+        returns (uint256[] memory makerTokenAmounts)
+    {
+        IBalancerV2Vault vault = IBalancerV2Vault(poolInfo.vault);
+        uint256 numSamples = takerTokenAmounts.length;
+        makerTokenAmounts = new uint256[](numSamples);
+        IBalancerV2Vault.FundManagement memory swapFunds =
+            _createSwapFunds();
+
+        for (uint256 i = 0; i < numSamples; i++) {
+            swapSteps[0].amount = takerTokenAmounts[i];
+            try
+                // For sells we specify the takerToken which is what the vault will receive from the trade
+                vault.queryBatchSwap(IBalancerV2Vault.SwapKind.GIVEN_IN, swapSteps, swapAssets, swapFunds)
+            // amounts represent pool balance deltas from the swap (incoming balance, outgoing balance)
+            returns (int256[] memory amounts) {
+                // Outgoing balance is negative so we need to flip the sign
+                // TO DO - Add index for asset out
+                int256 amountOutFromPool = amounts[1] * -1;
+                if (amountOutFromPool <= 0) {
+                    break;
+                }
+                makerTokenAmounts[i] = uint256(amountOutFromPool);
+            } catch (bytes memory) {
+                // Swallow failures, leaving all results as zero.
+                break;
+            }
+        }
+    }
+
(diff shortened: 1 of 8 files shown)


packages/asset-swapper/src/utils/market_operation_utils/pools_cache/balancer_v2_utils_new.ts
```diff
@@ -0,0 +1,134 @@
...
+        // Helper function exposed from SDK/SOR that formats and returns swapSteps and assets
+        return formatSwaps(paths);
+    }
+
+    protected async _loadTopPoolsAsync(): Promise<void> {
+        const fromToSwapInfo: {
+            [from: string]: { [to: string]: BalancerSwapInfo[] };
+        } = {};
+
+        const pools = this.poolDataService.fetchPools();
```
dorothy-zbornak, line 93: Does this make any blocking external requests?

johngrantuk: It would only be a subgraph query which I believe should be the same as the current `_fetchTopPoolsAsync()`.

johngrantuk: Apologies, I've missed out the `await` there. It will be `const pools = await this.poolDataService.fetchPools();`

packages/asset-swapper/contracts/src/BalancerV2Sampler.sol
```diff
@@ -109,6 +109,47 @@ contract BalancerV2Sampler is SamplerUtils {
...
+
+        for (uint256 i = 0; i < numSamples; i++) {
+            swapSteps[0].amount = takerTokenAmounts[i];
+            try
+                // For sells we specify the takerToken which is what the vault will receive from the trade
+                vault.queryBatchSwap(IBalancerV2Vault.SwapKind.GIVEN_IN, swapSteps, swapAssets, swapFunds)
+            // amounts represent pool balance deltas from the swap (incoming balance, outgoing balance)
+            returns (int256[] memory amounts) {
+                // Outgoing balance is negative so we need to flip the sign
+                // TO DO - Add index for asset out
```
dorothy-zbornak, line 140: What do you mean by this?

packages/asset-swapper/contracts/src/BalancerV2Sampler.sol
```diff
@@ -109,6 +109,47 @@ contract BalancerV2Sampler is SamplerUtils {
         }
     }
 
+    // TODO - THIS IS PSEUDO CODE AS A DEMO
+    // Replaces amount for first step with each takerTokenAmount and calls queryBatchSwap using supplied steps
+    /// @dev Sample sell quotes from Balancer V2 supporting multihops.
+    /// @param swapSteps Array of swap steps (can be >= 1).
+    /// @param swapAssets Array of token address for swaps.
+    /// @param takerTokenAmounts Taker token sell amount for each sample.
+    function sampleMultihopSellsFromBalancerV2(
```
dorothy-zbornak, line 118: Is there a way to do this for buys?

packages/asset-swapper/contracts/src/BalancerV2Sampler.sol
johngrantuk: I may be reading it wrong but previously queryBatchSwap was only returning two deltas - for tokenIn and tokenOut, because it was always a single swap and there were only two tokens involved.
For longer paths queryBatchSwap will return a delta for each token in the assets array and we want the delta for the last index which should be tokenOut.

packages/asset-swapper/src/utils/market_operation_utils/pools_cache/balancer_v2_utils_new.ts
```diff
@@ -0,0 +1,212 @@
...
+        private readonly subgraphUrl: string = BALANCER_V2_SUBGRAPH_URL_BY_CHAIN[chainId],
+        private readonly maxPoolsFetched: number = BALANCER_MAX_POOLS_FETCHED,
+        private readonly _topPoolsFetched: number = BALANCER_TOP_POOLS_FETCHED,
+        private readonly _warningLogger: LogFunction = DEFAULT_WARNING_LOGGER,
+        cache: { [key: string]: CacheValue } = {},
+    ) {
+        super(cache);
+        const config: BalancerSdkConfig = {
+            network: Network[ChainId[chainId]],
+            rpcUrl: `https://mainnet.infura.io/v3/${process.env.INFURA}`,
```
dorothy-zbornak, line 52: Is there a way to pass in a generic provider object? I.e., something that just implements an interface like `{ send(payload); }`

johngrantuk: Unfortunately there's not at this time but in this case the provider isn't actually used within the SDK so you could just do: `rpcUrl: '',`.

dekz:
Closing as it's been merged in #462

PR #449, opened 2022-03-16
ugprade tools deps and regenerate wrappers

dorothy-zbornak:
## Checklist:


package.json
```diff
@@ -75,6 +75,7 @@
         "wsrun": "^5.2.4"
     },
     "resolutions": {
-        "merkle-patricia-tree": "3.0.0"
+        "merkle-patricia-tree": "3.0.0",
+        "**/bignumber.js": "^9.0.2"
```
dorothy-zbornak, line 79: only way I could get it to reconcile all the hoisted 9.0.1s :shrug:

kimpers: I think it's probably fine 🙈  I guess the biggest risk is that some third-party dep either depends on a much older version (or in the future a newer version) and breaks

kimpers (approved):
Thank you for fixing this 🙏 The bignumber version discrepancies have been a huge pita when yarn linking locally

kimpers:
Some changelogs have been updated but others have not, so don't forget to give that a check before merging :)

merklejerk:
> Some changelogs have been updated but others have not, so don't forget to give that a check before merging :)

The publish script should actually automatically create "dependencies updated" changelog entries for all the other packages, which exactly describes what I did anyway.

package.json
merklejerk: There actually already are some deps that hoist an older version but it doesn't seem to break anything.

PR #455, opened 2022-03-31
chore: bump ts version to solve type issue with ethereumjs

kimpers:
## Description

This fixes the [ethereumjs TypeScript issue](https://app.circleci.com/pipelines/github/0xProject/protocol/3290/workflows/44f17c3a-b513-4c7d-8d1b-d4e364b78c05/jobs/16440) blocking publishing by upgrading our TS version to 4.6

## Checklist:

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -13,7 +13,8 @@
         // These settings are required for TypeScript project references
         "declaration": true,
         "declarationMap": true,
-        "sourceMap": true
+        "sourceMap": true,
+        "useUnknownInCatchVariables": false
     },
     "exclude": [],
     // The root of the project is just a list of references and does not contain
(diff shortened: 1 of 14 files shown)


tsconfig.json
```diff
@@ -13,7 +13,8 @@
         // These settings are required for TypeScript project references
         "declaration": true,
         "declarationMap": true,
-        "sourceMap": true
+        "sourceMap": true,
+        "useUnknownInCatchVariables": false
```
kimpers, line 17: This is a new thing in [TS 4.4](https://www.typescriptlang.org/tsconfig#useUnknownInCatchVariables) where the value in a catch clause is assumed to be `unknown` and you have to assert that the type is actually `Error`. Since you can reject promises with any value in JS 🤦 But this causes a bunch of errors in our code and since we only throw Errors I think it's fine to turn this check off.

dekz (approved):

PR #466, opened 2022-04-27
Fix the filter for considered sources on indicative sells for Quote Report

Ktl-XV:
## Description

The quote report considered sources where incorrectly filtered out when doing a Sell. This PR fixes the bug

## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/quote_report_generator.ts
+++ b/packages/asset-swapper/src/utils/quote_report_generator.ts
@@ -207,7 +207,7 @@ export function generateExtendedQuoteReportSources(
         ..._.flatten(
             quotes.dexQuotes.map(dex =>
                 dex
-                    .filter(quote => isDexSampleForTotalAmount(quote, marketOperation, amount))
+                    .filter(quote => isDexSampleForTotalAmount(quote, amount))
                     .map(quote => dexSampleToReportSource(quote, marketOperation)),
             ),
         ),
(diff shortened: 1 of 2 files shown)


packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -952,7 +952,7 @@ export const DEFAULT_TOKEN_ADJACENCY_GRAPH_BY_CHAIN_ID = valueByChainId<TokenAdj
                 builder
                     .add(AVALANCHE_TOKENS.aWETH, AVALANCHE_TOKENS.nETH)
                     .add(AVALANCHE_TOKENS.nETH, AVALANCHE_TOKENS.aWETH);
-                //Trader Joe MAG/MIM pool
+                // Trader Joe MAG/MIM pool
```
dekz, line 955:
```suggestion
                //Trader Joe MAG/MIM pool
```
Just saw this is getting fixed inadvertently  another PR

packages/asset-swapper/src/utils/quote_report_generator.ts
```diff
@@ -306,16 +306,8 @@ export function dexSampleToReportSource(ds: DexSample, marketOperation: MarketOp
...
-    // based on the market operation
-    if (marketOperation === MarketOperation.Buy) {
-        return ds.input === amount;
-    } else if (marketOperation === MarketOperation.Sell) {
-        return ds.output === amount;
-    } else {
-        throw new Error(`Unexpected marketOperation ${marketOperation}`);
-    }
+function isDexSampleForTotalAmount(ds: DexSample, amount: BigNumber): boolean {
+    return ds.input === amount;
```
dekz, line 310: I think I'd suggest this rather than the `===` operator.
```suggestion
    return ds.input.eq(amount);
```

dekz (approved):
I'd also recommend updating the CHANGELOG.json file so we can find this a bit easier in the future if needed.

PR #467, opened 2022-04-28
feat: add IRfqClient

phil-ociraptor:
## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/irfq_client.ts
+++ b/packages/asset-swapper/src/utils/irfq_client.ts
@@ -1,0 +1,55 @@
+import { OtcOrder, RfqOrder, Signature } from '@0x/protocol-utils';
+import { BigNumber } from '@0x/utils';
+
+export interface RfqClientPriceRequest {
+    takerAddress: string;
+    txOrigin: string;
+    makerToken: string;
+    takerToken: string;
+    assetFillAmount: BigNumber;
+    marketOperation: 'sell' | 'buy';
+    comparisonPrice: BigNumber | undefined;
+    feeAmount: BigNumber | undefined;
+    feeToken: string | undefined;
+}
+
+export interface RfqClientQuoteRequest extends RfqClientPriceRequest {}
+
+export interface RfqClientIndicativeQuote {
+    maker: string;
+    makerUri: string;
+    makerToken: string;
+    takerToken: string;
+    makerAmount: BigNumber;
+    takerAmount: BigNumber;
+    expiry: BigNumber;
+}
+
+export interface RfqClientPriceResponse {
+    quotes: RfqClientIndicativeQuote[];
+}
+
+export interface RfqClientRfqOrderFirmQuote {
+    order: RfqOrder;
+    makerUri: string;
+    signature: Signature;
+    kind: 'rfq';
+}
+
+export interface RfqClientOtcOrderFirmQuote {
+    order: OtcOrder;
+    makerUri: string;
+    signature: Signature;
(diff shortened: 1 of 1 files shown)


packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -666,14 +667,30 @@ export class MarketOperationUtils {
...
+                                  makerToken,
+                                  takerToken,
+                                  assetFillAmount: amount,
+                                  marketOperation: side,
+                                  comparisonPrice: wholeOrderPrice,
+                                  feeAmount: undefined,
+                                  feeToken: undefined,
+                              })
+                          ).quotes as V4RFQIndicativeQuoteMM[])
+                        : await rfqt.quoteRequestor.requestRfqtIndicativeQuotesAsync(
```
rhinodavid, line 686: is there a way we could make quoterequestor conform to IRfqClient so we don't have this weird branching? Or is it not worth it?

packages/asset-swapper/src/utils/rfq_client_mappers.ts
```diff
@@ -0,0 +1,23 @@
+import { FillQuoteTransformerOrderType } from '@0x/protocol-utils';
```
rhinodavid, line 1: consider just putting this in the interface file -- will be easier to clean up later

rhinodavid (approved):

packages/asset-swapper/src/utils/market_operation_utils/index.ts
phil-ociraptor: I think it's better to have the IRfqClient conform to the existing code, otherwise, we need to change some stuff downstream. Riskier, for unclear benefit

packages/asset-swapper/src/utils/irfq_client.ts
```diff
@@ -1,13 +1,16 @@
 import { OtcOrder, RfqOrder, Signature } from '@0x/protocol-utils';
 import { BigNumber } from '@0x/utils';
 
+import { AltRfqMakerAssetOfferings } from '../types';
 export interface RfqClientPriceRequest {
+    integratorId: string;
```
rhinodavid, line 6: can you sort

packages/asset-swapper/src/utils/irfq_client.ts
```diff
@@ -0,0 +1,59 @@
+import { RfqOrder, Signature } from '@0x/protocol-utils';
+import { BigNumber } from '@0x/utils';
+
+import { AltRfqMakerAssetOfferings } from '../types';
+
+export interface RfqClientV1PriceRequest {
+    altRfqAssetOfferings: AltRfqMakerAssetOfferings | undefined;
+    assetFillAmount: BigNumber;
+    chainId: number;
```
kh-chang, line 9: would be beneficial to use enum `ChainId` here instead?

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -663,17 +666,49 @@ export class MarketOperationUtils {
             // Timing of RFQT lifecycle
             const timeStart = new Date().getTime();
             const { makerToken, takerToken } = nativeOrders[0].order;
```
kh-chang, line 668: [out of scope] it seems like `fooToken` sometimes mean the symbol(?) (e.g. `RfqClientV1PriceRequest` has both 
`takerAddress`, and `takerToken`) while here it's address.

PR #469, opened 2022-05-02
chore: Offboard Swerve Finance and LinkSwap [TKR-356]

kh-chang:
## Description

Offboard Swerve and LinkSwap as liquidity is dry and the projects are not maintained. 

## Testing instructions

Ran unit tests.

## Types of changes

* Cleanup

## Checklist:


dextracker:
LGTM! i dont think we will need to remove either of these from the bridge adapter as they are forks.

PR #471, opened 2022-05-06
Add BiSwap (as UniV2 clone) on BSC

idokleinman:
## Description

Added BiSwap on BSC

## Testing instructions

SimBot BSC

## Types of changes

New Feature:
Added BiSwap on BSC

## Checklist:

Link to Simbot run data:
https://metabase.spaceship.0x.org/dashboard/186?run_id=shock-flagrant-reading&adjusted_amount_bought_win_tolerance__bps_=0.5


dekz (approved):
Yeah this is looking pretty good to me with one small change on `BridgeProtocol`. 

Once you've got a simbot to confirm (it's been a while for me hah) then should be good to go.

packages/protocol-utils/src/transformer_utils.ts
```diff
@@ -135,6 +135,7 @@ export enum BridgeProtocol {
     AaveV2,
     Compound,
     BalancerV2Batch,
+    BiSwap,
```
dekz, line 138:
```suggestion
    BiSwap,
```
This isn't needed, as it has a 1:1 mapping to "custom bridge protocol" implementations inside of the [BridgeAdapter contract](https://github.com/0xProject/protocol/blob/development/contracts/zero-ex/contracts/src/transformers/bridges/BridgeProtocols.sol#L55). Since this is just a UniswapV2? clone, there is no need to update [BridgeAdapter](https://github.com/0xProject/protocol/blob/development/contracts/zero-ex/contracts/src/transformers/bridges/BridgeProtocols.sol#L55) and then no need to map it here.

dextracker: yeah i think this is the things causing asset-swapper to fail on build with balancerV2Batch. other than that lgtm with a simbot

PR #476, opened 2022-05-16
Add stETH wrap/unwrap support [TKR-377]

kh-chang:
## Description

Add stETH wrap/unwrap support

## Types of changes

## Checklist:

### DIFF
--- a/contracts/zero-ex/contracts/src/transformers/bridges/mixins/MixinLido.sol
+++ b/contracts/zero-ex/contracts/src/transformers/bridges/mixins/MixinLido.sol
@@ -115,8 +115,8 @@ contract MixinLido {
 
     ) private returns(uint256 boughtAmount){
         (IEtherTokenV06 stETH, IWstETH wstETH) = abi.decode(bridgeData, (IEtherTokenV06, IWstETH));
-        sellToken.approveIfBelow(address(wstETH), sellAmount);
         if (address(sellToken) == address(stETH) && address(buyToken) == address(wstETH) ) {
+            sellToken.approveIfBelow(address(wstETH), sellAmount);
             return wstETH.wrap(sellAmount);
         }
         if (address(sellToken) == address(wstETH) && address(buyToken) == address(stETH) ) {
--- a/contracts/zero-ex/contracts/src/transformers/bridges/mixins/MixinLido.sol
+++ b/contracts/zero-ex/contracts/src/transformers/bridges/mixins/MixinLido.sol
@@ -26,7 +26,7 @@ import "@0x/contracts-erc20/contracts/src/v06/IEtherTokenV06.sol";
 
 
 /// @dev Minimal interface for minting StETH
-interface ILido {
+interface IStETH {
     /// @dev Adds eth to the pool
     /// @param _referral optional address for referrals
     /// @return StETH Amount of shares generated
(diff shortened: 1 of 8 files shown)


contracts/zero-ex/contracts/src/transformers/bridges/mixins/MixinLido.sol
```diff
@@ -59,12 +86,43 @@ contract MixinLido {
...
+
+        return _tradeWstETH(sellToken, buyToken, sellAmount, bridgeData);
+    }
+
+    function _tradeStETH(
+        IERC20TokenV06 buyToken,
+        uint256 sellAmount,
+        bytes memory bridgeData
+    ) private returns (uint256 boughtAmount) {
+        (IStETH stETH) = abi.decode(bridgeData, (IStETH));
```
kh-chang, line 101: @dekz I tried to make the contract backward compatible. 

With the change in PR `bridgeData` will always contain two addresses (stETH and wstETH) but I think that's fine.

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -2478,7 +2486,17 @@ export const DEFAULT_GAS_SCHEDULE: Required<FeeSchedule> = {
...
+    [ERC20BridgeSource.Lido]: (fillData?: FillData) => {
+        const lidoFillData = fillData as LidoFillData;
+        const wethAddress = NATIVE_FEE_TOKEN_BY_CHAIN_ID[ChainId.Mainnet];
+        // WETH -> stETH
+        if (lidoFillData.takerToken === wethAddress) {
+            return 226e3;
+        }
+
+        // Wrapping or unwraping stETH.
+        return lidoFillData.takerToken === lidoFillData.stEthTokenAddress ? 120e3 : 95e3;
```
mzhu25, line 2498: can we turn this into an `if {} else if {} else {}`?

contracts/zero-ex/contracts/src/transformers/bridges/mixins/MixinLido.sol
```diff
@@ -59,12 +86,43 @@ contract MixinLido {
...
+        IERC20TokenV06 buyToken,
+        uint256 sellAmount,
+        bytes memory bridgeData
+
+    ) private returns(uint256 boughtAmount){
+        (IEtherTokenV06 stETH, IWstETH wstETH) = abi.decode(bridgeData, (IEtherTokenV06, IWstETH));
+        sellToken.approveIfBelow(address(wstETH), sellAmount);
+        if (address(sellToken) == address(stETH) && address(buyToken) == address(wstETH) ) {
+            return wstETH.wrap(sellAmount);
+        }
```
mzhu25, line 121:
```suggestion
        if (address(sellToken) == address(stETH) && address(buyToken) == address(wstETH) ) {
            sellToken.approveIfBelow(address(wstETH), sellAmount);
            return wstETH.wrap(sellAmount);
        }
```

mzhu25 (approved):
nice work!

contracts/zero-ex/contracts/src/transformers/bridges/mixins/MixinLido.sol
kh-chang: ah, thanks for catching!

packages/asset-swapper/contracts/src/LidoSampler.sol
```diff
@@ -72,20 +94,35 @@ contract LidoSampler is SamplerUtils {
...
 
-        // Minting stETH is always 1:1 therefore we can just return the same amounts back
-        return makerTokenAmounts;
-    }
+        if (takerToken == lidoInfo.wstEthToken && makerToken == lidoInfo.stEthToken) {
+            for (uint256 i = 0; i < numSamples; i++) {
+                takerTokenAmounts[i] = wstETH.getWstETHByStETH(makerTokenAmounts[i]);
+            }
+            return takerTokenAmounts;
+        }
```
dekz, line 123: On the surface these look like the same functions used in sells and buys? 

```
wstETH -> stETH ? wstETH.getWstETHByStETH
stETH -> wstETH ? wstETH.getStETHByWstETH
```

Probably a smaller bytecode footprint if you convert these to internal and reuse them.

dekz (approved):
So we will need to re-deploy the FQT/BridgeAdapter before this goes out to prod, else we might start giving out `stETH->wstETH` trades we can't settle.

packages/asset-swapper/contracts/src/LidoSampler.sol
kh-chang: very good point! -- I just re-used public sell function in buy. 
Could you take another look?

PR #478, opened 2022-05-17
feat/GMX:PTP

dextracker:
## Checklist:


packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -1951,6 +2044,32 @@ export class SamplerOperations {
                         }
                         return this.getCompoundBuyQuotes(cToken.tokenAddress, makerToken, takerToken, makerFillAmounts);
                     }
+                    case ERC20BridgeSource.GMX: {
+                        // bad mim pool dont quote
+                        if (
+                            takerToken === '0x130966628846bfd36ff31a822705796e8cb8c18d' ||
```
kh-chang, line 2050: nit: `AVALANCHE_TOKENS.MIM`

contracts/zero-ex/contracts/src/transformers/bridges/BridgeAdapter.sol
```diff
@@ -232,13 +225,6 @@ contract BridgeAdapter is
                 sellAmount,
                 order.bridgeData
             );
-        } else if (protocolId == BridgeProtocols.COFIX) {
```
dextracker, line 235: also deprecated coFix

contracts/zero-ex/contracts/src/transformers/bridges/mixins/MixinPlatypus.sol
```diff
@@ -0,0 +1,79 @@
...
+            sellAmount,
+             // Minimum buy amount.
+            0,
+            // Recipient is `this`.
+            address(this),
+
+            block.timestamp + 1
+        );
+        //calculate the buy amount from the tokens we recieved
+        boughtAmount = buyToken.balanceOf(address(this)).safeSub(beforeBalance);
```
dekz, line 76: if `amountOut` is returned by `router.swapTokensForTokens` you probably can rely on that rather than do 2x balance checks.

If Avalanche was expensive we would probably consider this

dekz (approved):

dekz:
TODO Follow up and completely remove `Kyber` (Sampler/Mixin) and the associated `ERC20BridgeSource`

PR #481, opened 2022-05-19
fix: BalancerV2 sor alias

dekz:
## Description

Inverted the [package alias](https://github.com/npm/rfcs/blob/main/implemented/0001-package-aliases.md) so load the latest version as `@balancer-labs/sdk` and `@balancer-labs/sor`.

So the previous version we relied `@balancer-labs/sor@0.3.2` on is now referenced by `balancer-labs-sor-v1`.

### Oddness with `@balancer-labs/sdk@0.1.11`
For some odd reason, in 0xAPI, when loading `@balancer-labs/sdk@0.1.11` no `@balancer-labs/sor` package was brought as a dependency.

```
λ cat node_modules/@balancer-labs/sdk/package.json| jq '.version'
"0.1.11"
λ ls node_modules/@balancer-labs
sdk
```

Asset-swapper has this version as `^0.1.6` which is what my local machine was using, which was populated with `sor` so it was difficult to reproduce. 

When restricting this version explicitly to `0.1.6` my 0xAPI was then downloading the `sor` dependency as expected.

```
0x-api λ cat node_modules/@balancer-labs/sdk/package.json| jq '.version'
"0.1.6"
0x-api λ ls node_modules/@balancer-labs
sdk sor
```

For now I have pinned us to `0.1.6`, maybe someone smarter than I can understand that exact cause.

[Simbot (Ethereum `includedSources=Balancer_V2`)](https://metabase.spaceship.0x.org/dashboard/186?run_id=verbalize-careful-thing&adjusted_amount_bought_win_tolerance__bps_=1)
[Simbot (Avalanche)](https://metabase.spaceship.0x.org/dashboard/186?run_id=influence-earthy-operation&adjusted_amount_bought_win_tolerance__bps_=1)

## Checklist:


dextracker (approved):
LGTM!

PR #489, opened 2022-06-01
[WIP] OTC Orders in fqt

dextracker:
…Consumer fill logic

## Checklist:

### DIFF
--- a/contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
+++ b/contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
@@ -20,24 +20,22 @@
 pragma solidity ^0.6.5;
 pragma experimental ABIEncoderV2;
 
-import "@0x/contracts-utils/contracts/src/v06/errors/LibRichErrorsV06.sol";
-import "@0x/contracts-erc20/contracts/src/v06/IERC20TokenV06.sol";
-import "@0x/contracts-erc20/contracts/src/v06/LibERC20TokenV06.sol";
-import "@0x/contracts-utils/contracts/src/v06/LibSafeMathV06.sol";
-import "@0x/contracts-utils/contracts/src/v06/LibMathV06.sol";
-import "../errors/LibTransformERC20RichErrors.sol";
-import "../features/interfaces/INativeOrdersFeature.sol";
-import "../features/interfaces/IOtcOrdersFeature.sol";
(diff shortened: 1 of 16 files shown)


contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -149,15 +160,19 @@ contract FillQuoteTransformer is
     /// @dev The exchange proxy contract.
     INativeOrdersFeature public immutable zeroEx;
 
+    // @dev the otc orders feature
+    IOtcOrdersFeature public immutable otcOrders;
+
```
mzhu25, line 165: we want to call `fillOtcOrder` on the `zeroEx` address, so you should just change the interface of the above `zeroEx` immutable variable to `IZeroEx` instead of introducing another variable here

phil-ociraptor:
Hey Noah - great to see this PR! I noticed that the "entrypoint" into the aggregation logic is either `getMarketSellLiquidityAsync` or `getMarketBuyLiquidityAsync`: https://github.com/0xProject/protocol/blob/a7f23a982e409076aea8670de2c4c90a4338f3e0/packages/asset-swapper/src/utils/market_operation_utils/index.ts#L632

These function signatures look like this:

```
public async getMarketBuyLiquidityAsync(
        nativeOrders: SignedNativeOrder[],
        makerAmount: BigNumber,
        opts?: Partial<GetMarketOrdersOpts>,
    ): Promise<MarketSideLiquidity> {
 ```
 
 In order for this to work, I need to pass in signed OtcOrders, which requires the `SignedNativeOrder` type, defined here: https://github.com/0xProject/protocol/blob/c9c7ac855906844f68a85d0963f7006ef03e7ea8/packages/asset-swapper/src/types.ts#L41
 
 to become:
 ```
 export type SignedNativeOrder = SignedOrder<LimitOrderFields> | SignedOrder<RfqOrderFields> | SignedOrder<OtcOrderFields>;
```

And for `SignedOrder` to be updated to:
```
export interface SignedOrder<T> {
    order: T;
    type: FillQuoteTransformerOrderType.Limit | FillQuoteTransformerOrderType.Rfq | FillQuoteTransformerOrderType.Otc;
    signature: Signature;
}
```

At some point, I imagine these will tie into the changes you have already made in the PR, but just mentioning this here as I believe a few more changes will be necessary to get everything flowing e2e

PR #493, opened 2022-06-03
Feat/ethereum bancor v3

dextracker:
## Checklist:


contracts/zero-ex/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "0.35.0",
```
kh-chang, line 3: we haven't released `0.34.0` yet so I think it can go under 34

packages/asset-swapper/contracts/src/BancorV3Sampler.sol
```diff
@@ -0,0 +1,120 @@
...
+pragma solidity ^0.6;
+pragma experimental ABIEncoderV2;
+
+import "./interfaces/IBancorV3.sol";
+
+
+contract BancorV3Sampler
+{
+    /// @dev Gas limit for BancorV3 calls.
+    uint256 constant private BancorV3_CALL_GAS = 1500e3; // 150k
```
kh-chang, line 29: another 0 here?

1500e3 = 1.5m?

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -2058,6 +2060,20 @@ export const BANCOR_REGISTRY_BY_CHAIN_ID = valueByChainId<string>(
...
 );
 
+export const BANCORV3_NETWORK_BY_CHAIN_ID = valueByChainId<string>(
+    {
+        [ChainId.Mainnet]: '0xeef417e1d5cc832e619ae18d2f140de2999dd4fb',
+    },
+    NULL_ADDRESS,
+);
+
+export const BANCORV3_NETWORK_INFO_BY_CHAIN_ID = valueByChainId<string>(
```
kh-chang, line 2070: maybe we can be more specific? 

Based on `MixinBancorV3.sol`, it looks like this is the router address, so maybe `BANCORV3_ROUTER_BY_CHAIN_ID`?

kh-chang (approved):

packages/asset-swapper/contracts/src/BancorV3Sampler.sol
dextracker: good catch this was for testing last night

PR #519, opened 2022-07-19
Fix Beethoven X Cache Issue [TKR-486]

kyu-c:
## Description

Bug Fix:
* Replace Beethoven X subgraph .

Improvements:
* Do not create `BalancerV2PoolsCache` on unsupported chains.
* Add `BalancerV2PoolsCache` test for Beethoven X.
* Remove `BalancerV2PoolsCache` test for Balancer V2 (no longer used in production). 
* Simplify `PoolsCacheMap` type.

## Checklist:

-   [ ] Add new entries to the relevant CHANGELOG.jsons. (will be updated before this PR is merged)

### DIFF
--- a/packages/asset-swapper/src/utils/market_operation_utils/pools_cache/balancer_v2_utils.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/pools_cache/balancer_v2_utils.ts
@@ -6,14 +6,14 @@ import { gql, request } from 'graphql-request';
 
 import { DEFAULT_WARNING_LOGGER } from '../../../constants';
 import { LogFunction } from '../../../types';
-import {
-    BALANCER_MAX_POOLS_FETCHED,
-    BALANCER_TOP_POOLS_FETCHED,
-    BALANCER_V2_SUBGRAPH_URL_BY_CHAIN,
-} from '../constants';
+import { BALANCER_MAX_POOLS_FETCHED, BALANCER_TOP_POOLS_FETCHED } from '../constants';
 
 import { parsePoolData } from './balancer_sor_v2';
 import { CacheValue, PoolsCache } from './pools_cache';
+
+const BEETHOVEN_X_SUBGRAPH_URL_BY_CHAIN = new Map<ChainId, string>([
+    [ChainId.Fantom, 'https://api.thegraph.com/subgraphs/name/beethovenxfi/beethovenx'],
+]);
 
 // tslint:disable-next-line:custom-no-magic-numbers
 const ONE_DAY_MS = 24 * 60 * 60 * 1000;
@@ -29,6 +29,15 @@ interface BalancerPoolResponse {
 }
 
 export class BalancerV2PoolsCache extends PoolsCache {
+    public static createBeethovenXPoolCache(chainId: ChainId): BalancerV2PoolsCache | undefined {
+        const subgraphUrl = BEETHOVEN_X_SUBGRAPH_URL_BY_CHAIN.get(chainId);
+        if (subgraphUrl === undefined) {
+            return undefined;
+        }
+
+        return new BalancerV2PoolsCache(subgraphUrl);
+    }
+
     private static _parseSubgraphPoolData(pool: any, takerToken: string, makerToken: string): Pool {
         const tToken = pool.tokens.find((t: any) => t.address === takerToken);
         const mToken = pool.tokens.find((t: any) => t.address === makerToken);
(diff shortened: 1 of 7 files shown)


packages/asset-swapper/src/utils/market_operation_utils/pools_cache/balancer_v2_utils.ts
```diff
@@ -29,6 +29,15 @@ interface BalancerPoolResponse {
 }
 
 export class BalancerV2PoolsCache extends PoolsCache {
+    public static createBeethovenXPoolCache(chainId: ChainId): BalancerV2PoolsCache | undefined {
+        const subgraphUrl = BEETHOVEN_X_SUBGRAPH_URL_BY_CHAIN.get(chainId);
+        if (subgraphUrl === undefined) {
+            return undefined;
```
dekz, line 35: What are your thoughts on usage of `undefined` versus a default identity like cache that implements `PoolCache` but always returns empty results?

kyu-c: I like the approach! 

I might need to refactor PoolCache a bit more first.
will update the PR.

kyu-c: Actually there are other related cache stuff that I want to modify as well. 

Do you mind if I address `undefined` vs no-op `PoolCache` in a separate PR?

kyu-c: addressed in #525  and #526

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -154,10 +155,7 @@ export class SamplerOperations {
         this.poolsCaches = poolsCaches
             ? poolsCaches
             : {
-                  [ERC20BridgeSource.Beethovenx]: new BalancerV2PoolsCache(
-                      chainId,
-                      BEETHOVEN_X_SUBGRAPH_URL_BY_CHAIN[chainId],
-                  ),
+                  [ERC20BridgeSource.Beethovenx]: BalancerV2PoolsCache.createBeethovenXPoolCache(chainId),
```
dekz, line 158: So only `Beethovenx` uses `BalancerV2PoolsCache` now? Since `BalancerV2` uses the newer `BalancerV2SwapInfoCache`? If that is the case I'd ultimately recommend a rename.

packages/asset-swapper/src/utils/market_operation_utils/pools_cache/balancer_v2_utils.ts
```diff
@@ -6,15 +6,15 @@ import { gql, request } from 'graphql-request';
...
-    BALANCER_TOP_POOLS_FETCHED,
-    BALANCER_V2_SUBGRAPH_URL_BY_CHAIN,
-} from '../constants';
+import { BALANCER_MAX_POOLS_FETCHED, BALANCER_TOP_POOLS_FETCHED } from '../constants';
 
 import { parsePoolData } from './balancer_sor_v2';
 import { CacheValue, PoolsCache } from './pools_cache';
 
+const BEETHOVEN_X_SUBGRAPH_URL_BY_CHAIN = new Map<ChainId, string>([
+    [ChainId.Fantom, 'https://api.thegraph.com/subgraphs/name/beethovenxfi/beethovenx'],
```
dekz, line 15: I'm on the fence about having the config/urls/addresses in files outside of `packages/asset-swapper/src/utils/market_operation_utils/constants.ts`. Do you have a preference?

dekz (approved):
Nice, left some nits, looking good to me.

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
kyu-c: Yes, that's the case!

I will rename in a follow up PR.

packages/asset-swapper/src/utils/market_operation_utils/pools_cache/balancer_v2_utils.ts
kyu-c: For the constants that are only used in a single place, I prefer those constants to be defined in the same file as it's used and not `export`ed. 

Ideally, if it's for sampling of a liquidity source, all related constants and logic (router addresses, pool info, gas schedule) are located in a single file specific to that source (when a new source is added, it should be mainly adding a single file vs touching multiple parts of the code base).

Let's discuss more on Slack!

PR #532, opened 2022-07-27
Use 0x gas api instead of eth gas station api [TKR-502]

kyu-c:
## Description

Context: 
* eth gas station api's gas price granularity is 1/10 gwei. While this level of precision has been good enough for most chains, that is no longer the case. For example, at the time of writing, Optimism gas price (fast) is 0.001 gwei. 

Change:
* Use 0x gas api which returns gas prices in wei 

## Checklist:


dekz (approved):
Yup yup yup

PR #553, opened 2022-08-18
chore: update packages

dekz:
## Description

Note: There appears to be an issue with the new Ganache where `gasPrice` is not respected. In fact, it seems to default to `0` which is a big no-no according to our `ProtocolFee` tests (they assume at least `1`).

For now I have skipped the ProtocolFee tests whilst I investigated and submit a repro to the Ganache team, or even better, we rewrite the tests in `foundry`. 

## Checklist:


kyu-c (approved):
nice!

(I will probably need to bump up packages in `0x-api` too before `asset-swapper` migration 😢 )

PR #591, opened 2022-09-26
Upgrade Prettier

eobbad:
## Checklist:

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -80,10 +80,8 @@ lib
 /contracts/erc1155/build/
 /contracts/extensions/build/
 /contracts/exchange-forwarder/build/
-/packages/asset-swapper/generated-artifacts
-/packages/asset-swapper/generated-wrappers
-/packages/asset-swapper/test/generated-artifacts
-/packages/asset-swapper/test/generated-wrappers
+/packages/asset-swapper/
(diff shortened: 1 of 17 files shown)


packages/contract-wrappers/src/generated-wrappers/broker.ts
```diff
@@ -522,7 +522,7 @@ export class BrokerContract extends BaseContract {
 
     public getABIDecodedTransactionData<T>(methodName: string, callData: string): T {
         const functionSignature = this.getFunctionSignature(methodName);
-        const self = (this as any) as BrokerContract;
```
kyu-c, line 525: we shouldn't format generated files. 

could we add `generated-wrappers` in `.prettierignore`?

packages/asset-swapper/src/utils/alt_mm_implementation_utils.ts
```diff
@@ -248,7 +248,7 @@ export async function returnQuoteFromAltMMAsync<ResponseT>(
         };
```
kyu-c, line 248: since `asset-swapper` is sort of readonly now how about if we undo changes in `asset-swapper` and add it to `.prettierignore` instead?

kyu-c (approved):

PR #594, opened 2022-09-27
fix: Migrate transformers to smart contract compatible `unsafeTransformerTransfer` [TKR-587]

kyu-c:
* Since `transfer` only forwards 2300 gas it can cause an `out of gas` revert when the receipient is a smart contract.

## Checklist:

### DIFF
--- a/contracts/zero-ex/contracts/src/transformers/LibERC20Transformer.sol
+++ b/contracts/zero-ex/contracts/src/transformers/LibERC20Transformer.sol
@@ -34,7 +34,9 @@ library LibERC20Transformer {
     ///      This is just `keccak256('TRANSFORMER_SUCCESS')`.
     bytes4 internal constant TRANSFORMER_SUCCESS = 0x13c9929e;
 
-    /// @dev Transfer ERC20 tokens and ETH.
+    /// @dev Transfer ERC20 tokens and ETH. Since it relies on `transfer` it may run out of gas when
+    /// the `recipient` is a smart contract wallet. See `unsafeTransformerTransfer` for smart contract
+    /// compatible transfer.
     /// @param token An ERC20 or the ETH pseudo-token address (`ETH_TOKEN_ADDRESS`).
     /// @param to The recipient.
     /// @param amount The transfer amount.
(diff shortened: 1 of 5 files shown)


contracts/zero-ex/contracts/src/transformers/LibERC20Transformer.sol
```diff
@@ -36,19 +36,21 @@ library LibERC20Transformer {
...
+    function unsafeTransformerTransfer(
         IERC20TokenV06 token,
         address payable to,
         uint256 amount
     )
         internal
     {
         if (isTokenETH(token)) {
-            to.transfer(amount);
+            (bool sent,) = to.call{value: amount}("");
```
idokleinman, line 52: So is `to.call{}()`  not limited to fixed 2300 gas? Do we want/need to specify max gas here in `call{gas: X}("")`? 
Why did you change the func name to unsafe...? as in what are the safety differences between transfer and call in this context?

kyu-c: `call` isn't limited to 2300 -- if we don't specify `gas` that would be forward all gas (that is what we're doing here). we could potentially specify a gas limit, but I don't think there is much benefit of doing that (would like to hear what @dekz  thinks).

> unsafe

if not safe from [re-entrancy](https://quantstamp.com/blog/what-is-a-re-entrancy-attack) attack. Since it's a common library that can be used by other contracts in the future, `unsafe` prefix is there to remind the caller to be extra cautious (suggested by @dekz).

idokleinman: Best summary I found for transfer/send/call:

image

contracts/zero-ex/contracts/src/features/FundRecoveryFeature.sol
```diff
@@ -59,7 +59,7 @@ contract FundRecoveryFeature is
         if(amountOut == uint256(-1)) {
             amountOut = LibERC20Transformer.getTokenBalanceOf(erc20, address(this));
         }
-        LibERC20Transformer.transformerTransfer(erc20, recipientWallet, amountOut);
```
dekz, line 62: IMO I would leave this unchanged. I'm actually scared that some features makes use of `LibERC20Transformer`, feels risky. 

I'd suggest keeping the old `transformerTransfer` and creating a duplicate `unsafeTransformerTransfer` to keep this audited feature unchanged.  

All the other Transformers are open season to modify though and make use of `unsafeTransformerTransfer`.

IMO in the future Features shouldn't touch `LibERC20Transformer` due to different security assumptions, but this is out of scope for this PR.

dekz (changes requested):

contracts/zero-ex/contracts/src/features/FundRecoveryFeature.sol
kyu-c: Sounds good. -- Done!

dekz (approved):

PR #596, opened 2022-10-07
Fix: FQT otc order encoding logic

dextracker:
## Description

swap rfq and otc in fillQuoteTransformerDataEncoder to match our enum
## Checklist:

### DIFF
--- a/packages/protocol-utils/src/transformer_utils.ts
+++ b/packages/protocol-utils/src/transformer_utils.ts
@@ -79,14 +79,14 @@ export const fillQuoteTransformerDataEncoder = AbiEncoder.create([
                 type: 'tuple[]',
                 components: RFQ_ORDER_INFO_ABI_COMPONENTS,
             },
+            { name: 'fillSequence', type: 'uint8[]' },
+            { name: 'fillAmount', type: 'uint256' },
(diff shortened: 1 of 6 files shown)


abls (changes requested):
I don't believe this will fix it. You have to remove the `otcOrders` field completely, it isn't present on-chain at all.

IMO would be best to just revert [this commit](https://github.com/0xProject/protocol/commit/036a2632679e3c3167162f47083d906fe305bbaa).

packages/protocol-utils/src/transformer_utils.ts
```diff
@@ -87,6 +82,11 @@ export const fillQuoteTransformerDataEncoder = AbiEncoder.create([
             { name: 'fillSequence', type: 'uint8[]' },
             { name: 'fillAmount', type: 'uint256' },
             { name: 'refundReceiver', type: 'address' },
+            {
```
kyu-c, line 85: shouldn't there be an accompanying change in FQT (solidity code)?

dextracker: yup forgot to add from /protocol was in protocol-utils

kyu-c (approved):
LGTM

PR #614, opened 2022-11-04
feat: adding AAVE V3 to selected Bridge Adapters [TKR-659]

SavDont:
## Description

Updating Bridge Adapters to support AAVE V3 MixIn.

## Checklist:


kyu-c (approved):

idokleinman (approved):

contracts/zero-ex/contracts/src/transformers/bridges/mixins/MixinAaveV3.sol
```diff
@@ -0,0 +1,166 @@
...
+        uint256 sellAmount,
+        bytes memory bridgeData
+    ) internal returns (uint256) {
+        if (_isL2) {
+            (IL2Pool pool, IL2Encoder encoder, address aToken) = abi.decode(bridgeData, (IL2Pool, IL2Encoder, address));
+
+            sellToken.approveIfBelow(address(pool), sellAmount);
+
+            if (address(buyToken) == aToken) {
+                bytes32 supplyParams = encoder.encodeSupplyParams(address(sellToken), sellAmount, 0);
```
kyu-c, line 140: If encoding happens on chain there will be no benefit of doing this (as the calldata size will the same/larger and it will cost gas to encode on chain).  I believe encoding should happen on the server-side.

kyu-c (changes requested):

PR #641, opened 2023-01-14
Decentralised governance of 0x protocol and treasury

elenadimitrova:
## Description

Decentralises governance of 0x Protocol and Treasury. This is enabled via a wrapped ZRX token and a Compound-like governors design. There are two separate governors for Protocol - `ZeroExProtocolGovernor` and Treasury - `ZeroExTreasuryGovernor` respectively working with two separate Timelock instances of the same contract implementation - `ZeroExTimelock`. 

### Upgradability
`ZRXWrappedToken` , `ZeroExProtocolGovernor` and `ZeroExTreasuryGovernor` governors are non-upgradable by design. However the voting implementation the governors use - `ZeroExVotes` is upgradable and using the OZ `ERC1967Proxy`. 

### Wrapped ZRX 
wZRX will be issued 1-to-1 for ZRX. No locking/vesting mechanisms will exist between wZRX and ZRX and the two will be freely interchangeable. The ZRX token is non-upgradable and same will be valid for its wrapped equivalent. 

The token supports delegation which allows a user to delegate their entire voting power to another account (which doesn't necessarily need to be a token holder). This is modelled on the standard openzeppelin `ERC20Votes` implementation. We have added logic for block number stamping delegators' balance changes stored in the `DelegateInfo.balanceLastUpdated` property. This block number information is sent in calls to `ZeroExVotes.moveVotingPower` in order to provide support for future upgrades to the vote power calculation.  
Note that for consistency `block.number` is used for the governor settings, voting checkpoints and this delegators' balance last updated property while timelock logic for the governor uses `block.timestamp`.

### Governors' settings
Changing governors' settings for `votingDelay`, `votingPeriod` and `proposalThreshold` can be done via the normal proposal mechanism. Governors are deployed with the following initial settings:

|                                | voting delay | voting period | proposal threshold |
|-------------------|--------------|---------------|--------------------|
| Protocol governor | 2 days            | 7 days             |  1000000e18           |
| Treasury governor | 2 days           |  7 days             |   250000e18          |

This is using standard openzeppelin `GovernorSettings` [implementation](https://docs.openzeppelin.com/contracts/4.x/api/governance#GovernorSettings).

### Quorum
Quorum for Protocol is fixed at 10m (10000000e18) while for Treasury this is calculated as 10% of voting power of the total supply (see voting strategies below for quadratic voting power implementation specifics). The quorum calculations for Treasury are using openzeppelin's `GovernorVotesQuorumFraction`.

Note that in-place updates to the quorum are not supported and will need to go through a governance upgrade. Reasoning behind this can be found in this discussion https://forum.openzeppelin.com/t/quorum-default-behaviour-on-governors/34560. 

### Voting strategies

The voting strategy will be linear 1-token-1-vote for Protocol and quadratic with threshold of 1000000e18 for Treasury (i.e. voting weight is linear up to 1m tokens and balance above that threshold is quadratic). 

Worth noting is the `Checkpoint` struct design. For packing every `Checkpoint` into a single storage slot we are using the minimum uint type size for `votes` and `quadraticVotes` members, e.g. 
``` 
struct Checkpoint {
        uint32 fromBlock;
        uint96 votes;
        uint96 quadraticVotes;
    }
```
since the maximum token supply is 1bn we can have maximum value for:
votes :                 1bn *10^18 => can be stored in 90 bits
quadraticVotes : due to the likelihood of threshold changing and potentially bringing it closer to a linear vote, we are preemptively keeping this to the same size as linear votes slot.

### Time locks
Governance proposals are subject to a 3 days delay for Protocol and 2 days for Treasury. This delay allows Security Council time to review passed proposals and take action where needed.

### Security Council 
The concept of a Security council is introduced which allows a multisig of security council members to cancel a proposal on the treasury or protocol governors and also rollback the protocol to an earlier version.

When no security council is assigned the following apply: 
- ongoing proposals can still be voted on
- new proposals cannot be created - except `assignSecurityCouncil`
- expired proposals that are successful cannot be queued - excepted `assignSecurityCouncil`

There is a provision for the governors to have different security council set although initially these are expected to be the same.

**Related design discussions**
Quadratic voting implementation https://forum.openzeppelin.com/t/quadratic-voting/34580
Governor administering another governor's settings https://forum.openzeppelin.com/t/governor-governing-another-governor-pattern/34683

Closes https://linear.app/0xproject/issue/GOV-62/contracts-implementation


contracts/governance/src/IZeroExVotes.sol
```diff
@@ -23,6 +23,12 @@ interface IZeroExVotes {
     struct Checkpoint {
         uint32 fromBlock;
         uint224 votes;
+        uint224 quadraticVotes;
```
duncancmt, line 26: This should be possible to pack this into a single struct as
```Solidity
struct Checkpoint {
    uint8 version;
    uint40 fromBlock; // best to use 40 bits to avoid overflow on all reasonable timescales
    uint96 votes; // actually only 90 bits
    uint48 quadraticVotes; // actually only 45 bits
}
```
Which also gives 64 bits left over for future work. The `version` tag makes future upgrades easier. I also recommend setting the initial version tag to 1, not 0, to make it easier to detect uninitialized structs.

contracts/governance/src/IZeroExVotes.sol
```diff
@@ -23,6 +23,12 @@ interface IZeroExVotes {
...
+    }
+
+    enum VotingType {
+        OneTokenOneVote,
+        Quadratic
     }
```
duncancmt, line 32: I recommend making 0 a forbidden value by
```Solidity
enum VotingType {
    FORBIDDEN,
    OneTokenOneVote,
    Quadratic
}
```

contracts/governance/src/ZeroExVotes.sol
```diff
@@ -168,10 +195,20 @@ contract ZeroExVotes is IZeroExVotes {
         oldWeight = oldCkpt.votes;
         newWeight = op(oldWeight, delta);
 
+        uint256 oldQuadraticWeight = oldCkpt.quadraticVotes;
+        uint256 newQuadraticWeight = op(oldQuadraticWeight, Math.sqrt(delta));
```
duncancmt, line 199: This doesn't give the correct value. Imagine Alice has 4 tokens, Bob has 0, and delta is 1. `sqrt(delta)` is 1, so at the end Bob would have 1 quadratic vote (correct) but Alice would have 3 quadratic votes, when she should have `sqrt(3) ~= 1.732`.

elenadimitrova: In your example `delta` is 4 not 1 though.

elenadimitrova: Okay so you mean in the case of transferring 1 token. Alices quadratic votes initial balance will be 2 not 3 ending up with 1 quadratic vote after the transfer which is correct minus the rounding down error.

elenadimitrova: There will be rounding errors in that though I agree.

duncancmt: This isn't just a rounding error issue. Let me be a little more explicit:

---

Starting conditions:
| Named quantity | Linear | As decimal (basis 10^18)| Quadratic | As decimal (basis 10^9) |
| :-- | :-- | :-- | :-- | :-- |
| Alice balance | 4ether  | 4.0 | 2gwei | 2.0 |
| Bob balance | 0 | 0.0 | 0 | 0.0 |
| delta | 1ether | 1.0 | 1gwei | 1.0 |

---

End conditions: 

| Named quantity | Linear | As decimal (basis 10^18) | Quadratic | As decimal (basis 10^9) |
| :-- | :-- | :-- | :-- | :-- |
| Alice balance | 3ether | 3.0 | 1gwei | 1.0 |
| Bob balance | 1ether | 1.0 | 1gwei | 1.0 |

Notice how Alice's quadratic balance is not equal to the sqrt of her linear balance. This is because the calculation on the referenced line is incorrect.

---

Correct end condition: 

| Named quantity | Linear | As decimal (basis 10^18) | Quadratic | As decimal (basis 10^9) |
| :-- | :-- | :-- | :-- | :-- |
| Alice balance | 3ether | 3.0 | 1732050807 | ~1.732 |
| Bob balance | 1ether | 1.0 | 1gwei | 1.0 |

elenadimitrova: I see now thanks

contracts/governance/src/IZeroExVotes.sol
elenadimitrova: Atm these are just the OZ types used I haven't storage optimised anywhere. I agree with the storage size for `votes` and `quadraticVotes` but I don't see why is adding a `version` field needed, considering only storage non-breaking changes will be added, e.g. at some point we might add a `uint64` to the struct. What problem do you see with that?

duncancmt: Setting a `version` makes it much easier to detect whether some future, hypothetical additional `uint64` field is semantically zero or unallocated/uninitialized (and should therefore be initialized to some nonzero value). It's probably possible in all cases to be clever about it and use the `fromBlock` field for the same purpose (assuming we know which block a backwards-incompatible change was made). This is a weak recommendation.

elenadimitrova: I prefer the data normalisation of your weak recommendation :) Just to step through the example if at block X we upgrade from V1 to V2, the voting logic as it stands is
```
if (checkpoint exists for block X) {
            // Update checkpoint
        } else {
            // Push new checkpoint
        }
```
This means that for blocks after the one of the upgrade tx, new compatible `Checkpoint`s will be pushed. For any checkpoint updates that happen in the _same_ block and _before_ the upgrade tx will succeed whereas the ones _after_ will succeed but only partially update the Checkpoint excluding any new property sets, i.e. they's behave like the previous version for that one block.

contracts/governance/src/ZeroExVotes.sol
```diff
@@ -0,0 +1,232 @@
...
+
+    function _subtract(uint256 a, uint256 b) public pure returns (uint256) {
+        return a - b;
+    }
+
+    /**
+     * @dev Access an element of the array without performing bounds check. The position is assumed to be within bounds.
+     */
+    function _unsafeAccess(Checkpoint[] storage ckpts, uint256 pos) private pure returns (Checkpoint storage result) {
+        assembly {
```
duncancmt, line 227:
```suggestion
        assembly ("memory-safe") {
```

without `"memory-safe"` the IR optimizer produces significantly worse code and it disables the stack limit evader

contracts/governance/src/ZeroExVotes.sol
```diff
@@ -0,0 +1,232 @@
...
+
+    mapping(address => Checkpoint[]) private _checkpoints;
+    Checkpoint[] private _totalSupplyCheckpoints;
+
+    modifier onlyToken() {
+        require(msg.sender == token, "Only the token is allowed to perform this operation");
+        _;
+    }
+
+    function initialize(address _token) public {
```
duncancmt, line 39:
```suggestion
    function initialize(address _token) public onlyProxy {
```

contracts/governance/test/BaseTest.t.sol
```diff
@@ -0,0 +1,69 @@
...
+        // address _address;
+        // assembly {
+        //     _address := create(0, add(_bytecode, 0x20), mload(_bytecode))
+        // }
+        // return address(_address);
+
+        ZRXMock mockZRX = new ZRXMock();
+        ZeroExVotes votes = new ZeroExVotes();
+        ZRXWrappedToken token = new ZRXWrappedToken(mockZRX, votes);
+        votes.initialize(address(token));
```
duncancmt, line 56: I'd like to implement an upgrade test for `ZeroExVotes`. Do you mind if I deploy this behind an ERC1967 proxy?

contracts/governance/src/IZeroExVotes.sol
elenadimitrova: For posterity just sharing what we agreed offline that is to not have versioning in the Checkpoint or otherwise as this can be safely added with an upgrade in the case where backward incompatible vote calculation logic is ever added, i.e. vote weight calculation that requires historical Checkpoint data.

contracts/governance/src/ZeroExVotes.sol
elenadimitrova: Just circling back to the high level quadratic voting logic and specifically calculating voting power of delegates, e.g. if person A holding X tokens and person B holding Y tokens have delegated their tokens to person C, then person C's voting power in a QV vote when calculated as `sqrt(X)+sqrt(Y)` rather than `sqrt(X+Y)` can potentially be gamed if the delegates split their balances between multiple accounts right?

duncancmt: Yes. This property is inherent to quadratic voting. Avoiding this would require adoption of advanced anti-collusion techniques, which we intend to adopt but are beyond the complexity acceptable for V1

contracts/governance/test/BaseTest.t.sol
elenadimitrova: Sorry can you hold off with the general review as this is not ready. I have not yet:
- gas optimised including packing storage and looking at types optimisation
- done any overall security tightening
- implemented upgradability
- implemented potentially separate thresholds for governance updates as opposed to other updates in either treasury or protocol - still under discussion
- added sufficient code coverage
etc.
This PR should be considered ready for overall review ONLY when it is marked as "ready to review".

elenadimitrova: Meanwhile I can really use this https://linear.app/0xproject/issue/PRO-44/zrx-artifact-is-incompatible-with-foundry plus a working Goerli environment for protocol and treasury.

gabririgo:
I would like to propose the following suggestion: a proposal that has already reached final status during voting period, i.e. that has reached quorum and > 2/3 of all voting power (in the context of this project 2/3 of WZRX total supply) should be executable during from the block after the one where final status has been reached. This is particularly relevant for protocol bugfixes, as those should reach quick consensus and be pushed forward as soon as possible.

contracts/governance/src/ZRXWrappedToken.sol
```diff
@@ -0,0 +1,125 @@
...
+*/
+pragma solidity ^0.8.17;
+
+import "@openzeppelin/token/ERC20/ERC20.sol";
+import "@openzeppelin/token/ERC20/extensions/draft-ERC20Permit.sol";
+import "@openzeppelin/token/ERC20/extensions/ERC20Wrapper.sol";
+import "@openzeppelin/governance/utils/IVotes.sol";
+import "./IZeroExVotes.sol";
+
+contract ZRXWrappedToken is ERC20, ERC20Permit, ERC20Wrapper {
```
duncancmt, line 27: Because ZRX is a well-behaved ERC20, we can override `totalSupply` in the wrapped token to call `ZRX.balanceOf(address(this))` instead of storing it in its own slot. This saves gas when minting/burning in the wrapped token.

contracts/governance/test/ZeroExProtocolGovernor.t.sol
```diff
@@ -83,4 +90,58 @@ contract ZeroExProtocolGovernorTest is ZeroExGovernorBaseTest {
...
+        vm.startPrank(securityCouncil);
+
+        bytes32 proposalId = timelock.hashOperationBatch(
+            targets,
+            values,
+            calldatas,
+            0,
+            keccak256(bytes("Emergency rollback"))
+        );
+        vm.expectEmit(false, false, false, false);
```
duncancmt, line 117: Why ignore all the topics?

contracts/governance/src/ZeroExProtocolGovernor.sol
```diff
@@ -0,0 +1,145 @@
...
+        bytes32 descriptionHash
+    ) public {
+        require(msg.sender == securityCouncil, "ZeroExProtocolGovernor: Only security council allowed");
+
+        // Execute the batch of rollbacks via the timelock controller
+        ZeroExTimelock timelockController = ZeroExTimelock(payable(timelock()));
+        timelockController.executeRollbackBatch(targets, values, calldatas, 0, descriptionHash);
+
+        // Eject security council
+        securityCouncil = address(0);
```
duncancmt, line 110: Emit an event? Same with all the other places `securityCouncil` gets set

contracts/governance/src/ZeroExTimelock.sol
```diff
@@ -0,0 +1,73 @@
...
+
+        for (uint256 i = 0; i < targets.length; ++i) {
+            address target = targets[i];
+            uint256 value = values[i];
+            bytes calldata payload = payloads[i];
+            // Check this is a rollback transaction
+            // function signature for rollback(bytes4,address)
+            // = bytes4(keccak256("rollback(bytes4,address)"))
+            // = 0x9db64a40
+            require(payload.readBytes4(0) == 0x9db64a40, "TimelockController: Not a rollback call");
```
duncancmt, line 67: This can be done without `LibBytes`
```suggestion
            require(bytes4(payload) == bytes4(0x9db64a40), "TimelockController: Not a rollback call");
```

contracts/governance/test/ZeroExProtocolGovernor.t.sol
elenadimitrova: Leftover from exclusion testing. My bad, reverted all to `true`.

contracts/governance/src/ZeroExProtocolGovernor.sol
elenadimitrova: Added `SecurityCouncilAssigned` and `SecurityCouncilEjected` events here
https://github.com/0xProject/protocol/pull/641/commits/b17c03d3e03c0229e57beced2893a3eeef468f42

contracts/governance/src/ZeroExTimelock.sol
elenadimitrova: I was trying to be consistent with the rest of the code on manipulating bytes. Okay to use the basic.

contracts/governance/src/ZeroExVotes.sol
elenadimitrova: Which proxy do you mean here?

I was assuming that this will only ever be instantiated once during deployment and if we limit it will need to be the deployer themselves but I thought it's not worth the complexity.

contracts/governance/src/ZeroExVotes.sol
```diff
@@ -0,0 +1,232 @@
...
+
+import "@openzeppelin/utils/math/SafeCast.sol";
+import "@openzeppelin/utils/math/Math.sol";
+import "@openzeppelin/token/ERC20/ERC20.sol";
+import "@openzeppelin/governance/utils/IVotes.sol";
+import "./IZeroExVotes.sol";
+
+/// @custom:security-contact security@0xproject.com
+contract ZeroExVotes is IZeroExVotes {
+    address public token;
```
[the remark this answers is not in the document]
elenadimitrova: Had to revert this as the `token` is now set in the `initialise` function rather than the constructor.

contracts/governance/src/ZRXWrappedToken.sol
elenadimitrova: Good idea, just that OZ contract still writes to the `private _totalSupply` in storage and I can't overwrite that without overriding their `_mint` and `_burn` logic.

contracts/governance/src/ZeroExVotes.sol
duncancmt: initializers that can be called on the implementation contract are extremely dangerous. It's best practice to at least add something like

```Solidity
address internal immutable _thisCopy;
constructor () {
  _thisCopy = address(this);
}
modifier onlyProxy() {
  require(address(this) != _thisCopy, "only proxy");
}
function initialize(...) onlyProxy external {
  ...
}
```

I'm pretty sure OZ supplies something equvalent

contracts/governance/src/ZeroExVotes.sol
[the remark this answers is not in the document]
duncancmt: Why is `token` set in the initializer? That makes all operations ~2.1k gas more expensive with no benefit. Is it because the OZ upgrades utility throws a fit when proxy implementations have constructors?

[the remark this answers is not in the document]
elenadimitrova: That but also because we have a catch 22 with the token using the voting implementation in its constructor and voting needing the token as well.

contracts/governance/src/ZRXWrappedToken.sol
```diff
@@ -0,0 +1,116 @@
...
+    constructor(
+        IERC20 wrappedToken,
+        IZeroExVotes _zeroExVotes
+    ) ERC20("Wrapped ZRX", "wZRX") ERC20Permit("Wrapped ZRX") ERC20Wrapper(wrappedToken) {
+        zeroExVotes = _zeroExVotes;
+    }
+
+    IZeroExVotes public zeroExVotes;
```
duncancmt, line 35:
```suggestion
    IZeroExVotes public immutable zeroExVotes;
```

contracts/governance/src/ZeroExVotes.sol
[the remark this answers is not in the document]
duncancmt: Because neither the token wrapper nor the votes contract make any calls each other as part of the constructor, it's not a big deal to take each others' addresses as arguments. Knowing the deployer address and the future deployment nonce, we can precompute the address that the votes contract will deploy to and pass that as the argument to the wrapper (or vice versa). Then, later, deploy the votes contract to that address.

This _would_ be a circular dependency if we were intending to deploy these contracts with CREATE2, but I don't think we're intending to do that.

contracts/governance/foundry.toml
```diff
@@ -0,0 +1,7 @@
+[profile.default]
+src = 'src'
+out = 'out'
+libs = ['lib']
+fs_permissions = [{ access = "read", path = "./"}]
+remappings = ['@openzeppelin/=./lib/openzeppelin-contracts/contracts/']
+optimizer_runs = 20_000
```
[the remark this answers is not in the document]
elenadimitrova: The default is 200 but since we're optimising for function runs rather than deployment size I've pushed up to 20K. Ran tests to see the difference between 20K and 1m and there's really no gas difference but deployment size is affected a little so I'd like to leave at 20K.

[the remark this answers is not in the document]
elenadimitrova: The new IR optimiser made a difference though so turned it on [here](https://github.com/0xProject/protocol/pull/641/commits/586f509f20d9658f2543057fe31ca2be7ca7c8ef)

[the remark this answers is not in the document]
elenadimitrova: I've had to revert this as unit test `testShouldNotBeAbleToDelegateWithSignatureAfterExpiry` starts failing and needs to be investigated further.

contracts/governance/src/ZeroExTreasuryGovernor.sol
```diff
@@ -0,0 +1,156 @@
...
+    GovernorVotesQuorumFraction,
+    GovernorTimelockControl
+{
+    constructor(
+        IVotes votes,
+        TimelockController _timelock,
+        address _securityCouncil
+    )
+        Governor("ZeroExTreasuryGovernor")
+        GovernorSettings(14400 /* 2 days */, 50400 /* 7 days */, 5e11)
```
SHA-2048, line 46: maybe I'm missing details here, but I think the proposal threshold is meant to be `500` which would translate to `5e20` instead of the `5e11`

contracts/governance/src/ZeroExVotes.sol
```diff
@@ -0,0 +1,359 @@
...
+        uint256 pos = _totalSupplyCheckpoints.length;
+        Checkpoint memory oldCkptTotalSuply = pos == 0
+            ? Checkpoint(0, 0, 0)
+            : _unsafeAccess(_totalSupplyCheckpoints, pos - 1);
+
+        // Remove the account sqrt balance from total quadratic supply.
+        // `accountBalance` is value _after_ minting
+        if (pos > 0) oldCkptTotalSuply.quadraticVotes -= SafeCast.toUint48(Math.sqrt(accountBalance - amount));
+
+        uint256 newLinearBalance = oldCkptTotalSuply.votes + amount;
```
SHA-2048, line 244: If I understand it correctly. the total supply quadratic votes is tracked as `sum(sqrt(user_balance))` which is higher than the `sqrt(sum(user_balance))`

in fact to reach the treasury threshold, I think it will need less than the 250K ZRX

example:
4 users with 25K each, have ~ 158 voting power each which combined will be 632 which is way bigger than the 500 = sqrt(250K) threshold and I think this gets worse the more users are involved

contracts/governance/src/ZeroExTreasuryGovernor.sol
duncancmt: The units on treasury governance voting power is in sqrt'd token balances. So the basis is 10**9, not 10**18. We'll probably be readjusting the basis back to 10**18 to support some additional features and because the difference in basis is confusing

contracts/governance/src/IZeroExVotes.sol
```diff
@@ -0,0 +1,123 @@
...
+     */
+    function getPastQuadraticTotalSupply(uint256 blockNumber) external view returns (uint256);
+
+    /**
+     * @dev Moves the voting power corresponding to `amount` number of tokens from `src` to `dst`.
+     * Note that if the delegator isn't delegating to anyone before the function call `src` = address(0)
+     * @param src the delegatee we are moving voting power away from
+     * @param dst the delegatee we are moving voting power to
+     * @param srcBalance balance of the delegate whose delegatee is `src`. This is value _after_ the transfer.
+     * @param dstBalance balance of the delegate whose delegatee is `dst`. This is value _after_ the transfer.
```
duncancmt, line 115:
```suggestion
     * @param srcBalance balance of the delegator whose delegatee is `src`. This is value _after_ the transfer.
     * @param dstBalance balance of the delegator whose delegatee is `dst`. This is value _after_ the transfer.
```

contracts/governance/src/ZeroExVotes.sol
```diff
@@ -0,0 +1,282 @@
...
+        // this always returns 0.
+        Checkpoint memory checkpoint = _checkpointsLookup(_totalSupplyCheckpoints, blockNumber);
+        return checkpoint.quadraticVotes;
+    }
+
+    /**
+     * @inheritdoc IZeroExVotes
+     */
+    function moveVotingPower(address src, address dst, uint256 srcBalance, uint256 dstBalance, uint256 amount) public {
+        if (src != dst && amount > 0) {
```
duncancmt, line 121: Although this seems like an optimization, it will make a future migration harder to apply.
```suggestion
        if (src != dst) {
```

contracts/governance/src/ZeroExVotes.sol
```diff
@@ -0,0 +1,282 @@
...
+
+        Checkpoint memory oldCkpt = pos == 0 ? Checkpoint(0, 0, 0) : _unsafeAccess(ckpts, pos - 1);
+
+        oldWeight = oldCkpt.votes;
+        newWeight = op(oldWeight, delta);
+
+        oldQuadraticWeight = oldCkpt.quadraticVotes;
+
+        // Remove the entire sqrt userBalance from quadratic voting power.
+        // Note that `userBalance` is value _before_ transfer.
```
duncancmt, line 243: This is inconsistent with the documentation in `IZeroExVotes`

contracts/governance/src/ZeroExVotes.sol
```diff
@@ -0,0 +1,282 @@
...
+            }
+        }
+
+        // Leaving here for posterity this is the original OZ implementation which we've replaced
+        // return high == 0 ? 0 : _unsafeAccess(ckpts, high - 1).votes;
+        Checkpoint memory checkpoint = high == 0 ? Checkpoint(0, 0, 0) : _unsafeAccess(ckpts, high - 1);
+        return checkpoint;
+    }
+
+    function _writeCheckpoint(
```
duncancmt, line 227: Can we make more liberal use of `unchecked` throughout this function? Seems like none of the math here can overflow. Likewise, we ought to be able to remove the use of `SafeCast`

duncancmt (commented):
The changes you've made to unify the interface on `IZeroExVotes` and reduce code duplication are _incredible_. Thanks a bunch. Expect a new PR from me early next week

contracts/governance/foundry.toml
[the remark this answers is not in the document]
elenadimitrova: I could put in alternative compilation settings for unit tests but testing it with a different compiler than production worries me so I'd rather not switch that on until the issue is fixed https://github.com/foundry-rs/foundry/issues/1373

contracts/governance/src/ZeroExVotes.sol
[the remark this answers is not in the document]
elenadimitrova: I've implemented OZ's initializable for the voting contract, see if you feel better about it

contracts/governance/src/ZeroExVotes.sol
elenadimitrova: This depends on the outcome of the discussion above https://github.com/0xProject/protocol/pull/641#discussion_r1135162016

contracts/governance/src/ZeroExVotes.sol
```diff
@@ -0,0 +1,307 @@
...
+        newWeight = op(oldWeight, delta);
+
+        oldQuadraticWeight = oldCkpt.quadraticVotes;
+
+        // Remove the entire sqrt userBalance from quadratic voting power.
+        // Note that `userBalance` is value _after_ transfer.
+        if (pos > 0) {
+            uint256 oldQuadraticVotingPower = userBalance <= quadraticThreshold
+                ? Math.sqrt(userBalance)
+                : quadraticThreshold + Math.sqrt(userBalance - quadraticThreshold);
```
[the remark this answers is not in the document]
elenadimitrova: Yes my bad on sqrt-ing the linear vote but what's with the 1e18 multiplier?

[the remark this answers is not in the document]
duncancmt: The 1e18 multiplier puts the 2 quantities in the same basis after the `sqrt`. If `userBalance - quadraticThreshold == 1 ether` then `Math.sqrt((userBalance - quadraticThreshold) * 1e18) == 1 ether`, which we want.

contracts/governance/src/ZeroExVotes.sol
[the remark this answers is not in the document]
duncancmt: This is more of a gas thing than missing `Initializeable`. If `token` is immutable, we get to avoid a cold `SLOAD` on each `onlyToken` call.

contracts/governance/foundry.toml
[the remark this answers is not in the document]
duncancmt: The specific usage we have in `testShouldNotBeAbleToDelegateWithSignatureAfterExpiry` can be fixed by setting the expiry to `block.timestamp - 1` and avoiding the use of `vm.warp` altogether. You can see what I've done in #674

contracts/governance/src/ZeroExVotes.sol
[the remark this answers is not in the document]
duncancmt: This is more of a gas thing than missing `Initializeable`. If `token` is immutable, we get to avoid a cold `SLOAD` on each `onlyToken` call.

contracts/governance/src/ZeroExVotes.sol
duncancmt: I would still recommend adding `onlyProxy` here to _completely_ prevent initialization of the implementation. It's _probably_ harmless, but let's be conservative.

contracts/governance/foundry.toml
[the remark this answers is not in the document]
elenadimitrova: Cherry-picked your commit https://github.com/0xProject/protocol/pull/674/commits/87b0e8b90789beeb3c3c8eabe4e282c2dd4ac464

contracts/governance/src/ZeroExVotes.sol
elenadimitrova: cherry-picked your commit https://github.com/0xProject/protocol/pull/674/commits/fd63a2521ae93e02b7892ae6e2017e0aa5d1da9e

contracts/governance/src/ZeroExVotes.sol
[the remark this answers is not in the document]
elenadimitrova: There is still a circular dependency though. 
voting proxy requires voting implementation
voting implementation requires wrapped token
wrapped token requires voting proxy
🤷‍♀️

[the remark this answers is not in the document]
elenadimitrova: But they _do_ require each other's addresses in their constructor / initializer logic 😕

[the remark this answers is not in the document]
elenadimitrova: `ERC1967Proxy` does check that code is present at the implementation address. This happens in its constructor.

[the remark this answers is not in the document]
duncancmt: Good catch. So we can't choose to deploy the votes implementation last. Deploying the votes proxy last doesn't help us at all. We _can_ choose to deploy wrapped token last, though.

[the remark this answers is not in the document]
elenadimitrova: Sorry I don't understand. Happy for you to push any updates you want for this.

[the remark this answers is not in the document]
duncancmt: I made #680 . If you merge that, it implements the changes I suggested above.

contracts/governance/src/ISecurityCouncil.sol
```diff
@@ -0,0 +1,81 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+*/
+pragma solidity ^0.8.19;
+
+abstract contract ISecurityCouncil {
```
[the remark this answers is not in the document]
elenadimitrova: I'll rename it as I am of the same opinion however note that openzeppelin's "interfaces" we use are in fact abstract contracts, see `IGovernor`, `IGovernorTimelock`.

contracts/governance/foundry.toml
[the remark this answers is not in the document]
elenadimitrova: Just leaving a note here as I found another strange side effect with timestamp demonstrated in the following code:

```
        uint256 lastTimestamp = block.timestamp; // e.g. if block.timestamp = 1 => lastTimestamp = 1

        vm.warp(lastTimestamp + 100); // sets block.timestamp = 101 but also sets lastTimestamp = 101 !?
```

Anyway I have gone around it but worth noting as it goes against the rules of variable assignement.

PR #643, opened 2023-01-17
Remove obsoleted code from 0x/contracts-utils

elenadimitrova:
## Description
Trying to justify the existence of `contracts-utils` by stripping it contract-by-contract.

In 0xProject this is used mainly in the API in the following contracts
https://github.com/0xProject/0x-api/blob/master/contracts/src/NativeOrderSampler.sol
https://github.com/0xProject/0x-api/blob/master/contracts/src/ApproximateBuys.sol
https://github.com/0xProject/0x-api/blob/master/contracts/src/TwoHopSampler.sol
https://github.com/0xProject/0x-api/blob/master/contracts/src/MakerPSMSampler.sol

@0x/contracts-utils/contracts/src/v06/LibBytesV06.sol
@0x/contracts-utils/contracts/src/v06/LibMathV06.sol
@0x/contracts-utils/contracts/src/v06/LibSafeMathV06.sol

Also used in the tools repo in `asset-proxy`
 https://github.com/0xProject/tools/blob/development/abi-gen/asset-proxy/contracts/src/bridges/DODOBridge.sol

@0x/contracts-utils/contracts/src/DeploymentConstants.sol

Predominantly the solc v6 contracts are in use, exceptions are
@0x/contracts-utils/contracts/src/LibSafeMath.sol
@0x/contracts-utils/contracts/src/LibRichErrors.sol
@0x/contracts-utils/contracts/src/LibBytes.sol
@0x/contracts-utils/contracts/src/Ownable.sol
Which are used in protocol contracts

It looks like we can strip down the majority of the functionality and when contracts are migrated to solc 08 we can get rid of the rest (mostly math and safemath logic).

Closes https://linear.app/0xproject/issue/PRO-45/remove-obsoleted-code-from-0xcontracts-utils


dekz (approved):
Unsure why [abi-gen/asset-proxy/contracts/src/bridges/CreamBridge.sol](https://github.com/0xProject/tools/blob/1eec1e081f73258aa42a96f530df76292d6d586d/abi-gen/asset-proxy/contracts/src/bridges/CreamBridge.sol) exists in the first place, but it can also be completely removed. As such we should be able to remove `DeploymentConstants.sol`

contracts/utils/contracts/test/TestLogDecoding.sol
```diff
@@ -1,43 +0,0 @@
-/*
```
duncancmt, line 1: I don't see any `LogDecoding`. Is this test still in use?

contracts/utils/contracts/test/TestLogDecodingDownstream.sol
```diff
@@ -1,37 +0,0 @@
-/*
```
duncancmt, line 1: I don't see the file that this tests in the diff. Is this test still in use?

contracts/utils/contracts/test/TestRefundable.sol
```diff
@@ -1,49 +0,0 @@
-/*
```
duncancmt, line 1: ditto

contracts/utils/contracts/test/TestRefundableReceiver.sol
```diff
@@ -1,154 +0,0 @@
-/*
```
duncancmt, line 1: ditto

contracts/utils/test/log_decoding.ts
```diff
@@ -1,98 +0,0 @@
-import { chaiSetup, provider, txDefaults, web3Wrapper } from '@0x/contracts-test-utils';
```
duncancmt, line 1: ditto

contracts/utils/test/refundable.ts
```diff
@@ -1,141 +0,0 @@
-import { blockchainTests, constants } from '@0x/contracts-test-utils';
```
duncancmt, line 1: ditto

duncancmt (approved):
Please double check that the files I've referenced are no longer in use

contracts/utils/contracts/test/TestLogDecoding.sol
elenadimitrova: I think this was testing decoding from `ethereum-types` library rather than any protocol logic https://github.com/0xProject/protocol/pull/643/commits/e0873b716b88d59a5ad245626547148cc8d439fb#diff-f84b381c715449735700973aa5cf405d5ebdecf1c6fcc7c0fa261b847e5cf476L5
if that's what you meant?

contracts/utils/contracts/test/TestLogDecodingDownstream.sol
elenadimitrova: Same as above

contracts/utils/contracts/test/TestRefundable.sol
elenadimitrova: This was testing the `Refundable` contract which was removed in the same commit https://github.com/0xProject/protocol/pull/643/commits/d19e8eead276d8191c071f7f41c39df6d2f710f7

contracts/utils/contracts/test/TestRefundableReceiver.sol
elenadimitrova: Same as above

duncancmt (approved):
LGTM

contracts/utils/test/refundable.ts
elenadimitrova: This is testing the `Refundable` contract which was removed so tests were removed

PR #650, opened 2023-01-27
feat: Migrate PositiveSlippage tests [LIT-804]

dekz:
## Checklist:


dextracker (approved):
Ayyyyyyyyyyy

phil-ociraptor (approved):
Looks good! 

Question though - how do we test the combination of these two transformers? I feel like those would best go into an "integration" test suite for `transformERC20`, but by the time we get that suite, we'll forget to add this kind of test

Maybe we just need to track this as an issue until we have the right tests migrated over?

PR #654, opened 2023-02-01
Migrate meta tx tests [RFQ-793]

patrick-dowell:
## Description

As protocol now can run forge tests, we want to port various tests from typescript to forge. Here we are porting over the metatransaction tests from meta_transaction_tests.ts. Not every test was ported, just the ones that test the most important functionality.

This is a WIP, so I'm putting it up as a draft PR.

## Testing instructions

Run forge test.

## Types of changes

* New forge tests.

## Checklist:

-   [ ] Add tests to cover changes as needed.
-   [ ] Update documentation as needed.
-   [ ] Add new entries to the relevant CHANGELOG.jsons.

PR #655, opened 2023-02-01
Migrate the multiplex tests to foundry

abls:
## Description

This is intended to be a very faithful migration of the all test cases from `contracts/zero-ex/test/features/multiplex_test.ts` into solidity. This PR is being drafted before all of the test cases are migrated to give anyone a chance to provide feedback before the general structure of these tests is replicated many times.

## Testing instructions

`forge test --match-test test_multiplex -vv`

## Types of changes

* New forge tests added
* Existing utility code for forge tests modified

## Checklist:

-   [x] Add the test cases that cover most of what is necessary
-   [x] Refactor all the new utility code created during this process out of the test file
-   [ ] Implement the remaining test cases


contracts/zero-ex/tests/Multiplex.t.sol
```diff
@@ -0,0 +1,1453 @@
...
+    address private signerAddress;
+    uint256 private signerKey;
+
+    modifier explain(string memory testName) {
+        log_string(testName);
+        vm.recordLogs();
+        _;
+    }
+
+    function infiniteApprovals() private {
```
duncancmt, line 144: https://github.com/0xProject/protocol/pull/654#discussion_r1094771125

contracts/zero-ex/tests/Multiplex.t.sol
```diff
@@ -0,0 +1,1453 @@
...
+        vm.startPrank(signerAddress);
+        infiniteApprovals();
+        vm.stopPrank();
+
+        vm.deal(address(this), 10e18);
+    }
+
+    // TODO refactor some of these utility functions out into helper contract
+
+    function makeTestRfqOrder(
```
duncancmt, line 193: These `make.*{Order,Subcall}` functions ought to be in a library or at least in a separate contract.

contracts/zero-ex/tests/Multiplex.t.sol
```diff
@@ -0,0 +1,1453 @@
...
+    ) private pure returns (IMultiplexFeature.BatchSellSubcall memory) {
+        return
+            IMultiplexFeature.BatchSellSubcall({
+                id: IMultiplexFeature.MultiplexSubcall.MultiHopSell,
+                sellAmount: sellAmount,
+                data: abi.encode(tokens, calls)
+            });
+    }
+
+    function makeArray(
```
duncancmt, line 398: Do we really need this many specialized versions of this pattern? Does this get reused sufficiently often that it warrants being separated out? This is a lot of code to scroll through.

contracts/zero-ex/tests/Multiplex.t.sol
```diff
@@ -0,0 +1,1453 @@
...
+import "../contracts/test/integration/TestUniswapV2Pool.sol";
+import "../contracts/test/integration/TestUniswapV3Factory.sol";
+import "../contracts/test/integration/TestUniswapV3Pool.sol";
+import "../contracts/test/integration/TestLiquidityProvider.sol";
+
+import "@0x/contracts-erc20/contracts/src/v06/WETH9V06.sol";
+
+contract Multiplex is Test, ForkUtils, TestUtils {
+    uint256 private constant MAX_UINT256 = type(uint256).max;
+    uint256 private constant HIGH_BIT = 2 ** 255;
```
duncancmt, line 45: This only gets used once. It shouldn't be broken out as a named constant

contracts/zero-ex/tests/Multiplex.t.sol
```diff
@@ -0,0 +1,1453 @@
...
+    ) private pure returns (LogDescription memory) {
+        assert(dataMask.length == 0 || dataMask.length == data.length);
+        return LogDescription(topics, data, dataMask);
+    }
+
+    function describeLog(bytes32[] memory topics, bytes memory data) private pure returns (LogDescription memory) {
+        return describeLog(topics, data, hex"");
+    }
+
+    function assertLogs(LogDescription[] memory expected) private {
```
duncancmt, line 587: What's the purpose of this function over [`vm.expectEmit`](https://book.getfoundry.sh/cheatcodes/expect-emit)?

contracts/zero-ex/tests/Multiplex.t.sol
```diff
@@ -0,0 +1,1453 @@
...
+    TestUniswapV2Factory private sushiFactory;
+    TestUniswapV2Factory private uniV2Factory;
+    TestUniswapV3Factory private uniV3Factory;
+    TestLiquidityProvider private liquidityProvider;
+    uint256 private transformerNonce;
+
+    address private signerAddress;
+    uint256 private signerKey;
+
+    modifier explain(string memory testName) {
```
duncancmt, line 138: Do we consume this log output in a GitHub action or something? Seems needlessly noisy to me. We definitely want the relevant output when the test fails, though

contracts/zero-ex/tests/Multiplex.t.sol
```diff
@@ -0,0 +1,1453 @@
...
+                rfqOrder.takerAmount,
+                rfqOrder.makerAmount + 1
+            )
+        {
+            fail("did not revert");
+        } catch Error(string memory reason) {
+            assertEq(reason, "MultiplexFeature::_multiplexBatchSell/UNDERBOUGHT", "wrong revert reason");
+        } catch {
+            fail("low-level revert");
+        }
```
duncancmt, line 649: Use [`vm.expectRevert`](https://book.getfoundry.sh/cheatcodes/expect-revert). In general, `try .. catch` in Solidity is an antipattern because it has many non-obvious failure modes that can result in security vulnerabilities. Of course, this is test code, but it's bad to build the habit.

contracts/zero-ex/tests/Multiplex.t.sol
```diff
@@ -0,0 +1,1453 @@
...
+                        )
+                    })
+                )
+            );
+        } catch Error(string memory reason) {
+            fail("reverted");
+            fail(reason);
+        } catch {
+            fail("low-level revert");
+        }
```
duncancmt, line 738: Why?

Just let the error propagate and the test will fail with a relevant message to the console

duncancmt (commented):
I added some comments about idiomatic use of foundry to improve the test cases. I'll re-review after either I understand why we can't use the test framework functions or the tests have been rewritten to use them. If we _do_ end up avoiding the use of foundry's builtins for a good reason, we need to refactor our custom error handling into another file so that it's easier to see what is infrastructure vs what is application test code

contracts/zero-ex/tests/Multiplex.t.sol
abls: Yep I agree, hence the `TODO` immediately above. Just while this is WIP I have been dumping everything inside this contract. Probably a library in the same file would've been a little cleaner, the next version will at least have that.

contracts/zero-ex/tests/Multiplex.t.sol
abls: This function can mask out parts of the log's data when comparing the expected logs with the emitted logs. We can do a sort of "fuzzy match" on logs with this. In these tests it is used mostly for masking out the `value` field of `Transfer` events, but it is also used to mask out several fields from a `MintTransform` event. I have yet to encounter a case in the original typescript tests where we would need to mask topics, but if we ever do encounter that case then we would add that ability to this function as well.

We will want to mask pieces of the logs out whenever the component we are testing interacts with something outside itself and we want to know some details of their interaction but do not care about all of the details. Some of those details we ignore can even be incorrect if something is wrong with the "other" component, but as long as it doesn't affect what we are testing then it is good that we can ignore it. I imagine that this would be especially useful for something like metatransactions, but I haven't looked at those tests to confirm.

abls: Afaik `vm.expectEmit` cannot match _parts_ of the data, only ignore it or match it entirely.

duncancmt: Ahh, you're right. Is that feature really this important that we're willing to add this much complexity to our testing infra? It makes me itchy

contracts/zero-ex/tests/Multiplex.t.sol
abls: The idea here was to increase readability in each of the test cases. We don't need it, but imo it helps readability to be able to initialize dynamic arrays inline. It lets us express the tests more concisely and clearly than we otherwise could.

Some of these `makeArray` functions specific to objects used only in `MultiplexFeature` wouldn't make sense to move outside of this file, but some of the more generic `makeArray` functions would ideally be in some helper code shared by all tests.

contracts/zero-ex/tests/Multiplex.t.sol
abls: Oh yea definitely, good point

contracts/zero-ex/tests/Multiplex.t.sol
abls: This is here because I was trying to find a good way to label the tests with the descriptions from the original typescript tests. This is just about as noisy as running the typescript tests we use right now, but I do agree that maybe even the typescript tests are a bit noisy. Unsure what a good alternative would be though.

It doesn't seem right to me to shoehorn an English sentence into the already-long name of a function, like `test_multiplexBatchSellTokenForToken_revertsIfMinBuyAmountIsNotSatisfied` especially if we want to be any more descriptive or include special characters. It encourages writers of tests to make the descriptions more abbreviated, truncated, and unclear. So, logging the description of the test seems to be a good solution. I agree that only logging on failure would be better, but afaik we can't reliably do that with forge (we can if we always use try/catch in tests, but if we rely on reverts to fail a test then we can't).

contracts/zero-ex/tests/Multiplex.t.sol
abls: This whole try/catch thing I'm doing is a sort of vestigial pattern left over from when I had the tests structured in a different way, and I neglected to realize it was no longer necessary after some rewriting. Definitely a good catch 😅

contracts/zero-ex/tests/utils/DeployZeroEx.sol
```diff
@@ -293,7 +323,10 @@ contract DeployZeroEx is Test {
 
         ZERO_EX_DEPLOYED.zeroEx = IZERO_EX;
         isDeployed = true;
-        logDeployedZeroEx();
+        if (ZERO_EX_DEPLOY_CONFIG.logDeployed) {
```
patrick-dowell, line 326: why make this optional?

contracts/zero-ex/tests/utils/MultiplexUtils.sol
```diff
@@ -0,0 +1,304 @@
...
+            makerAmount: 1e18,
+            takerAmount: 1e18,
+            maker: signerAddress,
+            taker: address(this),
+            txOrigin: tx.origin,
+            pool: 0x0000000000000000000000000000000000000000000000000000000000000000,
+            expiry: uint64(block.timestamp + 60),
+            salt: 123
+        });
+        _mintTo(address(order.makerToken), order.maker, order.makerAmount);
```
patrick-dowell, line 47: don't want to mint the taker token as well here?

contracts/zero-ex/tests/utils/MultiplexUtils.sol
```diff
@@ -0,0 +1,304 @@
...
+    ) internal pure returns (IMultiplexFeature.BatchSellSubcall memory) {
+        return
+            IMultiplexFeature.BatchSellSubcall({
+                id: IMultiplexFeature.MultiplexSubcall.MultiHopSell,
+                sellAmount: sellAmount,
+                data: abi.encode(tokens, calls)
+            });
+    }
+
+    function _makeArray(
```
patrick-dowell, line 236: It seems a little strange to have to define so many different versions of makeArray, that are basically all the same but can have different types and different numbers of arguments. I assume there's no inline way to do this in solidity. You do use this in almost 50 places, so I can see that it is saving some lines of code to have these functions. I just wonder if there's a better / cleaner and more general way to do this.

contracts/zero-ex/tests/Multiplex.t.sol
```diff
@@ -0,0 +1,651 @@
...
+            abi.encode(dai, zrx, rfqOrder.takerAmount, 5e17, 0),
+            rfqOrder.takerAmount,
+            0
+        );
+
+        vm.expectEmit(true, true, true, true);
+        emit Transfer(address(dai), address(flashWallet), address(0), 1e18);
+
+        vm.expectEmit(true, true, true, true);
+        emit Transfer(address(zrx), address(flashWallet), rfqOrder.taker, 5e17);
```
patrick-dowell, line 274: It looks like you would rather just check the events vs check the token balances for these tests. Wouldn't checking the balances be a more reliable way of verifying the trade went through?

contracts/zero-ex/tests/Multiplex.t.sol
```diff
@@ -0,0 +1,651 @@
...
+            dai,
+            zrx,
+            _makeArray(rfqSubcall, nestedMultiHopSubcall),
+            sellAmount,
+            0
+        );
+    }
+
+    function testMultiplexBatchSellEthForToken() public {
+        // RFQ
```
patrick-dowell, line 405: what's going on here and on line 415?

contracts/zero-ex/tests/Multiplex.t.sol
```diff
@@ -0,0 +1,651 @@
...
+                _makeNestedBatchSellSubcall(_makeArray(rfqSubcall, uniswapV2Subcall)),
+                _makeUniswapV3MultiHopSubcall(_makeArray(address(shib), address(zrx)))
+            ),
+            sellAmount,
+            0
+        );
+    }
+
+    function testMultiplexMultiHopSellEthForToken() public {
+        // reverts if first token is not WETH
```
patrick-dowell, line 637: ??

contracts/zero-ex/tests/WrapEth.t.sol
```diff
@@ -15,7 +15,7 @@
 pragma solidity ^0.6;
 pragma experimental ABIEncoderV2;
 
-import "utils/ForkUtils.sol";
+import {ForkUtils} from "utils/ForkUtils.sol";
```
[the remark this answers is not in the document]
abls: I didn't write this file, was just blindly fixing import issues with this change. Tbh it doesn't look like it was ever using anything from ForkUtils?

contracts/zero-ex/tests/WrapEth.t.sol
```diff
@@ -33,7 +33,9 @@ contract WrapEth is Test, ForkUtils, TestUtils {
     DeployZeroEx.ZeroExDeployed zeroExDeployed;
 
     function setUp() public {
-        zeroExDeployed = new DeployZeroEx().deployZeroEx();
+        zeroExDeployed = new DeployZeroEx(
```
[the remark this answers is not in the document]
abls: We don't have overloaded constructors in solidity, but maybe some sort of utility function somewhere that just constructs DeployZeroEx with all zeros for the config would be nice. All the DeployConfig code really is awkwardly bolted onto DeployZeroEx, I was just shooting for something quick with minimal changes to DeployZeroEx so that we could get the tests we needed running. Definitely can and should be cleaned up, probably should loop @dextracker into that conversation.

contracts/zero-ex/tests/utils/DeployZeroEx.sol
abls: That logging is way too noisy imo. If we are using `DeployZeroEx` in `setUp` (as it is probably best used), for every local test we get a large identical dump of addresses. It is unclear to me why we are logging these addresses in the first place, and in the cases I can think of that this would be useful in, some combination of assertions and `vm.label` seems like a better alternative.

contracts/zero-ex/tests/Multiplex.t.sol
duncancmt: I think there's some value in checking both. Making the tests robust so that if something strange changes the test starts failing is a good idea.

For example, the way these `vm.expectEmit`s are structure, if the exchange proxy emits the `Transfer` event (or the wrong token, or any other contract involved), the test will still pass. Checking the balances in addition to the events would catch that.

FWIW, there's a `vm.expectEmit(bool, bool, bool, bool, address)` variant that would fix :point_up:

contracts/zero-ex/tests/Multiplex.t.sol
duncancmt: ++

Please remove

contracts/zero-ex/tests/Multiplex.t.sol
duncancmt: ++

duncancmt (approved):

PR #665, opened 2023-02-21
feat: Multiplex + MetaTransaction integration and MetaTransaction Multi-Fee Support [RFQ-795] [LIT-870]

patrick-dowell:
## Description

There are four big changes included in this PR:

1) Allow `executeMetaTransaction `to pay out multiple fees by modifying the data structure `MetaTransactionData`. This results in the creation of a v2 MetaTransactions feature contract and `MetaTransactionV2Data `struct.

2) Inclusion of four new selectors in `MetaTransactionsFeatureV2`:
 * `_executeMultiplexBatchSellTokenForTokenCall`
 * `_executeMultiplexBatchSellTokenForEthCall`
 * `_executeMultiplexMultiHopSellTokenForTokenCall`
 * `_executeMultiplexMultiHopSellTokenForEthCall`

3) Modifications to Multiplex to account for the possibility that the `MultiplexFeature `is entered via `MetaTransactionFeatureV2`. Specifically, we refactor calls to `msg.sender` into `BatchSellParams.payer` and `MultiHopSellParams.payer`

4) Add a call to `multiHopBatchSellOtc `to `MultiplexFeature `and add a conditional for `MultiplexSubcall.OTC` in Multiplex's `_computeHopTarget`. To accomplish this, changes were cherrypicked from #667 at https://github.com/0xProject/protocol/pull/665/commits/f32e834e111ffb9717ae39874ffc70192b688c5b.

## Testing instructions

Run forge test from contracts/zero-ex. There are numerous new local tests and a new forked test that cover the new functionality and contract, located at `MetaTransactionV2Test.t.sol`, `MultiplexMetaTransactionsV2.t.sol`, and `MultiplexRfqTest.t.sol`.

## Types of changes

* A new feature contract `MetaTransactionsFeatureV2 `was created, copying logic from `MetaTransactionsFeature`.
* New functionality was added to `MetaTransactionsFeatureV2`
* New functionality was added to `MultiplexFeature`
* forge tests were added for the new feature contract and to test all new functionality.

## Checklist:


contracts/zero-ex/contracts/src/features/interfaces/IMultiplexFeature.sol
```diff
@@ -46,6 +46,8 @@ interface IMultiplexFeature {
         bool useSelfBalance;
         // The recipient of the bought output tokens.
         address recipient;
+        // The sender of the transaction.
+        address msgSender;
```
duncancmt, line 50: Is this the beneficiary of the metatransaction, or the relayer? We should invent new terminology here so that it's clear.

contracts/zero-ex/tests/utils/TestUtils.sol
```diff
@@ -17,9 +17,16 @@ pragma experimental ABIEncoderV2;
...
+import "./DeployZeroEx.sol";
 import "src/transformers/LibERC20Transformer.sol";
+import "src/features/libs/LibSignature.sol";
+import "src/features/libs/LibNativeOrder.sol";
+import "../../contracts/test/tokens/TestMintableERC20Token.sol";
+import "@0x/contracts-erc20/src/IEtherToken.sol";
 
 contract TestUtils is Test {
+    address private constant ZERO_ADDRESS = 0x0000000000000000000000000000000000000000;
+
```
duncancmt, line 29: Why not just use `address(0)`?

contracts/zero-ex/tests/utils/TestUtils.sol
```diff
@@ -29,4 +36,126 @@ contract TestUtils is Test {
...
+    }
+
+    function makeTestRfqOrder(
+        DeployZeroEx.ZeroExDeployed memory zeroExDeployed, 
+        IERC20Token makerToken, 
+        IERC20Token takerToken,
+        address makerAddress,
+        address takerAddress,
+        uint256 makerKey
+    ) public returns (bytes memory callData) {
```
duncancmt, line 54: Generally, tests shouldn't have `public`/`external` functions that aren't test cases

contracts/zero-ex/contracts/src/features/MetaTransactionsFeatureV2.sol
```diff
@@ -0,0 +1,620 @@
...
+    FixinEIP712,
+    FixinTokenSpender
+{
+    using LibBytesV06 for bytes;
+    using LibRichErrorsV06 for bytes;
+
+    /// @dev Describes the state of a meta transaction.
+    struct ExecuteState {
+        // Sender of the meta-transaction.
+        address sender;
```
duncancmt, line 51: Again, this is confusing terminology

contracts/zero-ex/contracts/src/features/MetaTransactionsFeatureV2.sol
```diff
@@ -0,0 +1,620 @@
...
+    /// @param state The `ExecuteState` for this metatransaction, with `sender`,
+    ///              `hash`, `mtx`, and `signature` fields filled.
+    /// @return returnResult The ABI-encoded result of the underlying call.
+    function _executeMetaTransactionPrivate(ExecuteState memory state) private returns (bytes memory returnResult) {
+        _validateMetaTransaction(state);
+
+        // Mark the transaction executed by storing the block at which it was executed.
+        // Currently the block number just indicates that the mtx was executed and
+        // serves no other purpose from within this contract.
+        LibMetaTransactionsStorage.getStorage().mtxHashToExecutedBlockNumber[state.hash] = block.number;
```
duncancmt, line 242: If this isn't used elsewhere, we should get rid of it. Penalizing the relayer an extra 20k gas per transaction is hefty

contracts/zero-ex/contracts/src/features/MetaTransactionsFeatureV2.sol
```diff
@@ -0,0 +1,620 @@
...
+        }
+
+        if (LibSignature.getSignerOfHash(state.hash, state.signature) != state.mtx.signer) {
+            LibSignatureRichErrors
+                .SignatureValidationError(
+                    LibSignatureRichErrors.SignatureValidationErrorCodes.WRONG_SIGNER,
+                    state.hash,
+                    state.mtx.signer,
+                    // TODO: Remove this field from SignatureValidationError
+                    //       when rich reverts are part of the protocol repo.
```
duncancmt, line 293: Open TODO

contracts/zero-ex/contracts/src/features/MetaTransactionsFeatureV2.sol
```diff
@@ -0,0 +1,620 @@
...
+                mstore(add(encodedStructArgs, 32), 32)
+                // Copy everything after the selector.
+                fromMem := add(fromCallData, 36)
+                // Start copying after the struct offset.
+                toMem := add(encodedStructArgs, 64)
+            }
+            LibBytesV06.memCopy(toMem, fromMem, fromCallData.length - 4);
+            // Decode call args for `ITransformERC20Feature.transformERC20()` as a struct.
+            args = abi.decode(encodedStructArgs, (ExternalTransformERC20Args));
+        }
```
duncancmt, line 360: I still think this would be cleaner with the destructive version:
```Solidity
{
    bytes memory data = state.mtx.callData;
    assert(data.length > 4);
    assembly {
        let newLen := sub(mload(data), 4)
        data := add(data, 4)
        mstore(data, newLen)
    }
    state.mtx.callData = data;
}
args = abi.decode(state.mtx.callData, ExternalTransformERC20Args);
```

@dekz do you have an opinion?

contracts/zero-ex/tests/MetaTransactionV2Test.t.sol
```diff
@@ -0,0 +1,311 @@
...
+import "../contracts/test/TestMintTokenERC20Transformer.sol";
+import "../contracts/src/features/libs/LibSignature.sol";
+import "src/features/libs/LibNativeOrder.sol";
+import "../contracts/test/tokens/TestMintableERC20Token.sol";
+import "@0x/contracts-utils/contracts/src/v06/errors/LibRichErrorsV06.sol";
+import "@0x/contracts-erc20/src/IEtherToken.sol";
+
+contract MetaTransactionTest is BaseTest, TestUtils {
+    DeployZeroEx.ZeroExDeployed zeroExDeployed;
+    address private constant ZERO_ADDRESS = 0x0000000000000000000000000000000000000000;
```
duncancmt, line 33: Use `address(0)`

contracts/zero-ex/tests/MetaTransactionV2Test.t.sol
```diff
@@ -0,0 +1,311 @@
...
+        assertEq(usdcToken.balanceOf(USER_ADDRESS), 0);
+        assertEq(wethToken.balanceOf(address(this)), 1);
+    }
+
+    function test_rfqOrder() external {
+        bytes memory callData = makeTestRfqOrder(zeroExDeployed, zrxToken, usdcToken, signerAddress, USER_ADDRESS, signerKey);
+        IMetaTransactionsFeatureV2.MetaTransactionDataV2 memory mtxData = _getMetaTransaction(callData);
+
+        assertEq(usdcToken.balanceOf(USER_ADDRESS), oneEth);
+        vm.expectEmit(true, false, false, true);
```
duncancmt, line 163: Are topics 2 and 3 not stable?

contracts/zero-ex/contracts/src/features/MetaTransactionsFeatureV2.sol
```diff
@@ -0,0 +1,620 @@
...
+    bytes32 public immutable MTX_EIP712_TYPEHASH =
+        keccak256(
+            "MetaTransactionDataV2("
+            "address signer,"
+            "address sender,"
+            "uint256 expirationTimeSeconds,"
+            "uint256 salt,"
+            "bytes callData,"
+            "address feeToken,"
+            "uint256 feeAmount,"
```
abls, line 89: `feeAmount` is not a member of this struct

contracts/zero-ex/contracts/src/features/MetaTransactionsFeatureV2.sol
```diff
@@ -0,0 +1,620 @@
...
+                keccak256(
+                    abi.encode(
+                        MTX_EIP712_TYPEHASH,
+                        mtx.signer,
+                        mtx.sender,
+                        mtx.expirationTimeSeconds,
+                        mtx.salt,
+                        keccak256(mtx.callData),
+                        mtx.feeToken,
+                        mtx.fees
```
abls, line 226: This isn't a valid struct hash afaik. Pretty sure the correct way to encode the fees is something like `keccak256(abi.encode(mtx.fees))`, but you should verify. It may be that you need to recursively encode the fee structs with their typehash too. See the [original branch](https://github.com/0xProject/protocol/blob/929d35c30cb20f1f4ce7936539fe7412c8177af1/contracts/zero-ex/contracts/src/features/MetaTransactionsFeature.sol#L225) with this change. Also see [EIP712 encodeData](https://eips.ethereum.org/EIPS/eip-712#definition-of-encodedata).

abls (changes requested):

contracts/zero-ex/tests/MetaTransactionV2Test.t.sol
```diff
@@ -0,0 +1,311 @@
...
+import "src/features/libs/LibNativeOrder.sol";
+import "../contracts/test/tokens/TestMintableERC20Token.sol";
+import "@0x/contracts-utils/contracts/src/v06/errors/LibRichErrorsV06.sol";
+import "@0x/contracts-erc20/src/IEtherToken.sol";
+
+contract MetaTransactionTest is BaseTest, TestUtils {
+    DeployZeroEx.ZeroExDeployed zeroExDeployed;
+    address private constant ZERO_ADDRESS = 0x0000000000000000000000000000000000000000;
+    address private constant USER_ADDRESS = 0x6dc3a54FeAE57B65d185A7B159c5d3FA7fD7FD0F;
+    uint256 private constant USER_KEY = 0x1fc1630343b31e60b7a197a53149ca571ed9d9791e2833337bbd8110c30710ec;
```
abls, line 35: While `USER_KEY` and address constants were useful when thrown together for that [reference script](https://gist.github.com/abls/bb91ef9bb1ee1dd04dade9b2afeb2bf0#file-gas-s-sol-L22), we should probably have a better system for this in the test utils. A lot of things from that script are not good practice because that script was intended for a single use within an entirely different context.

Regarding improving our utils, a single `getSigner` function that can return only one signer isn't enough. It is common in our tests that we need multiple signers. Also, maybe we should be able to sign for any of the `account`s in `BaseTest`. Unsure exactly the best solution, but imo all the people writing tests should have a discussion about organization of the shared test utils at some point.

contracts/zero-ex/contracts/src/features/MetaTransactionsFeatureV2.sol
duncancmt: Yeah, that's a good point. The correct way to encode an array of structs is to first encode (hash) each struct individually (with its typehash) and then hash the array. If you're encoding an array of basic types, you just hash the array.

contracts/zero-ex/contracts/src/features/MetaTransactionsFeatureV2.sol
patrick-dowell: This terminology is consistent with meta-txn v1, though I agree, it is a bit confusing. The sender is the person submitting the meta-txn on someone else's behalf. The signer is the taker.

contracts/zero-ex/contracts/src/features/MetaTransactionsFeatureV2.sol
patrick-dowell: Other than storing information in a map, how else would we know that a metatransaction has been executed? It's important you can't execute the metatransaction again, since these are being submitted by a third party on the signer's behalf.

contracts/zero-ex/contracts/src/features/MetaTransactionsFeatureV2.sol
patrick-dowell: This is sadly copy/pasted from the original metatransaction v1. Happy to open a TODO if this is something we actually want to fix.

contracts/zero-ex/contracts/src/features/MetaTransactionsFeatureV2.sol
```diff
@@ -0,0 +1,630 @@
...
+    bytes32 public immutable MTX_EIP712_TYPEHASH =
+        keccak256(
+            "MetaTransactionDataV2("
+            "address signer,"
+            "address sender,"
+            "uint256 expirationTimeSeconds,"
+            "uint256 salt,"
+            "bytes callData,"
+            "address feeToken,"
+            "uint256 feeAmount,"
```
Haozhuo, line 89: `feeAmount` field does not exist in `MetaTransactionDataV2` right?

contracts/zero-ex/contracts/src/features/MetaTransactionsFeatureV2.sol
patrick-dowell: This issue should be fixed in the latest version.

contracts/zero-ex/contracts/src/features/MetaTransactionsFeatureV2.sol
patrick-dowell: This issue should be fixed in the latest version. Jacob and I tested against MetaMask to verify that the signature we are produces matches what MetaMask would produce signing a MetaTransactionDataV2 struct.

contracts/zero-ex/contracts/src/features/interfaces/IMultiplexFeature.sol
patrick-dowell: In a discussion we had earlier today, @duncancmt and @abls and I felt that the best naming would be payer (// the sender of the input token), which kinda mirrors what's going on with recipient - the receiver of the output token.

contracts/zero-ex/contracts/src/features/MetaTransactionsFeatureV2.sol
```diff
@@ -0,0 +1,641 @@
...
+    /// @param state The `ExecuteState` for this metatransaction, with `sender`,
+    ///              `hash`, `mtx`, and `signature` fields filled.
+    /// @return returnResult The ABI-encoded result of the underlying call.
+    function _executeMetaTransactionPrivate(ExecuteState memory state) private returns (bytes memory returnResult) {
+        _validateMetaTransaction(state);
+
+        // Mark the transaction executed by storing the block at which it was executed.
+        // Currently the block number just indicates that the mtx was executed and
+        // serves no other purpose from within this contract.
+        LibMetaTransactionsStorage.getStorage().mtxHashToExecutedBlockNumber[state.hash] = block.number;
```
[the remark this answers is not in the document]
patrick-dowell: Ok, I made the adjustment.

contracts/zero-ex/contracts/src/features/MetaTransactionsFeatureV2.sol
```diff
@@ -0,0 +1,641 @@
...
+        IERC20Token outputToken;
+        uint256 inputTokenAmount;
+        uint256 minOutputTokenAmount;
+        ITransformERC20Feature.Transformation[] transformations;
+    }
+
+    /// @dev Name of this feature.
+    string public constant override FEATURE_NAME = "MetaTransactionsV2";
+    /// @dev Version of this feature.
+    uint256 public immutable override FEATURE_VERSION = _encodeVersion(1, 2, 1);
```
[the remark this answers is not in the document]
patrick-dowell: Done

contracts/zero-ex/contracts/src/features/MetaTransactionsFeatureV2.sol
```diff
@@ -0,0 +1,641 @@
...
+            msg.sender.transfer(remainingBalance);
+        }
+    }
+
+    /// @dev Ensures that the ETH balance of `this` does not go below the
+    ///      initial ETH balance before the call (excluding ETH attached to the call).
+    modifier doesNotReduceEthBalance() {
+        uint256 initialBalance = address(this).balance - msg.value;
+        _;
+        require(initialBalance <= address(this).balance, "MetaTransactionDataV2/ETH_LEAK");
```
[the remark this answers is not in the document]
patrick-dowell: Done

contracts/zero-ex/tests/MetaTransactionV2Test.t.sol
```diff
@@ -0,0 +1,457 @@
...
+import "../contracts/src/features/interfaces/IMetaTransactionsFeature.sol";
+import "../contracts/test/TestMintTokenERC20Transformer.sol";
+import "../contracts/src/features/libs/LibSignature.sol";
+import "src/features/libs/LibNativeOrder.sol";
+import "../contracts/test/tokens/TestMintableERC20Token.sol";
+import "@0x/contracts-utils/contracts/src/v06/errors/LibRichErrorsV06.sol";
+import "@0x/contracts-erc20/src/IEtherToken.sol";
+
+contract MetaTransactionTest is LocalTest {
+    address private constant ZERO_ADDRESS = address(0);
```
duncancmt, line 32: Why is this a named constant?

contracts/zero-ex/contracts/src/features/MetaTransactionsFeatureV2.sol
```diff
@@ -0,0 +1,641 @@
...
+                    payer: state.mtx.signer
+                }),
+                minBuyAmount
+            )
+        );
+
+        // Unwrap and transfer WETH
+        uint256 boughtAmount = abi.decode(returnResult, (uint256));
+        WETH.withdraw(boughtAmount);
+        _transferEth(state.mtx.signer, boughtAmount);
```
duncancmt, line 548: Is there a reentrancy guard somewhere or other justification for why reentrancy is not a concern?

duncancmt (approved):
As  instructed by @patrick-dowell , I re-reviewed this by diffing `MetaTransactionsFeatureV2.sol` against `MetaTransactionsFeature.sol`. Seems pretty straightforward. I left a comment about reentrancy; I haven't gone deep on whether this is a real concern, but it's something to think about. Please add a comment addressing this concern before merging.

contracts/zero-ex/contracts/src/features/interfaces/IMultiplexFeature.sol
duncancmt: @dekz I see what you mean. I think this gets a little confusing with OTC/RFQ when the flow of value is bidirectional (equivalently: when the operation being executed is effectively a metatransaction with 2 signers). I concede that this isn't a big deal for the present PR, though.

contracts/zero-ex/contracts/src/features/MetaTransactionsFeatureV2.sol
patrick-dowell: Yea its protected by the fact that executeMetaTransaction is non-reentrant.

contracts/zero-ex/tests/MetaTransactionV2Test.t.sol
patrick-dowell: Just copy/pasted from previous test code that we wrote. I can get rid of it.

contracts/zero-ex/contracts/src/features/multiplex/MultiplexFeature.sol
```diff
@@ -544,7 +544,11 @@ contract MultiplexFeature is
                 // is executed, so we the target is the address encoded
                 // in the subcall data.
                 (target, ) = abi.decode(subcall.data, (address, bytes));
-            } else if (subcall.id == MultiplexSubcall.UniswapV3 || subcall.id == MultiplexSubcall.BatchSell) {
+            } else if (
+                subcall.id == MultiplexSubcall.UniswapV3 ||
+                subcall.id == MultiplexSubcall.BatchSell ||
+                subcall.id == MultiplexSubcall.OTC
```
duncancmt, line 550: The logic in `_multiHopSellOtcOrder` that uses `state.hopIndex` to compute the address to pull funds from is not compatible with `params.useSelfBalance`. It would be a pretty quick fix, but as it stands I would expect it not to work now that `OTC` is handled the same as `UniswapV3` and `BatchSell`

contracts/zero-ex/contracts/src/features/multiplex/MultiplexOtc.sol
```diff
@@ -65,4 +65,35 @@ abstract contract MultiplexOtc is FixinEIP712 {
...
+            "MultiplexOtcOrder::_multiHopSellOtcOrder/INVALID_TOKENS"
+        );
+        uint256 sellAmount = state.outputTokenAmount;
+        // Try filling the Otc order. Bubble up reverts.
+        (uint128 takerTokenFilledAmount, uint128 makerTokenFilledAmount) = IOtcOrdersFeature(address(this))
+            ._fillOtcOrder(
+                order,
+                signature,
+                sellAmount.safeDowncastToUint128(),
+                state.hopIndex == 0 ? params.payer : address(this),
```
duncancmt, line 92:
```suggestion
                state.from, // AKA: state.hopIndex == 0 ? (params.useSelfBalance ? address(this) : params.payer) : address(this),
```

patrick-dowell: Resolved via https://github.com/0xProject/protocol/pull/665/commits/172822044b2c21c221e63d1a1fe9d1af9b642d53

contracts/zero-ex/contracts/src/features/multiplex/MultiplexFeature.sol
patrick-dowell: https://github.com/0xProject/protocol/pull/665/commits/172822044b2c21c221e63d1a1fe9d1af9b642d53

abls (approved):

contracts/zero-ex/tests/MultiplexMetaTransactionsV2.t.sol
```diff
@@ -0,0 +1,407 @@
...
+    }
+
+    function test_metaTransaction_multiplexMultiHopSellTokenForEth_uniswapV2UniswapV3() external {
+        _createUniswapV2Pool(uniV2Factory, dai, shib, 10e18, 10e18);
+        _createUniswapV3Pool(uniV3Factory, shib, weth, 10e18, 10e18);
+        _mintTo(address(dai), otherSignerAddress, 1e18);
+
+        _executeMetaTransaction(
+            abi.encodeWithSelector(
+                zeroExDeployed.zeroEx.multiplexMultiHopSellTokenForToken.selector,
```
patrick-dowell, line 396: @abls this should be multiplexMultiHopSellTokenForEth.selector

PR #667, opened 2023-02-22
Feat/multi hop sell otc [LIT-870]

dextracker:
## Checklist:

### DIFF
--- a/contracts/zero-ex/contracts/src/features/multiplex/MultiplexFeature.sol
+++ b/contracts/zero-ex/contracts/src/features/multiplex/MultiplexFeature.sol
@@ -416,7 +416,9 @@ contract MultiplexFeature is
                 _multiHopSellLiquidityProvider(state, params, subcall.data);
             } else if (subcall.id == MultiplexSubcall.BatchSell) {
                 _nestedBatchSell(state, params, subcall.data);
-            } else {
+            } else if (subcall.id == MultiplexSubcall.OTC){
+                _multiHopSellOtcOrder(state, params, subcall.data);
+            }else {
                 revert("MultiplexFeature::_executeMultiHopSell/INVALID_SUBCALL");
             }
             // The recipient of the current hop will be the source
--- a/contracts/zero-ex/contracts/src/features/multiplex/MultiplexFeature.sol
+++ b/contracts/zero-ex/contracts/src/features/multiplex/MultiplexFeature.sol
@@ -529,9 +529,9 @@ contract MultiplexFeature is
                 // which `msg.sender` may not have an allowance set for. Thus
                 // target must be set to `address(this)` for `i > 0`.
                 if (i == 0 && !params.useSelfBalance) {
-                    target = msg.sender;
-                } else {
                     target = address(this);
+                } else {
+                    target = msg.sender;
                 }
             } else if (subcall.id == MultiplexSubcall.OTC) {
                 //on the first call we want to pull tokens from the taker, subsequent calls should use the EP balance
--- a/contracts/zero-ex/contracts/src/features/multiplex/MultiplexFeature.sol
+++ b/contracts/zero-ex/contracts/src/features/multiplex/MultiplexFeature.sol
@@ -30,6 +30,7 @@ import "./MultiplexRfq.sol";
 import "./MultiplexTransformERC20.sol";
 import "./MultiplexUniswapV2.sol";
 import "./MultiplexUniswapV3.sol";
+import "forge-std/Test.sol";
 
 /// @dev This feature enables efficient batch and multi-hop trades
 ///      using different liquidity sources.
@@ -391,6 +392,7 @@ contract MultiplexFeature is
         // Compute the expected address and transfer the input tokens
         // there if necessary.
         state.from = _computeHopTarget(params, 0);
+        
         // If the input tokens are currently held by `msg.sender` but
         // the first hop expects them elsewhere, perform a `transferFrom`.
         if (!params.useSelfBalance && state.from != msg.sender) {
@@ -407,7 +409,6 @@ contract MultiplexFeature is
             // Compute the recipient of the tokens that will be
             // bought by the current hop.
             state.to = _computeHopTarget(params, state.hopIndex + 1);
-
             if (subcall.id == MultiplexSubcall.UniswapV2) {
                 _multiHopSellUniswapV2(state, params, subcall.data);
             } else if (subcall.id == MultiplexSubcall.UniswapV3) {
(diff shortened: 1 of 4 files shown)


dextracker:
still writing the test but here is the protocol related code

duncancmt (approved):
Short and sweet. I like it.

contracts/zero-ex/contracts/src/features/multiplex/MultiplexOtc.sol
```diff
@@ -65,4 +65,47 @@ abstract contract MultiplexOtc is FixinEIP712 {
...
+        bytes memory wrappedCallData
+    ) internal {
+        // Decode the tokens[], Otc order, and signature.
+        (address[] memory tokens, LibNativeOrder.OtcOrder memory order, LibSignature.Signature memory signature) = abi.decode(
+            wrappedCallData,
+            (address[], LibNativeOrder.OtcOrder, LibSignature.Signature)
+        );
+        // Validate tokens.
+        require(
+            tokens.length >= 2 &&
```
duncancmt, line 81: Perhaps this is a dumb question: why allow `tokens` to have length `> 2`?

dextracker: this tokens array will contain all of the tokens in the swap, i.e. it needs to be more than 2 tokens to perform a multihop sell. we need token A, intermediate token, token B

contracts/zero-ex/contracts/src/features/multiplex/MultiplexOtc.sol
```diff
@@ -65,4 +65,47 @@ abstract contract MultiplexOtc is FixinEIP712 {
...
+        // Pre-emptively check if the order is expired.
+        uint64 expiry = uint64(order.expiryAndNonce >> 192);
+        if (expiry <= uint64(block.timestamp)) {
+            bytes32 orderHash = _getEIP712Hash(LibNativeOrder.getOtcOrderStructHash(order));
+            emit ExpiredOtcOrder(orderHash, order.maker, expiry);
+            return;
+        }
+
+        uint256 sellAmount = state.outputTokenAmount;
+        // Try filling the Otc order. Swallows reverts.
```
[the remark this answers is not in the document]
dextracker: ill probably revert to jacob on this one, there was a reason that otc order reverts did not bubble up but i dont remember what it was.

contracts/zero-ex/contracts/src/features/multiplex/MultiplexOtc.sol
duncancmt: Your comment made me look into this in more detail. As far as I can tell, that's not enforced anywhere. `tokens` comes from the subcall data and isn't even signed. As best I can tell, it's totally unconstrained and this `require` is basically vacuous

contracts/zero-ex/contracts/src/features/multiplex/MultiplexOtc.sol
```diff
@@ -65,4 +65,47 @@ abstract contract MultiplexOtc is FixinEIP712 {
...
+        try
+            IOtcOrdersFeature(address(this))._fillOtcOrder(
+                order,
+                signature,
+                sellAmount.safeDowncastToUint128(),
+                msg.sender,
```
patrick-dowell, line 101: This line is going to conflict with https://github.com/0xProject/protocol/pull/655

abls: *#665

contracts/zero-ex/contracts/src/features/multiplex/MultiplexFeature.sol
```diff
@@ -488,7 +489,7 @@ contract MultiplexFeature is
     // If `i == 0`, the target is the address which should hold the input
     // tokens prior to executing `calls[0]`. Otherwise, it is the address
     // that should receive `tokens[i]` upon executing `calls[i-1]`.
-    function _computeHopTarget(MultiHopSellParams memory params, uint256 i) private view returns (address target) {
+    function _computeHopTarget(MultiHopSellParams memory params, uint256 i) private  returns (address target) {
```
duncancmt: Why no longer view?

contracts/zero-ex/contracts/src/features/multiplex/MultiplexOtc.sol
duncancmt: Bump

abls: Duncan is right, unless there is something I'm missing, you should be checking `params.tokens` against the `makerToken` and `takerToken` of the `OtcOrder`.

Something like
```
require(
    order.takerToken == params.tokens[state.hopIndex] &&
    order.makerToken == params.tokens[state.hopIndex + 1],
   "MultiplexOtcOrder::_multiHopSellOtcOrder/INVALID_TOKENS"
);
```

contracts/zero-ex/contracts/src/features/multiplex/MultiplexFeature.sol
dextracker: was emitting an event for my test and had forgot to change it back, should be fixed now

contracts/zero-ex/contracts/src/features/multiplex/MultiplexOtc.sol
duncancmt: Can we get rid of `tokens` altogether? Seems like it's not being used now

contracts/zero-ex/contracts/src/features/multiplex/MultiplexOtc.sol
```diff
@@ -65,4 +65,33 @@ abstract contract MultiplexOtc is FixinEIP712 {
...
+            ._fillOtcOrder(
+                order,
+                signature,
+                sellAmount.safeDowncastToUint128(),
+                state.to == msg.sender ? address(this) : msg.sender,
+                params.useSelfBalance,
+                state.to
+            );
+        //store the bought amount for the next hop
+        state.outputTokenAmount = makerTokenFilledAmount;
```
duncancmt, line 95: What happens if there's leftover input token? Is that even possible?

duncancmt (approved):
LGTM. Please fix the linter complaints

contracts/zero-ex/contracts/src/features/multiplex/MultiplexOtc.sol
duncancmt: After talking to @dextracker , we're going to ignore any positive slippage on hops that come before an OTC. We hypothesize that the amounts will always be dust because with increasing order size comes a decreasing likelihood of positive slippage.

PR #674, opened 2023-03-07
Test votes migration for adding a new vote weight mechanism

duncancmt:
Add a migration test for the voting logic. We simulate upgrading this to add a cube root vote weight mechanism. This utilizes of the currently unused `DelegateInfo.balanceLastUpdated` property which is passed to the voting contract by the wrapped token.

### DIFF
--- a/contracts/governance/src/ZeroExVotes.sol
+++ b/contracts/governance/src/ZeroExVotes.sol
@@ -146,7 +146,7 @@ contract ZeroExVotes is IZeroExVotes, Initializable, OwnableUpgradeable, UUPSUpg
                     uint256 newWeight,
                     uint256 oldQuadraticWeight,
                     uint256 newQuadraticWeight
-                ) = _writeCheckpoint(_checkpoints[src], _subtract, srcBalance, amount);
+                ) = _writeCheckpoint(_checkpoints[src], _subtract, srcBalance, srcBalanceLastUpdated, amount);
 
                 emit DelegateVotesChanged(src, oldWeight, newWeight);
                 emit DelegateQuadraticVotesChanged(src, oldQuadraticWeight, newQuadraticWeight);
@@ -158,7 +158,7 @@ contract ZeroExVotes is IZeroExVotes, Initializable, OwnableUpgradeable, UUPSUpg
                     uint256 newWeight,
                     uint256 oldQuadraticWeight,
                     uint256 newQuadraticWeight
-                ) = _writeCheckpoint(_checkpoints[dst], _add, dstBalance, amount);
+                ) = _writeCheckpoint(_checkpoints[dst], _add, dstBalance, dstBalanceLastUpdated, amount);
 
                 emit DelegateVotesChanged(dst, oldWeight, newWeight);
                 emit DelegateQuadraticVotesChanged(dst, oldQuadraticWeight, newQuadraticWeight);
@@ -178,6 +178,7 @@ contract ZeroExVotes is IZeroExVotes, Initializable, OwnableUpgradeable, UUPSUpg
             _totalSupplyCheckpoints,
             _add,
             accountBalance,
+            0,
             amount
         );
 
@@ -196,6 +197,7 @@ contract ZeroExVotes is IZeroExVotes, Initializable, OwnableUpgradeable, UUPSUpg
             _totalSupplyCheckpoints,
             _subtract,
             accountBalance,
+            0,
             amount
         );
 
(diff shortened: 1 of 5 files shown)


contracts/governance/test/ZeroExVotesTest.t.sol
```diff
@@ -53,6 +55,53 @@ contract ZeroExVotesTest is BaseTest {
...
+        votes.upgradeToAndCall(address(newImpl), abi.encodeWithSignature("initialize()"));
+        assertEq(address(uint160(uint256(vm.load(address(votes), 0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc)))), address(newImpl));
+
+        vm.roll(block.number + 1);
+
+        vm.prank(account2);
+        wToken.transfer(address(this), 50e18);
+        vm.prank(account3);
+        wToken.transfer(address(this), 100e18);
+    }
```
duncancmt, line 87: Observe that we can't migrate to new forms of voting tabulation because balances need to be tagged with the blockheight at which they were last updated. The lined in `ZeroExVotesMigration`  with the comment `// underflow` is the culprit.

contracts/governance/test/ZeroExVotesTest.t.sol
```diff
@@ -53,6 +55,53 @@ contract ZeroExVotesTest is BaseTest {
...
+        vm.stopPrank();
+
+        // malicious upgrade
+        IZeroExVotes maliciousImpl = new ZeroExVotesMalicious();
+        votes.upgradeTo(address(maliciousImpl));
+
+        // try to withdraw withdraw
+        vm.prank(account2);
+        wToken.withdrawTo(account2, 1 ether);
+    }
```
duncancmt, line 103: If `ZRXWrappedToken` were to use `functionCallWithGas` or similar, this scenario could be avoided.

contracts/governance/test/ZeroExVotesMigration.sol
```diff
@@ -0,0 +1,137 @@
...
+        if (checkpoint.fromBlock <= migrationBlock) {
+            return 0;
+        }
+        return _toMigration(checkpoint).migratedVotes;
+    }
+
+    // TODO: we're not handling totalSupply
+
+    // TODO: need to return the migrated weight
+    function _writeCheckpoint(
```
duncancmt, line 85: This function is almost entirely copy-pasted from `ZeroExVotes`, except for the bits handling `.migratedVotes`

contracts/governance/src/ZeroExVotes.sol
```diff
@@ -233,18 +245,24 @@ contract ZeroExVotes is IZeroExVotes, Initializable, OwnableUpgradeable, UUPSUpg
...
+        // TODO: bad. very bad. only works on accident
+        if (high > 0) {
+            success = true;
+            result = _unsafeAccess(ckpts, high - 1);
+        } else {
+            // suppress compiler warning, which really shouldn't be suppressed
+            assembly {
+                result.slot := 0xDEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEF
+            }
+        }
```
duncancmt, line 257: I am absolutely not suggesting that we make this change to ZeroExVotes. It was simply the most expedient way for me to get my demonstration to work. An actual migration would involve either making this function generic or modifying it in-place.

contracts/governance/test/ZeroExVotesTest.t.sol
elenadimitrova: Just to be clear this scenario is only possible if the voting implementation is upgraded to a malicious one through governance?

duncancmt: Correct

contracts/governance/test/ZeroExVotesMigration.sol
```diff
@@ -0,0 +1,197 @@
...
+
+        // Remove the entire sqrt userBalance from quadratic voting power.
+        // Note that `userBalance` is value _after_ transfer.
+        if (pos > 0) {
+            uint256 oldQuadraticVotingPower = userBalance <= quadraticThreshold
+                ? userBalance
+                : quadraticThreshold + Math.sqrt((userBalance - quadraticThreshold) * 1e18);
+            oldCkpt.quadraticVotes -= SafeCast.toUint96(oldQuadraticVotingPower);
+
+            if (oldCkpt.fromBlock > migrationBlock && balanceLastUpdated > migrationBlock) {
```
duncancmt, line 168:
```suggestion
            if (balanceLastUpdated > migrationBlock) {
```

contracts/governance/test/ZeroExVotesMigration.sol
```diff
@@ -0,0 +1,197 @@
...
+        }
+    }
+
+    function getMigratedVotes(address account) public view returns (uint256) {
+        uint256 pos = _checkpoints[account].length;
+        if (pos == 0) {
+            return 0;
+        }
+        Checkpoint storage ckpt = _unsafeAccess(_checkpoints[account], pos - 1);
+        if (ckpt.fromBlock <= migrationTimestamp) {
```
duncancmt, line 60: comparison of block number and timestamp

contracts/governance/test/ZeroExVotesTest.t.sol
elenadimitrova: Added `DelegateInfo.balanceLastUpdated` block number stamps to the wrapped token.

contracts/governance/test/ZeroExVotesMigration.sol
elenadimitrova: That was a rushed idea and is now changed to be consistently a `block.number` for the governor, voting checkpoints and delegators' balance last updated property.

PR #676, opened 2023-03-09
feat: removing L2 Encoding from AaveV3 MixIn [LIT-901]

SavDont:
## Description

Removing the L2 Encoding for Aave V3 MixIn. The L2 encoding makes settlement more brittle and is causing some high revert rates specifically for MultiHop tokens involving Aave V3 tokens. 

## Types of changes

Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


dextracker (approved):
LGTM

duncancmt (approved):
This change improves gas efficiency on L2 because the original feature was not a good match with the L2 calldata fee model. These changes should also be backwards compatible because extra `bridgeData` should get ignored

PR #682, opened 2023-03-22
Add BridgeAdapter for Base Goerli testnet

duncancmt:
## Description

Add a BridgeAdapter for Base Goerli testnet, which has only UniV2 and UniV3 liquidity sources

## Testing instructions

No testing required


contracts/zero-ex/contracts/src/transformers/bridges/BaseGoerliBridgeAdapter.sol
```diff
@@ -0,0 +1,49 @@
...
+
+pragma solidity ^0.6.5;
+pragma experimental ABIEncoderV2;
+
+import "./AbstractBridgeAdapter.sol";
+import "./BridgeProtocols.sol";
+import "./mixins/MixinUniswapV3.sol";
+import "./mixins/MixinUniswapV2.sol";
+
+contract BaseGoerliBridgeAdapter is
```
kyu-c, line 23: I added [an escape patch](https://github.com/0xProject/protocol/commit/ec82c42c1b89ff79ba3133d3648b95b9f11415f8) for testnet couple weeks ago.

So we could potentially have a single adapter for both base and base goerli.

kyu-c (approved):

contracts/zero-ex/contracts/src/transformers/bridges/BaseGoerliBridgeAdapter.sol
kyu-c: or even just deploy `EthereumBridgeAdapter` on base goerli 🤔

duncancmt: Do we know what sources of liquidity will be available on base mainnet at launch? Will it still be just UniV2 and UniV3?

idokleinman: No, it's an unknown at this point. Really depends on CB's partnerships team results...

kyu-c: It doesn't hurt to have extra sources in the bridge adapter 🤔 

It will actually save us from needing to deploy FQT again so we may want to start with a comprehensive list (include it if we think there is a good chance). 

How about starting with:
* Uniswap v2 
* Uniswap v3
* Solidly
* Kyber Elastic
* Balancer v2 batch
* Curve 
* Aave v3

duncancmt:
subsumed by #734

PR #683, opened 2023-03-23
Governance integration testing

elenadimitrova:
## Description

Segregate unit tests integration tests and mocks, and wire up integration test base.

Implements production simulation of the exchange proxy governance and treasury governance migrations.

Closes https://linear.app/0xproject/issue/GOV-125/integration-testing


contracts/governance/test/BaseTest.t.sol
```diff
@@ -86,31 +80,88 @@ contract BaseTest is Test {
...
+    }
+
+    function mockZRXToken() internal returns (IERC20 zrxToken) {
+        vm.startPrank(account1);
+        bytes memory _bytecode = vm.getCode("./ZRXToken.json");
+        assembly {
+            zrxToken := create(0, add(_bytecode, 0x20), mload(_bytecode))
+        }
+        vm.stopPrank();
+    }
```
duncancmt, line 106: Should we use `create` here or a combination of `vm.record`, `vm.etch`, and `vm.store` to get ZRX on its actual addres?

duncancmt (commented):
Looks good so far. Do you want me to send you the keys for the vanity addresses?

contracts/governance/test/BaseTest.t.sol
elenadimitrova: This is only used in unit tests so I don't think we care the address it's deployed to as long as the bytecode's identical to production.

duncancmt (approved):

PR #685, opened 2023-03-28
Extendable Bridge Adapter

joshrrose:
## Description

Makes Bridge adapter infinitely extendable. 

Now there is a single bridge adapter that handles all chains. The large list of liquidity sources is broken down in groups `BridgeAdapterGroup1` and `BridgeAdapterGroup2`. that the `BridgeAdapter` will dispatch between by BridgeProtocolId

I re-did this changes because there were a lot of changes on top and rebase wasn't working very cleanly. 

I have used simbot to test on a few different chains. First i would remove all of the liquidity sources to prove my override worked. then after seeing many failures I would put the liquidity source back and run simbot again making sure that the run was clean. 

## Checklist:


dextracker (approved):
LGTM 👍

PR #690, opened 2023-04-03
Governance audit and fixes

elenadimitrova:
## Description

Adds audit report by Ourovoros for the governance contracts implementation. Below is a breakdown of each audit item:

### F-1: Unlocked Compiler Version
While it is true that we have not locked the solc version in the contracts we are instead using the foundry config to set the version specifically as `solc = '0.8.19'`. 

### F-2: Unused Returned Value
This is for the purposes of saving 100 extra gas in cases where a function has a return value as opposed to returning nothing.

### F-3: Explicit Variable Return
Fixed here.

### F-4: Unused Function
On the `SecurityCouncil.ejectSecurityCouncil` function we are going to potentially use that at some point. There is also a test for that which will be enabled together with the logic, see `testFailSecurityCouncilAreEjectedAfterCancellingAProposal`

`CallWithGas` we are accepting that there is an unused function but prefer to have an on chain copy of the library that's fully audited.

### F-5: modifier Optimization
This optimisation ends up saving between 3 to 6 gas in a large number of functions. imo the code is only slightly less readable so leaving this open for consideration on whether we should keep. 

Additionally while fixing this I noticed an opportunity to reuse the `onlySecurityCouncil` modifier so have applied that too.

### F-6: return Statement Optimization
Fixed here. Gas improvements are 118 less gas for each `getQuadraticVotes` and `getVotes`.

### F-7: Input Sanity Check 
Fixed here.

Closes https://linear.app/0xproject/issue/GOV-126/complete-audit


contracts/governance/src/ZeroExVotes.sol
```diff
@@ -72,15 +72,15 @@ contract ZeroExVotes is IZeroExVotes, Initializable, OwnableUpgradeable, UUPSUpg
      */
     function getVotes(address account) public view returns (uint256) {
         uint256 pos = _checkpoints[account].length;
-        return pos == 0 ? 0 : _checkpoints[account][pos - 1].votes;
+        return pos == 0 ? 0 : _unsafeAccess(_checkpoints[account], pos - 1).votes;
```
duncancmt, line 75: Did we determine that solc is smart enough to elide the underflow check here?

duncancmt (approved):
Can you please note our response to F-7 in the PR description? Otherwise, LGTM

contracts/governance/src/ZeroExVotes.sol
elenadimitrova: Ah so we're saving the array bounds check but still doing the underflow check. Wrapping these in `unchecked` did save 24 gas in addition to the 118 gas saved by the array bounds check in `getQuadraticVotes` and `getVotes`.

duncancmt (approved):
LGTM

PR #694, opened 2023-04-05
remove old fallback function sell/buy enforcement logic [LIT-960]

dextracker:
## Checklist:


contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -208,18 +208,6 @@ contract FillQuoteTransformer is Transformer {
 
         // Fill the orders.
         for (uint256 i = 0; i < data.fillSequence.length; ++i) {
-            // Check if we've hit our targets.
-            if (data.side == Side.Sell) {
-                // Market sell check.
-                if (state.soldAmount >= data.fillAmount) {
```
kyu-c, line 214: should we mark those `soldAmount`, `boughtAmount`, and etc. as deprecated in the state struct?

kyu-c (approved):

PR #696, opened 2023-04-06
Actual script for governance deployment

duncancmt:
Adding the script updates that were made for the production deployment.

Also adds a readme with the deployed contracts and high level design for governance.

Also cleaning up and removing an unused legacy json for the Treasury.


elenadimitrova (approved):
Thanks for the smooth deploy @duncancmt ❤️

PR #698, opened 2023-04-07
chore: update dependencies

rhinodavid:


contracts/test-utils/src/web3_wrapper.ts
```diff
@@ -31,26 +28,9 @@ export const providerConfigs: Web3Config = {
 export const provider: Web3ProviderEngine = web3Factory.getRpcProvider(providerConfigs);
 provider.stop();
 const isCoverageEnabled = env.parseBoolean(EnvVars.SolidityCoverage);
-const isProfilerEnabled = env.parseBoolean(EnvVars.SolidityProfiler);
```
rhinodavid, line 34: profiler provider is gone now -- only showed up once in any post-V3 code and suspect it was a copy paste since the line had not been touched in 4 years

contracts/zero-ex/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "0.39.2",
+        "changes": [
+            {
+                "note": "Uprgade dependencies"
+            }
+        ],
+        "timestamp": 1680893079
```
kyu-c, line 9: I think publish action will probably error out if timestamp is present. 

the same probably applies to other `CHANGELOG.json`s

kyu-c (approved):

kyu-c (approved):

PR #714, opened 2023-04-20
MetaTransaction V2 Packages

abls:
## Description

Accompanying typescript utilities for #665 

## Testing instructions

```
cd packages/protocol-utils
yarn test --fgrep "mtxs v2"
```

### DIFF
--- a/packages/protocol-utils/src/meta_transactions_v2.ts
+++ b/packages/protocol-utils/src/meta_transactions_v2.ts
@@ -1,0 +1,143 @@
+import { getContractAddressesForChainOrThrow } from '@0x/contract-addresses';
+import { EIP712TypedData } from '@0x/types';
+import { BigNumber, hexUtils, NULL_ADDRESS } from '@0x/utils';
+import { ZERO } from './constants';
+import {
+    createExchangeProxyEIP712Domain,
+    EIP712_DOMAIN_PARAMETERS,
+    getExchangeProxyEIP712Hash,
+    getTypeHash,
+} from './eip712_utils';
+
+export interface MetaTransactionV2Fee {
+    recipient: string,
+    amount: BigNumber,
+}
+
+const MTX_DEFAULT_VALUES = {
+    signer: NULL_ADDRESS,
+    sender: NULL_ADDRESS,
+    expirationTimeSeconds: ZERO,
+    salt: ZERO,
+    callData: hexUtils.leftPad(0),
+    feeToken: NULL_ADDRESS,
+    fees: [] as MetaTransactionV2Fee[],
+    chainId: 1,
+    verifyingContract: getContractAddressesForChainOrThrow(1).exchangeProxy,
+};
+
+export type MetaTransactionV2Fields = typeof MTX_DEFAULT_VALUES;
+
+export class MetaTransactionV2 {
+	public static readonly FEE_STRUCT_NAME = 'MetaTransactionFeeData';
+    public static readonly FEE_STRUCT_ABI = [
+        { type: 'address', name: 'recipient' },
+        { type: 'uint256', name: 'amount' },
+    ];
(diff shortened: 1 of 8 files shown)


packages/protocol-utils/test/meta_transactions_v2_test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+        it('can get the struct hash', () => {
+            const actual = mtx.getStructHash();
+            const expected = '0x441c8b8a8f25c1716f2880326d008d07d0b8eb9606623c0a81dd5d9fa14dd12e';
+            expect(actual).to.eq(expected);
+        });
+
+        it('can get the EIP712 hash', () => {
+            const actual = mtx.getHash();
+            const expected = '0xfc85ef2149bd49fcc8fee2571ed8f0ecd671dec03845637ab1ded3d891ac3386';
```
patrick-dowell, line 62: in the original metatransaction v1 typescript tests, this is done by comparing mtx.getHash() to the MetaTransactionsFeatureContract getMetaTransactionHash() function. Not sure if that'd be valuable here, but I thought I'd share the code:

```
const mtx = getRandomMetaTransaction();
const expected = mtx.getHash();
const actual = await feature.getMetaTransactionHash(mtx).callAsync();
expect(actual).to.eq(expected);
```

packages/protocol-utils/src/meta_transactions_v2.ts
```diff
@@ -0,0 +1,143 @@
...
+                hexUtils.leftPad(this.expirationTimeSeconds),
+                hexUtils.leftPad(this.salt),
+                hexUtils.hash(this.callData),
+                hexUtils.leftPad(this.feeToken),
+                hexUtils.leftPad(feesHash),
+            ),
+        );
+    }
+
+    public getEIP712TypedData(): EIP712TypedData {
```
patrick-dowell, line 116: Is there a way we can test this function for correctness?

packages/protocol-utils/test/meta_transactions_v2_test.ts
```diff
@@ -0,0 +1,66 @@
...
+            callData: '0x12345678',
+            feeToken: '0xcc3c7ea403427154ec908203ba6c418bd699f7ce',
+            fees: [] as any,
+            chainId: 8008,
+            verifyingContract: '0x6701704d2421c64ee9aa93ec7f96ede81c4be77d',
+        });
+
+        it('can get the struct hash', () => {
+            const actual = mtx.getStructHash();
+            const expected = '0x57db4055edfed82a6d86103197c21390bf29412fb4585e08c708454e03d92516';
```
Haozhuo, line 24: Quick question: how are the expected hash generated? If it's computed via `ethers` or solidity then we are good. Just want to make sure the expected hash is computed in a different method so that we can make sure our method implementation is correct. Same for other hashes in the test

packages/protocol-utils/test/meta_transactions_v2_test.ts
abls: That is designed to test `MetaTransactionsFeature`'s solidity mtx hash implementation using this package's implementation as a baseline. We shouldn't also circularly use that solidity code as the baseline for this test, because that makes it easier for both implementations to be wrong. The tests here use hashes generated by a third-party implementation as their baseline, just like the v1 package tests in `meta_transactions_test.ts`.

packages/protocol-utils/test/meta_transactions_v2_test.ts
abls: It's computed via `ethers`, so hopefully we are fine

Haozhuo: Got it. Yeah should be fine 👍

patrick-dowell (approved):
LGTM

PR #727, opened 2023-05-31
Fix usage of expectEmit in MetaTransactionsFeatureV2 tests

abls:
## Description

We are using forge's `vm.expectEmit` incorrectly in a few tests causing the tests to fail. This change corrects our usage in `MetaTransactionV2Test.t.sol`. The tests were previously passing, perhaps because of some bug in forge that must have been patched recently.

## Testing instructions

`forge test --match-contract MetaTransactionTest`

## Types of changes

* Bug fix (non-breaking change which fixes an issue)


patrick-dowell (approved):
I checked this out locally. It turns out my forge was a bit out of date and while it was passing on the old version, the new version correctly fails these tests without your fix. LGTM!

PR #734, opened 2023-07-17
Base mainnet

duncancmt:
## Checklist:


SavDont (approved):

dekz (approved):

kyu-c:
Have we considered not adding base goerli? 
I slightly prefer not adding base goerli unless we are going to use it or maintain it.

duncancmt:
I don't anticipate putting effort into maintaining the Base Goerli deployment, especially since it's in a marginally different state from the Base Mainnet deployment. I'm inclined to defer to @kyu-c since he would be the one whom this decision affects

kyu-c:
@duncancmt let's remove it for now (and only add if we end up needing it).

PR #735, opened 2023-07-17
Add UniswapV3 to Avalanche and BSC bridge adapters

kyu-c:
## Description
* Fixes LIT-1176
* Add Uniswap V3 to Avalanche and BSC bridge adapters
* Add router v2 support in `MixinUniswapV3` (backward compatible)
  * This change is needed as newer Uniswap V3 deployments do not have router v1 (e.g. BSC, Avalanche).
* Update Avalanche/BSC FillQuoteTransformer addresses 

## Types of changes

## Checklist:

### DIFF
--- a/contracts/zero-ex/contracts/src/transformers/bridges/mixins/MixinUniswapV3.sol
+++ b/contracts/zero-ex/contracts/src/transformers/bridges/mixins/MixinUniswapV3.sol
@@ -31,6 +31,18 @@ interface IUniswapV3Router {
     function exactInput(ExactInputParams memory params) external payable returns (uint256 amountOut);
 }
 
+// https://github.com/Uniswap/swap-router-contracts/blob/main/contracts/interfaces/IV3SwapRouter.sol
+interface IUniswapV3Router2 {
+    struct ExactInputParams {
+        bytes path;
+        address recipient;
+        uint256 amountIn;
+        uint256 amountOutMinimum;
+    }
+
+    function exactInput(ExactInputParams memory params) external payable returns (uint256 amountOut);
+}
+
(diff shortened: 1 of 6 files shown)


contracts/zero-ex/contracts/src/transformers/bridges/mixins/MixinUniswapV3.sol
```diff
@@ -39,19 +51,30 @@ contract MixinUniswapV3 {
         uint256 sellAmount,
         bytes memory bridgeData
     ) internal returns (uint256 boughtAmount) {
-        (IUniswapV3Router router, bytes memory path) = abi.decode(bridgeData, (IUniswapV3Router, bytes));
+        (address router, bytes memory path, uint256 routerVersion) = abi.decode(bridgeData, (address, bytes, uint256));
```
SavDont, line 54: Can we instead flat out replace router v1 with router v2 instead of encoding a version into the bridge data? Are there chains that don't have router v2 deployed?

SavDont (approved):

contracts/zero-ex/contracts/src/transformers/bridges/mixins/MixinUniswapV3.sol
kyu-c: we can if we atomically update contract addresses and router address in `api` (could be problematic in the unlikely event of rollback)

this will require redeploying FQTs on all chains that uses MixinUniswapV3 😣

SavDont: ah good point

SavDont (approved):

PR #760, opened 2023-11-03
Update docs with missing chains that 0x protocol is deployed on

wonge97:
## Description

Docs was missing Arbitrum, Base, zkEVM, Goerli, and Mumbai.

## Checklist:


duncancmt:
Since Goerli is deprecated, perhaps we shouldn't list it here? Especially since we plan to eventually switch over to Sepolia as is industry standard

duncancmt (approved):

PR #809, opened 2026-07-07
fix: replace node20 very_good_coverage with lcov threshold check

talalashraf:
GitHub removes Node ≤20 action runtimes from runners on 2026-09-16 — [changelog](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/). `VeryGoodOpenSource/very_good_coverage` is archived (node20); the coverage floor is enforced inline instead:

- `VeryGoodOpenSource/very_good_coverage@v2` (node20, archived) → an `awk` step that reads the same `lcov.info`, excludes test dirs, and fails below the 6.98% line-coverage floor

References
- https://linear.app/0xproject/issue/PLA-1666/upgrade-all-node-20-github-actions-org-wide-before-runner-removal


.github/workflows/ci.yml
```diff
@@ -132,11 +132,23 @@ jobs:
...
+                      let linesHit = 0;
+                      for (const line of lcov.split('\n')) {
+                        if (line.startsWith('LF:')) linesFound += Number(line.slice(3));
+                        else if (line.startsWith('LH:')) linesHit += Number(line.slice(3));
+                      }
+                      const coverage = linesFound ? (100 * linesHit) / linesFound : 0;
+                      core.info(`Line coverage: ${coverage.toFixed(2)}% (min ${MIN_COVERAGE}%)`);
+                      if (coverage < MIN_COVERAGE) {
+                        core.setFailed(`Coverage ${coverage.toFixed(2)}% is below the ${MIN_COVERAGE}% threshold`);
+                      }
```
tobernguyen, line 151: Are we missing the `exclude: '**/tests'`?

talalashraf: > The exclude was dropped intentionally because it was a no-op: very_good_coverage's exclude: '**/tests' uses minimatch, and **/tests matches a path that is tests — not files under tests/. Forge emits relative tests/... paths, so the action never actually excluded anything; coverage was always computed over all files, and the 6.98% floor is calibrated to that. So omitting it preserves the exact number.

talalashraf: Let's keep this as is and then see if anything changes ?

tobernguyen (approved):
LGTM. One note: the old very_good_coverage config had exclude: '**/tests' which the inline lcov check doesn't replicate, so the computed % may shift slightly — with a 6.98% threshold that's unlikely to matter. CI failure here ('Install dependencies') is pre-existing on every branch incl. development, unrelated to this change.

